PR #1, opened 2021-12-08
Initial charm

marceloneppel:
This PR introduces the following implementations:

* Initial charm structure based on the [operator template](https://github.com/canonical/operator-template).
* PostgreSQL deployment on bionic and focal series (without HA at this moment) with superuser password sharing among the units.
* Unit and integration tests.


src/charm.py
```diff
@@ -0,0 +1,165 @@
...
+            self.unit.status = WaitingStatus("waiting superuser password generation")
+            event.defer()
+
+    def _on_get_postgres_password(self, event: ActionEvent) -> None:
+        """Returns the password for the postgres user as an action response."""
+        event.set_results({"postgres-password": self._get_postgres_password()})
+
+    def _get_distro_codename(self) -> str:
+        """Return the distro release code name, eg. 'bionic' or 'focal'."""
+        return subprocess.check_output(["lsb_release", "-cs"]).decode().strip()
```
jnsgruk, line 104: might be worth using `distro`

```python
import distro
distro.codename() # => 'focal'
```

src/cluster.py
```diff
@@ -0,0 +1,226 @@
...
+        """Stop the PostgreSQL packages from creating the default cluster."""
+        os.makedirs(os.path.dirname(CREATE_CLUSTER_CONF_PATH), mode=0o755, exist_ok=True)
+        with open(CREATE_CLUSTER_CONF_PATH, mode="w") as file:
+            file.write("create_main_cluster = false\n")
+            file.write(f"include '{self.conf_path}/conf.d/postgresql-operator.conf'")
+
+    def _create_cluster(self, password: str) -> None:
+        """Create a new PostgreSQL cluster."""
+        # Write the password in a temporary file for pg_createcluster.
+        with open("/tmp/pwfile", mode="w") as file:
```
jnsgruk, line 81: Might be worth using the `tempfile` module here

src/cluster.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            command = [
+                "pg_createcluster",
+                self.version,
+                "main",
+                "--datadir=/var/lib/postgresql/data/pgdata",
+                "--",
+                "--pwfile=/tmp/pwfile",
+            ]
+            logger.info(f"pg_createcluster call: {' '.join(command)}")
```
jnsgruk, line 92: this should be `logger.debug`

src/cluster.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            clusters = (
+                subprocess.check_output(["pg_lsclusters", "--no-header"]).decode().splitlines()
+            )
+
+            def is_online(cluster: str):
+                """Return whether or not a cluster is online."""
+                if cluster.split()[3] == "online":
+                    return True
+                return False
```
jnsgruk, line 111: Just a stylistic point, and by no means mandatory, but this could probably be quite cleanly represented as a lambda when you call filter?

src/cluster.py
```diff
@@ -0,0 +1,226 @@
...
+
+            online_clusters = list(filter(is_online, clusters))
+            return len(online_clusters) == 1
+        except subprocess.CalledProcessError:
+            raise
+
+    def _render_file(self, path: str, content: str, mode: int) -> None:
+        """Write a content rendered from a template to a file."""
```
jnsgruk, line 119: Probably worth documenting the args in this docstring

src/cluster.py
```diff
@@ -0,0 +1,226 @@
...
+    def _start_cluster(self) -> None:
+        """Start a PostgreSQL cluster."""
+        try:
+            command = [
+                "pg_ctlcluster",
+                self.version,
+                "main",
+                "start",
+            ]
+            logger.info(f"pg_ctlcluster call: {' '.join(command)}")
```
jnsgruk, line 223: `logger.debug` for this :)

tests/integration/test_focal.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
```
jnsgruk, line 1: I think this file could be dropped and you could use pytest's parameterize feature in the other integration test to just deploy for both series'

tests/integration/test_bionic.py
```diff
@@ -0,0 +1,57 @@
...
+    # Testing the connection to each PostgreSQL instance.
+    status = await ops_test.model.get_status()  # noqa: F821
+    for _, unit in status["applications"][APP_NAME]["units"].items():
+        host = unit["public-address"]
+        logger.info("connecting to the database host: %s", host)
+        connection = psycopg2.connect(
+            f"dbname='postgres' user='postgres' host='{host}' password='{password}' connect_timeout=1"
+        )
+        assert connection.status == psycopg2.extensions.STATUS_READY
+        connection.close()
```
jnsgruk, line 57: Could we add another test here that gets cluster status/number of members/etc?

jnsgruk: On reflection, I wonder if its possible to use pylibjuju to pull files from the unit and check their contents are what we expect (for example config files?)

jnsgruk (changes requested):
Overall this is a great start! Just a few comments on tests and a couple of nits.

src/cluster.py
```diff
@@ -0,0 +1,226 @@
...
+            logger.info(f"pg_createcluster call: {' '.join(command)}")
+            subprocess.check_call(command)
+        except subprocess.CalledProcessError as e:
+            raise ClusterCreateError(e.stdout)
+        finally:
+            # Remove the temporary file.
+            os.remove("/tmp/pwfile")
+
+    def _is_cluster_running(self) -> bool:
+        """Check whether the cluster is running or not."""
```
wolsen, line 101: Just an FYI - not sure if you are aware of [py-postgresql](https://pypi.org/project/py-postgresql/). While it also has database drivers to query and talk with the database, it has some built-ins regarding [cluster management](https://pythonhosted.org/py-postgresql/cluster.html) which may be more convenient; however I think it does raw process management rather than leveraging systemd (though I'm sure that could be subclassed and extended that way, and even possibly contributed back upstream).

The advantage of using this library is that there's already users who are using it and helping to work out some various bugs on connections, cluster status, etc, etc.

src/cluster.py
```diff
@@ -0,0 +1,226 @@
...
+        # Set the correct ownership for the file.
+        os.chown(path, uid=u.pw_uid, gid=u.pw_gid)
+
+    def _render_pg_hba_conf_file(self) -> None:
+        """Render the application auth configuration file."""
+        # Open the template pg_hba.conf file.
+        with open("templates/pg_hba.conf.j2", "r") as file:
+            template = Template(file.read())
+        # Define the client authentication rules.
+        connections = [
```
wolsen, line 136: Is it intended that these will change at some point? Since they're all constants currently, doesn't seem like much of a need to provide them as variables to the template itself rather than just adding the contents to the file directly.

templates/pg_hba.conf.j2
```diff
@@ -0,0 +1,11 @@
+#########################################################################################
+# [ WARNING ]
+# postgresql client authentication configuration file maintained by the postgres-operator
+# local changes may be overwritten.
+#########################################################################################
+{%- for connection in connections %}
+{%- if connection.comment is defined %}
+# {{connection.comment}}
```
wolsen, line 8: nit: If a developer includes a newline in the comment this will break. I suppose its dev-mode and should be caught, but it might cause unintended confusion later.

tox.ini
```diff
@@ -0,0 +1,75 @@
...
+    black
+    isort
+commands =
+    isort {[vars]all_path}
+    black {[vars]all_path}
+
+[testenv:lint]
+description = Check code against coding style standards
+deps =
+    black
```
wolsen, line 37: nit: could just throw these in the test-requirements.txt rather than inline some and not others. Its kind of nice if there's generally one or two places to go and its consistent. Gets to be an awkward contributor experience if you have really dive into the different ways dependencies are managed and defined.

wolsen (commented):
Overall, looks like a really solid start on this. A few comments inline with some initial thoughts.

src/charm.py
```diff
@@ -0,0 +1,165 @@
...
+
+        self._postgresql_service = "postgresql"
+
+        self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.leader_elected, self._on_leader_elected)
+        self.framework.observe(self.on.start, self._on_start)
+        self.framework.observe(
+            self.on.get_postgres_password_action, self._on_get_postgres_password
+        )
+        self._cluster = PostgresqlCluster(self._get_postgresql_version())
```
wolsen, line 49: You essentially have a factory pattern here, but I think your factory is in the wrong place. Rather than include the `_get_postgresql_version` in the charm itself, it might make more sense to have a getPostgresqlCluster() method which introspects aspects of the system and returns the correct PostgresqlCluster object.

The only real reason you have the get_postgresql_version is for version strings in file paths, which could probably be contained within the PostgresqlCluster class itself; removing the need for a factory altogether.

src/charm.py
```diff
@@ -0,0 +1,165 @@
...
+        """Get postgres user password.
+
+        Returns:
+            The password from the peer relation or None if the
+            password has not yet been set by the leader.
+        """
+        data = self._peers.data[self.app]
+        return data.get("postgres-password", None)
+
+    def _get_postgresql_version(self) -> str:
```
wolsen, line 116: See my above comment about factory...

However, here you're hardcoding which versions map to which release; which works but its a bit static. You'll have to come in and update the mapping for each release.

An alternative might be to determine which postgresql version will be installed by referencing the apt-cache and pulling the version string from there. Then it becomes sufficiently insulated from the distro packaging tasks.

wolsen (changes requested):

actions.yaml
```diff
@@ -0,0 +1,5 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+get-postgres-password:
+  description: Get the initial postgres user password for the database.
```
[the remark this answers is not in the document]
jnsgruk: I've gone with 'get-initial-password' on other charms as a nod that it's generated, and should probably be changed. Until secrets arrives in Juju, I've not been providing a way to retrieve passwords set by the user.

tox.ini
jnsgruk: Just a comment here: in general I've been encouraging charm teams to have all runtime dependencies in requirements.txt, and all dev/test dependencies in the respective environment in tox.ini - but either way if test-requirements.txt is knocking around we could kill it in favour of this.

src/charm.py
marceloneppel: Good to know. I changed it to use the `distro` package.

src/cluster.py
marceloneppel: Fixed.

src/cluster.py
marceloneppel: Fixed.

src/cluster.py
marceloneppel: Yes, you are right. I removed this function and used a lambda expression.

src/cluster.py
marceloneppel: It makes sense. I've documented the args of this method.

tests/integration/test_focal.py
marceloneppel: Great suggestion Jon. I changed the code to use `pytest's parametrize` and also moved the charm `build` call to a `fixture` in order to run it just once (as we don't need different builds at this time for the two different series).

tests/integration/test_bionic.py
marceloneppel: I added a test to check the cluster status and a TODO to get the number of members in the replication when the replication is implemented.

marceloneppel: I also added a test to pull the configuration files contents and check that they are as expected.

actions.yaml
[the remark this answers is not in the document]
marceloneppel: I makes sense, Billy. I'll change that.

src/charm.py
marceloneppel: I agree, Billy! Thanks for catching that. I also use the version in the `apt.add_package` call (which didn't worked with the package `postgresql`, without the version - I'll check the call to make sure it's correct), which can also be retrieved with your suggestion about `apt-cache` in the comment below. I'll change that to get the version in PostgresqlCluster object, which makes more sense.

src/charm.py
marceloneppel: Thanks for the suggestion. I'll check `apt-cache` for the version info.

src/cluster.py
marceloneppel: Probably not. Accesses by relation will be made using the last rule. Therefore, the current file content is enough. I'll change to a file with all the content to copy to the file in the unit instead of the template.

templates/pg_hba.conf.j2
marceloneppel: I agree.
This template will be removed as said in the comment above  (there will be no future changes to the configuration file).

src/charm.py
marceloneppel: The version is now retrieved from installed packages or apt-cache using the apt library.

src/charm.py
marceloneppel: I removed the factory. I also updated the apt library in order to install the postgresql package instead of the ones with the version in the name.

tox.ini
marceloneppel: I  kept the dev dependencies in tox.ini as we talked in the meetings today.

src/cluster.py
marceloneppel: I removed the template and created a file with all the contents to be copied to the server instance.

src/cluster.py
marceloneppel: Great Billy. I agree with what you wrote and it's really good to know about this package. I tested it but found some problems for our use case. For the first one, related to how Postgres version is displayed on Ubuntu, I submitted a [PR](https://github.com/python-postgres/fe/pull/116), which needs the approval of the repo maintainers. The second problem is related to run the library code as postgres user instead of root, which can be achieved by using su like you said in our meeting. As we talked, we can address this change in a future PR.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,36 @@
+name: Tests
```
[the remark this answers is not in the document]
marceloneppel: Great. I added it to the github actions configuration file. I'm not sure, but I believe we may need to allow access from codecov to this repo.

actions.yaml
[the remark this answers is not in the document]
marceloneppel: I changed the action to get-initial-passwor. I'll also change in the k8s charm.

src/charm.py
marceloneppel: I removed this now as it's not needed anymore after the [update](https://github.com/canonical/operator-libs-linux/pull/35#issuecomment-993278821) in the apt library  and because the PostgresqlCluster class is now using the version retrieved through the apt library.

tests/integration/test_focal.py
marceloneppel: I also moved the expected content of the configuration files from the code to files in the repo in order to not duplicate them, as they are used in multiple source and test files.

tox.ini
wolsen: I think this is reasonable and really up to the development team to make this decision. As a note, the openstack charms (of which I have most familiarity) do it by dropping requirements into various {test-,build-,}requirements.txt files and we occasionally sync centralized copies of these files across all charms to have standard dependencies across all openstack charm projects.

jnsgruk (approved):
LGTM Marcelo, thanks!

src/cluster.py
wolsen: That's unfortunate, but yeah - its not something required here anyways. I was beginning to wonder about k8s usage as well.

wolsen (approved):

PR #3, opened 2022-03-29
DPE-71 Patroni auxiliary files

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-71](https://warthogs.atlassian.net/browse/DPE-71). The remaining implementation is split in two separate PRs to make the review easier.

In long:
* For this specific PR we need to provide some auxiliary files to start Patroni in the next PRs.

# Solution
* Create a minimum template to later generate content for the Patroni configuration file (patroni.yml).
* Create a minimum template to generate systemd unit file for Patroni service.
* Create test data files to match template generation on unit tests later
* Fetch the [systemd library ]((https://charmhub.io/operator-libs-linux/libraries/systemd)) from Charmhub and add it to the repository.

# Context
You don't have to review the systemd.py, since it's an imported charm library.

# Testing
* They will be implemented in the next PRs related to this JIRA ticket, as there is no logic being added or changed in this PR (only auxiliary files were added).

# Release Notes
* Add systemd library to the repository
* Add templates for patroni.yml (Patroni configuration file) and Patroni systemd unit
* Add test data files to be used later on unit tests


templates/patroni.service.j2
```diff
@@ -0,0 +1,22 @@
...
+User=postgres
+Group=postgres
+
+ExecStart=/usr/local/bin/patroni {{ conf_path }}/patroni.yml
+KillMode=process
+TimeoutSec=30
+Restart=no
+
+[Install]
+WantedBy=multi-user.targ
```
jnsgruk, line 22:
```suggestion
WantedBy=multi-user.target
```

Is this a typo? :)

templates/patroni.yml.j2
```diff
@@ -0,0 +1,46 @@
...
+  bin_dir: /usr/lib/postgresql/{{ version }}/bin/
+  data_dir: {{ conf_path }}/pgdata
+  pgpass: /tmp/pgpass
+  authentication:
+    replication:
+      username: replication
+      password: {{ replication_password }}
+    superuser:
+      username: postgres
+      password: {{ superuser_password }}
```
jnsgruk, line 46: It may be useful to comment some of the non-obvious lines here, it will serve as useful documentation for people trying to understand how the charm behaves. Another option is to just drop a link to the Patroni docs in a comment at the op if they're any good ;-)

tests/data/patroni.service
```diff
@@ -0,0 +1,22 @@
...
+User=postgres
+Group=postgres
+
+ExecStart=/usr/local/bin/patroni /var/lib/postgresql/data/patroni.yml
+KillMode=process
+TimeoutSec=30
+Restart=no
+
+[Install]
+WantedBy=multi-user.targ
```
jnsgruk, line 22:
```suggestion
WantedBy=multi-user.target
```

tests/data/patroni.service
```diff
@@ -0,0 +1,22 @@
+#########################################################################################
```
jnsgruk, line 1: Could this file not just be templated out at the time when the tests need it? Seems like it'll be a pain to keep this in sync with the template?

tests/data/patroni.yml
```diff
@@ -0,0 +1,46 @@
+#########################################################################################
```
jnsgruk, line 1: As above on templating at test-time to avoid sync issues?

zmraul (approved):
I don't know about Patroni specifics, but the template files look good, and they match the variables that are on the Postgresql charm (superuser_password, replication_password...). :)

templates/patroni.service.j2
marceloneppel: Yes, you're right Jon. Thanks for catching it.

jnsgruk (approved):

PR #4, opened 2022-03-29
DPE 71 Patroni resource install

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-71](https://warthogs.atlassian.net/browse/DPE-71). The remaining implementation is split in two separate PRs to make the review easier.

In long:
* Patroni is released as a Python package, which should be installed by the charm to later be started.

# Solution
* Enable the human operator to inform a resource containing Patroni Python package when deploying.
* Installing this resource together with PostgreSQL database in the charm

# Context
* Additional Patroni dependencies (like python-dev and psycopg2) will be changed to be installed as offline packages in a future PR.

# Testing
* Unit tests
* Some manual testing to ensure the charm still gets to the desired state
* Update of the `tox.ini` file and the deploy integration test to download and add Patroni as a resource (the successful install of the resource will be better tested in the next PR, where the bootstrap process is switched to use Patroni)

# Release Notes
* Add Patroni as a resource
* Add pip as a charm-python-packages part for bionic series
* Install Patroni dependencies (python-dev and psycopg2)
* Implement logic to install Patroni Python package
* Update `apt update` unit test matching the implementation of the `test_install_pip_packages` unit test

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -55,7 +55,11 @@ class PostgresqlOperatorCharm(CharmBase):
         self._cluster.inhibit_default_cluster_creation()
 
         # Install the PostgreSQL and Patroni requirements packages.
-        self._install_apt_packages(event, ["postgresql", "python3-pip", "python3-psycopg2"])
+        try:
+            self._install_apt_packages(event, ["postgresql", "python3-pip", "python3-psycopg2"])
+        except (subprocess.CalledProcessError, apt.PackageNotFoundError):
+            self.unit.status = BlockedStatus("failed to install apt packages")
+            return
 
         try:
             resource_path = self.model.resources.fetch("patroni")
@@ -65,8 +69,12 @@ class PostgresqlOperatorCharm(CharmBase):
             return
 
         # Build Patroni package path with raft dependency and install it.
-        patroni_package_path = f"{str(resource_path)}[raft]"
-        self._install_pip_packages([patroni_package_path])
+        try:
+            patroni_package_path = f"{str(resource_path)}[raft]"
+            self._install_pip_packages([patroni_package_path])
+        except subprocess.SubprocessError:
+            self.unit.status = BlockedStatus("failed to install Patroni python package")
+            return
 
         self.unit.status = WaitingStatus("waiting to start PostgreSQL")
 
@@ -120,24 +128,37 @@ class PostgresqlOperatorCharm(CharmBase):
         return data.get("postgres-password", None)
 
     def _install_apt_packages(self, _, packages: List[str]) -> None:
-        """Simple wrapper around 'apt-get install -y."""
+        """Simple wrapper around 'apt-get install -y.
+
+        Raises:
+            CalledProcessError if it fails to update the apt cache.
+            PackageNotFoundError if the package is not in the cache.
+            PackageError if the packages could not be installed.
+        """
         try:
             logger.debug("updating apt cache")
             apt.update()
         except subprocess.CalledProcessError as e:
             logger.exception("failed to update apt cache, CalledProcessError", exc_info=e)
-            self.unit.status = BlockedStatus("failed to update apt cache")
-            return
+            raise
 
-        try:
-            logger.debug(f"installing apt packages: {', '.join(packages)}")
-            apt.add_package(packages)
-        except apt.PackageNotFoundError:
-            logger.error("a specified package not found in package cache or on system")
-            self.unit.status = BlockedStatus("failed to install packages")
+        for package in packages:
+            try:
+                apt.add_package(package)
+                logger.debug(f"installed package: {package}")
+            except apt.PackageNotFoundError:
+                logger.error(f"package not found: {package}")
+                raise
+            except apt.PackageError:
+                logger.error(f"package error: {package}")
+                raise
 
     def _install_pip_packages(self, packages: List[str]) -> None:
-        """Simple wrapper around pip install."""
+        """Simple wrapper around pip install.
+
+        Raises:
+            SubprocessError if the packages could not be installed.
+        """
         try:
             command = [
                 "pip3",
@@ -148,7 +169,7 @@ class PostgresqlOperatorCharm(CharmBase):
             subprocess.check_call(command)
         except subprocess.SubprocessError:
             logger.error("could not install pip packages")
-            self.unit.status = BlockedStatus("failed to install pip packages")
+            raise
 
     def _new_password(self) -> str:
         """Generate a random password string.
--- a/src/charm.py
+++ b/src/charm.py
@@ -17,6 +17,7 @@ from ops.model import (
     ActiveStatus,
     BlockedStatus,
     MaintenanceStatus,
+    ModelError,
     Relation,
     WaitingStatus,
 )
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -73,4 +73,10 @@ deps =
     psycopg2-binary
     -r{toxinidir}/requirements.txt
 commands =
+    # Download patroni resource to use in the charm deployment.
+    sh -c 'stat patroni.tar.gz > /dev/null 2>&1 || curl "https://github.com/zalando/patroni/archive/refs/tags/v2.1.3.tar.gz" -L -s > patroni.tar.gz'
```
jnsgruk, line 77: Could this be some sort of fixture, or downloaded in the pytest code using urllib then cleaned up after the test?

The benefit here, is that once you have an uploaded version on chamrhub, you can fetch the resource straight from Charmhub using the API and get a bit more certainty in the test? (Perhaps a future PR?)

jnsgruk (approved):
Overall looking good, thanks Marcelo!

src/charm.py
```diff
@@ -53,8 +54,19 @@ def _on_install(self, event) -> None:
         # Prevent the default cluster creation.
         self._cluster.inhibit_default_cluster_creation()
 
-        # Install the PostgreSQL packages.
-        self._install_apt_packages(event, ["postgresql"])
+        # Install the PostgreSQL and Patroni requirements packages.
+        self._install_apt_packages(event, ["postgresql", "python3-pip", "python3-psycopg2"])
```
MiaAltieri, line 58: If this fails (due to a `subprocess.CalledProcessError` or `PackageNotFoundError`) the charm goes into `BlockedStatus` on lines [129](https://github.com/canonical/postgresql-operator/blob/6f0ef9eaaf039f56ff96ac6c3bd31a773f718e86/src/charm.py#L129_) and [137](https://github.com/canonical/postgresql-operator/blob/6f0ef9eaaf039f56ff96ac6c3bd31a773f718e86/src/charm.py#L137), but this isn't handled in the `install` hook, so I think it will go into the `BlockedStatus` and then try to execute the remaining code here. 

You'll want to add a check with a try or if statement. Up to you how you do it, but maybe:
```
# need to make _install_apt_packages return a boolean
install_successful = self._install_apt_packages(event, ["postgresql", "python3-pip", "python3-psycopg2"])
if not install_successful:
      return
```

src/charm.py
```diff
@@ -53,8 +54,19 @@ def _on_install(self, event) -> None:
...
+        try:
+            resource_path = self.model.resources.fetch("patroni")
+        except ModelError as e:
+            logger.error(f"missing patroni resource {str(e)}")
+            self.unit.status = BlockedStatus("Missing 'patroni' resource")
+            return
+
+        # Build Patroni package path with raft dependency and install it.
+        patroni_package_path = f"{str(resource_path)}[raft]"
+        self._install_pip_packages([patroni_package_path])
```
MiaAltieri, line 69: Same here,  on failure to install `pip` packages the code goes into `BlockedStatus` and then executes then and then goes into `WaitingStatus`

To prevent it from going into waiting status, maybe something like this (or similar) would do
```
# need to make _install_pip_packages return a boolean
install_successful = self._install_pip_packages([patroni_package_path]) 
if not install_successful:
      return
```

tests/unit/test_charm.py
```diff
@@ -29,17 +30,29 @@ def setUp(self):
         self.harness.begin()
         self.charm = self.harness.charm
 
+    @patch("charm.PostgresqlOperatorCharm._install_pip_packages")
     @patch("charm.PostgresqlOperatorCharm._install_apt_packages")
     @patch("charm.PostgresqlCluster.inhibit_default_cluster_creation")
     def test_on_install(
```
MiaAltieri, line 36: You might want to add two tests here:

`test_on_install_pip_failure`
`test_on_install_apt_failure`

To verify that on failure they stay in the `BlockedStatus`

MiaAltieri (changes requested):
Great stuff! Amazing PR description and some super thorough commenting. Also thanks for making the PR into two chunks, this is super manageable.

Just have one concern about error handling in install, feel free to PM me if you want to chat about it 😄  Other than that, it looks great!

src/charm.py
paulomach: I'm interested in the conclusion of this discussion, since I don't have a strong opinion on how to handle blocking errors like this.

paulomach (approved):
LGTM, considering Mia's comments

src/charm.py
marceloneppel: Nice catch Mia! Thank you! I'll update this trying to get the best approach (also thinking about the on start hook that is part of the next PR of this group). Also, thanks for the suggestions on how to fix it.

src/charm.py
marceloneppel: Thanks for catching it, Mia!

tests/unit/test_charm.py
marceloneppel: Nice addition! I'll add them.

tox.ini
marceloneppel: Yes, it can be, Jon. Nice suggestion! I'll create a new PR to change this that and upload the resource to Charmhub to have this improved flow of testing with the resource. I created an [issue](https://github.com/canonical/postgresql-operator/issues/5) to remind about it.

src/charm.py
marceloneppel: Code updated to correctly handle this situation. Thanks again, Mia. I'm using a try/except approach as [EAFP](https://docs.python.org/3.10/glossary.html#term-EAFP) seems to be the recommended way to do that.

MiaAltieri (commented):
Looks great to me! I like how you handled those cases on failure to install. Great work. 

Made some suggestions to the tests but if you disagree then feel free to ignore my suggestions :)

MiaAltieri (approved):

tests/unit/test_charm.py
```diff
@@ -29,23 +30,65 @@ def setUp(self):
...
+        # Mock the result of the call.
+        _install_apt_packages.side_effect = apt.PackageNotFoundError
+        # Trigger the hook.
+        self.charm.on.install.emit()
+        # Assert that the needed calls were made.
+        _inhibit_default_cluster_creation.assert_called_once()
+        _install_apt_packages.assert_called_once()
+        # Assert that the needed calls were made.
+        _install_pip_packages.assert_not_called()
+
```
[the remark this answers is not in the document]
marceloneppel: Nice suggestion. Added.

tests/unit/test_charm.py
```diff
@@ -29,23 +30,65 @@ def setUp(self):
...
+        self, _inhibit_default_cluster_creation, _install_apt_packages, _install_pip_packages
+    ):
+        # Add an empty file as Patroni resource just to check that the correct calls were made.
+        self.harness.add_resource("patroni", "")
+        self.charm.on.install.emit()
+        # Assert that the needed calls were made.
+        _inhibit_default_cluster_creation.assert_called_once()
+        _install_apt_packages.assert_called_once()
+        _install_pip_packages.assert_called_once()
+
```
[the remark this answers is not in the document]
marceloneppel: Nice suggestion. Added.

PR #8, opened 2022-04-16
DPE-71 Change bootstrap process to use Patroni

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-71](https://warthogs.atlassian.net/browse/DPE-71). The remaining implementation is split in three separate PRs to make the review easier.

In long:
* For this specific PR we change the bootstrap process to use Patroni instead of the tools that PostgreSQL itself provides.

# Solution
* Remove the default bootstrap that was implemented before.
* Change the on_start hook to use a new bootstrap process using Patroni, which later starts PostgreSQL for us.
* Also, checks that Patroni service was correctly started.

# Context
* The check for a blocked status before trying to start on on_start hook was made in a simple way to not make the code so complex at this moment. It probably should evolve as we find more situations where a blocked status should prevent some action to run, like on on_config_changed hook.
* The PostgresqlCluster class was renamed to Patroni to match better the purpose of its interactions. The file was not renamed to not generate a lot of diff.
* The bootstrap function is not tested in the unit tests as it would be just a lot of mocked calls. Its correct behavior guarantees the correct run of the integration tests.
* In the integration tests, the number of units were changed to 1 because for this PR it's not needed to have more than 1 unit, and also to speed up the tests. This should change later when testing the replication of the data between multiple units.

# Testing
* Unit and integrations tests were updated to check everything that was implemented.
* Manual testing was done.

# Release Notes
* Change PostgreSQL bootstrap process to be run through Patroni.
* Add check to not run on_start hook call to bootstrap if the unit is in a blocked state.


src/charm.py
```diff
@@ -93,36 +87,42 @@ def _on_leader_elected(self, _) -> None:
...
-                self.unit.status = BlockedStatus(f"failed to start cluster with error {e}")
-            except subprocess.CalledProcessError as e:
-                logging.error("failed to bootstrap cluster")
-                self.unit.status = BlockedStatus(f"failed to bootstrap cluster with error {e}")
-            else:
+        """Handle the start event."""
+        # Doesn't try to bootstrap the cluster if it's in a blocked state
+        # caused, for example, because a failed installation of packages.
+        if self._has_blocked_status:
+            return
```
shayancanonical, line 94: would `return`ing or `defer`ing be right thing to do here? i think `return`, but want to confirm the logic - once in a blocked status, the human operator will resolve all issues, run all actions not run and  then run `upgrade-charm`. is that right?

src/charm.py
```diff
@@ -93,36 +87,42 @@ def _on_leader_elected(self, _) -> None:
...
+
+        postgres_password = self._get_postgres_password()
+        replication_password = self._get_postgres_password()
+        # If the leader was elected and it generated the needed passwords,
+        # the cluster can be bootstrapped.
+        if postgres_password is not None and replication_password is not None:
```
shayancanonical, line 100: nit: i usually find it more readable and less complex to avoid if/else blocks. if i were to suggest a refactor of this condition, it would be like:
```
if not postgres_password or not replication_password:
     defer()
     return

<all other logic in the if statement without an indentation - thus reduced indentation complexity>
```

shayancanonical (approved):
LGTM, just left a minor comment and a minor confirmation

src/charm.py
```diff
@@ -93,36 +87,42 @@ def _on_leader_elected(self, _) -> None:
...
+            )
+            if success:
                 # The cluster is up and running.
                 self.unit.status = ActiveStatus()
+            else:
+                self.unit.status = BlockedStatus("failed to start Patroni")
         else:
             logger.info("leader not elected and/or superuser password not yet generated")
-            self.unit.status = WaitingStatus("waiting superuser password generation")
+            self.unit.status = WaitingStatus("waiting passwords generation")
```
WRFitch, line 114: minor nit: 
```suggestion
            self.unit.status = WaitingStatus("awaiting passwords generation")
```

src/cluster.py
```diff
@@ -25,57 +26,39 @@
 PATRONI_SERVICE = "patroni"
 
 
-class ClusterAlreadyRunningError(Exception):
-    """Exception raised when there is already a running cluster."""
+class Patroni:
+    """This class handles the bootstrap of a PostgreSQL database through Patroni."""
 
     pass
```
WRFitch, line 32: This line is unnecessary.

tests/integration/test_charm.py
```diff
@@ -96,26 +80,54 @@ async def test_database_is_up(ops_test: OpsTest, series: str):
...
+        # Here the SQL query gets a key-value pair composed by the name of the setting
+        # and its value, filtering the retrieved data to return only the settings
+        # that were set by Patroni.
+        cursor = connection.cursor()
+        cursor.execute(
+            """SELECT name,setting
+                FROM pg_settings
+                WHERE name IN
+                ('data_directory', 'cluster_name', 'data_checksums', 'listen_addresses');"""
+        )
```
WRFitch, line 101: You've forgotten to close `cursor`

tests/unit/test_charm.py
```diff
@@ -113,60 +101,56 @@ def test_on_leader_elected(self):
             self.charm._peers.data[self.charm.app].get("postgres-password", None), password
         )
 
-    @patch("charm.PostgresqlCluster.bootstrap_cluster")
+    @patch("charm.Patroni.bootstrap_cluster")
+    @patch("charm.PostgresqlOperatorCharm._replication_password")
     @patch("charm.PostgresqlOperatorCharm._get_postgres_password")
```
WRFitch, line 106: You can skip lines 115 and 116 by setting `return_value` in the decorator: 

```suggestion
    @patch("charm.PostgresqlOperatorCharm._replication_password", return_value="fake-replication-pw")
    @patch("charm.PostgresqlOperatorCharm._get_postgres_password", return_value="fake-postgres-pw")
```

tests/unit/test_charm.py
```diff
@@ -113,60 +101,56 @@ def test_on_leader_elected(self):
...
+
+        # Set an initial waiting status (like after the install hook was triggered).
+        self.harness.model.unit.status = WaitingStatus("fake message")
+
+        # Then test the event of a correct cluster bootstrapping.
+        self.charm.on.start.emit()
         self.assertEqual(
             self.harness.model.unit.status,
             ActiveStatus(),
         )
```
WRFitch, line 134: I like that you're asserting the unit status is an instance of BlockedStatus (or whatever status it is) in the rest of the tests, since it makes the test more robust to changes in status messages. Could you do the same here?

```suggestion
self.assertTrue(isinstance(self.harness.model.unit.status, ActiveStatus))
```

WRFitch (changes requested):
Mostly minor comments. My comment about adding an error to handle the BlockedStatus in _on_start isn't necessary to handle before approval, since you mentioned in the description that it's just a simple handler for now.

src/charm.py
WRFitch: I was reading about BlockedStatus and the `juju resolved` command, and I think if certain hooks run while the charm is blocked, it makes sense to raise an error instead. My reasoning is that a BlockedStatus should be raised only when a user is required to fix the charm. Once they finish fixing whatever caused the blockage, they can run juju resolved to re-run failed hooks, so hooks such as `on_start` should fail when the charm is blocked to make use of this command. If you go this route, it is worth documenting that this is the intended behaviour on BlockedStatus.

Incidentally, I don't think a user should need to run upgrade-charm when fixing a blockage, since they would likely be downloading the charm again.

src/charm.py
marceloneppel: It looks a lot better this way. Thanks Shayan! Updated on https://github.com/canonical/postgresql-operator/pull/8/commits/d2d0915c789894145f29c296de6b4ec5492cf382.

src/cluster.py
marceloneppel: Good catch. Removed `pass` on https://github.com/canonical/postgresql-operator/pull/8/commits/d2d0915c789894145f29c296de6b4ec5492cf382.

src/charm.py
marceloneppel: Updated on https://github.com/canonical/postgresql-operator/pull/8/commits/d2d0915c789894145f29c296de6b4ec5492cf382. Thanks!

tests/integration/test_charm.py
marceloneppel: Thanks for catching this, Will. Fixed on https://github.com/canonical/postgresql-operator/pull/8/commits/d2d0915c789894145f29c296de6b4ec5492cf382.

tests/unit/test_charm.py
marceloneppel: Updated on https://github.com/canonical/postgresql-operator/commit/d2d0915c789894145f29c296de6b4ec5492cf382.

src/charm.py
WRFitch: Thanks for the explanation, your approach makes more sense Marcelo :)

WRFitch (approved):

shayancanonical (approved):

PR #10, opened 2022-05-05
Scaling helper methods

marceloneppel:
# Issue
This is a PR that was on hold to copy the feedback that improved the PR from https://github.com/canonical/postgresql-k8s-operator/pull/9.

It is part of the scale up/down logic that is needed to avoid adding or removing more than one member to a PostgreSQL cluster if another member is still syncing data after it has joined.

In long:
* To avoid overload of the network during copy of data between the replicas that are being added to the cluster during a scale up or downtime in the case of a scale down, the charm needs to add and remove on member at a time from the cluster so.
* If the member is the current primary, we need to trigger a switchover before the unit is removed, to make avoid downtime in the cluster.

# Solution
* Add some helper methods to add/remove one member at a time and also trigger a switchover when the primary is the unit that is being removed in a scale down process.

# Context
This PR only adds some helper methods that will be used on https://github.com/canonical/postgresql-operator/pull/11.
In `charm.py`:
* Added an action to get the primary unit.
* Added methods for chaning the primary (on a scaling down process if the primary is the unit that is being removed).
* Added some methods to list the IPs and the units that are part of the cluster.
* Added methods to add and remove members to the cluster (through adding them to peer relation data that is later used to update the patroni configuration file, which is used to update the list of cluster members).
In `cluster.py`:
* Added custom exceptions for cluster not ready for adding/removing new members (because of not all members are ready, probably because of a data sync).
* Added methods for getting the primary and cluster members.
* Added methos for checking if all members are ready and whether the member has started (Patroni and PostgreSQL are running).
* Added methods to change the primary through a switchover and check that it happended.
* Added methods for updating and reloading Patroni configuration (this will be used to update the list of members of the cluster).

# Testing
* The new tests are implemented on https://github.com/canonical/postgresql-operator/pull/14.

# Release Notes
* Add helper methods for scaling up/down.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -148,7 +148,14 @@ class PostgresqlOperatorCharm(CharmBase):
         self._update_members_ips(ip_to_remove=ip)
 
     def _update_members_ips(self, ip_to_add: str = None, ip_to_remove: str = None) -> None:
-        """Update cluster members IPs."""
+        """Update cluster member IPs on application data.
+
+        Member IPs on application data are used to determine when a unit of PostgreSQL
+        should be added or removed from the PostgreSQL cluster.
+
+        NOTE: this function does not update the IPs on the PostgreSQL cluster
+        in the Patroni configuration.
+        """
         # Allow leader to reset which members are part of the cluster.
         if not self.unit.is_leader():
             return
--- a/src/charm.py
+++ b/src/charm.py
@@ -3,7 +3,7 @@
 # See LICENSE file for licensing details.
 
 """Charmed Machine Operator for the PostgreSQL database."""
-
+import json
 import logging
 import secrets
 import string
@@ -19,10 +19,12 @@ from ops.model import (
     MaintenanceStatus,
     ModelError,
     Relation,
+    Unit,
     WaitingStatus,
 )
+from tenacity import RetryError, retry, stop_after_delay, wait_fixed
 
-from cluster import Patroni
+from cluster import Patroni, SwitchoverFailedError
 
 logger = logging.getLogger(__name__)
 
@@ -39,9 +41,123 @@ class PostgresqlOperatorCharm(CharmBase):
 
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.leader_elected, self._on_leader_elected)
+        self.framework.observe(self.on.get_primary_action, self._on_get_primary)
         self.framework.observe(self.on.start, self._on_start)
         self.framework.observe(self.on.get_initial_password_action, self._on_get_initial_password)
         self._cluster = Patroni(self._unit_ip)
+
+    def _on_get_primary(self, event: ActionEvent) -> None:
+        """Get primary instance."""
+        try:
+            primary = self._patroni.get_primary(unit_name_pattern=True)
+            event.set_results({"primary": primary})
+        except RetryError as e:
+            logger.error(f"failed to get primary with error {e}")
+
+    @retry(
+        stop=stop_after_delay(60),
+        wait=wait_fixed(5),
+        reraise=True,
+    )
+    def _change_primary(self) -> None:
+        """Change the primary member of the cluster."""
+        # Inform the first of the remaining available members to not incur the risk
+        # of triggering a switchover to a member that is also being removed.
+        if not self._patroni.are_all_members_ready():
+            logger.info("could not switchover because not all members are ready")
+            return
+
+        # Try switchover and raise and exception if it doesn't succeed.
+        # If it doesn't happen on time, Patroni will automatically run a fail-over.
+        try:
+            self._patroni.switchover()
+            logger.info("successful switchover")
+        except SwitchoverFailedError as e:
+            logger.error(f"switchover failed with reason: {e}")
+
+    def _get_ip_by_unit(self, unit: Unit) -> str:
+        """Get the IP address of a specific unit."""
+        return self._peers.data[unit].get("private-address")
+
+    @property
+    def _hosts(self) -> set:
+        """List of the current Juju hosts.
+
+        Returns:
+            a set containing the current Juju hosts
+                with the names in the k8s pod name format
+        """
+        peers = self.model.get_relation(PEER)
+        hosts = [self.unit.name.replace("/", "-")] + [
+            unit.name.replace("/", "-") for unit in peers.units
+        ]
+        return set(hosts)
+
+    @property
+    def _patroni(self) -> Patroni:
+        """Returns an instance of the Patroni object."""
+        return Patroni(
+            self._unit_ip,
+            self._storage_path,
+            self._cluster_name,
+            self._member_name,
+            self.app.planned_units(),
+            self._peers_ips,
+            self._get_postgres_password(),
+            self._replication_password,
+        )
+
+    @property
+    def _peers_ips(self) -> List[str]:
+        """Fetch current list of peers IPs.
+
+        Returns:
+            A list of peers addresses (strings).
+        """
+        # Get all members IPs and remove the current unit IP from the list.
+        addresses = self.members_ips
+        current_unit_ip = self._unit_ip
+        if current_unit_ip in addresses:
+            addresses.remove(current_unit_ip)
+        return addresses
+
+    @property
+    def _units_ips(self) -> List[str]:
+        """Fetch current list of peers IPs.
+
+        Returns:
+            A list of peers addresses (strings).
+        """
+        # Get all members IPs and remove the current unit IP from the list.
+        addresses = [self._get_ip_by_unit(unit) for unit in self._peers.units]
+        addresses.append(self._unit_ip)
+        return addresses
+
+    @property
+    def members_ips(self) -> List[str]:
+        """Returns the list of IPs addresses of the current members of the cluster."""
+        return json.loads(self._peers.data[self.app].get("members_ips", "[]"))
+
+    def _add_to_members_ips(self, ip: str) -> None:
+        """Add one IP to the members list."""
+        self._update_members_ips(ip_to_add=ip)
+
+    def _remove_from_members_ips(self, ip: str) -> None:
+        """Remove IPs from the members list."""
+        self._update_members_ips(ip_to_remove=ip)
+
+    def _update_members_ips(self, ip_to_add: str = None, ip_to_remove: str = None) -> None:
+        """Update cluster members IPs."""
+        # Allow leader to reset which members are part of the cluster.
+        if not self.unit.is_leader():
+            return
+
+        ips = json.loads(self._peers.data[self.app].get("members_ips", "[]"))
+        if ip_to_add and ip_to_add not in ips:
+            ips.append(ip_to_add)
+        elif ip_to_remove:
+            ips.remove(ip_to_remove)
+        self._peers.data[self.app]["members_ips"] = json.dumps(ips)
 
     @property
     def _unit_ip(self) -> str:
@@ -85,6 +201,13 @@ class PostgresqlOperatorCharm(CharmBase):
         # The leader sets the needed password on peer relation databag if they weren't set before.
         data.setdefault("postgres-password", self._new_password())
         data.setdefault("replication-password", self._new_password())
+
+    def _get_ips_to_remove(self) -> List[str]:
+        """List the IPs that were part of the cluster but departed."""
+        old = self.members_ips
+        current = self._units_ips
+        endpoints_to_remove = list(set(old) - set(current))
+        return endpoints_to_remove
 
     def _on_start(self, event) -> None:
         """Handle the start event."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -11,13 +11,7 @@ import subprocess
 from typing import List
 
 from charms.operator_libs_linux.v0 import apt
-from ops.charm import (
-    ActionEvent,
-    CharmBase,
-    LeaderElectedEvent,
-    RelationChangedEvent,
-    RelationDepartedEvent,
-)
+from ops.charm import ActionEvent, CharmBase
 from ops.main import main
 from ops.model import (
     ActiveStatus,
--- a/src/charm.py
+++ b/src/charm.py
@@ -61,19 +61,23 @@ class PostgresqlOperatorCharm(CharmBase):
     )
     def _change_primary(self) -> None:
         """Change the primary member of the cluster."""
-        # Inform the first of the remaining available members to not incur the risk
-        # of triggering a switchover to a member that is also being removed.
-        if not self._patroni.are_all_members_ready():
-            logger.info("could not switchover because not all members are ready")
-            return
-
-        # Try switchover and raise and exception if it doesn't succeed.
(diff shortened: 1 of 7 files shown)


src/charm.py
```diff
@@ -39,10 +41,124 @@ def __init__(self, *args):
...
+    def _add_to_members_ips(self, ip: str) -> None:
+        """Add one IP to the members list."""
+        self._update_members_ips(ip_to_add=ip)
+
+    def _remove_from_members_ips(self, ip: str) -> None:
+        """Remove IPs from the members list."""
+        self._update_members_ips(ip_to_remove=ip)
+
+    def _update_members_ips(self, ip_to_add: str = None, ip_to_remove: str = None) -> None:
+        """Update cluster members IPs."""
```
MiaAltieri, line 150: nit: 
"""Update cluster member IPs on app data.

Member IPs on app data are used to determine when a unit of PostGres should be added or removed from the PostGres Cluster. 

NOTE: this function does not update the IPs on the PostGres cluster.
"""

src/charm.py
```diff
@@ -39,10 +41,124 @@ def __init__(self, *args):
...
+        # If it doesn't happen on time, Patroni will automatically run a fail-over.
+        try:
+            self._patroni.switchover()
+            logger.info("successful switchover")
+        except SwitchoverFailedError as e:
+            logger.error(f"switchover failed with reason: {e}")
+
+    def _get_ip_by_unit(self, unit: Unit) -> str:
+        """Get the IP address of a specific unit."""
+        return self._peers.data[unit].get("private-address")
```
MiaAltieri, line 80: this won't work when refering to the self unit. For getting the self unit IP I'd follow [this](https://github.com/canonical/mongodb-operator/blob/f8d9299605ae34ecde8a9de57f62f599e4560a12/src/charm.py#L411-L421)

Note if you follow this suggestion you'll have to update the function `_units_ips` to not add the self unit.

If you don't follow this suggestion change the name/docstring to make it clear that this only works for peer units.

src/charm.py
```diff
@@ -39,10 +41,124 @@ def __init__(self, *args):
...
+
+        Returns:
+            A list of peers addresses (strings).
+        """
+        # Get all members IPs and remove the current unit IP from the list.
+        addresses = self.members_ips
+        current_unit_ip = self._unit_ip
+        if current_unit_ip in addresses:
+            addresses.remove(current_unit_ip)
+        return addresses
```
MiaAltieri, line 122: if the function `update_member_ips` hasn't been called the data in `members_ips` might not include all the peers, to be certain that you get the peer ips. I'd suggest doing 

```suggestion
    def _peers_ips(self) -> List[str]:
        """Fetch current list of peers IPs.

        Returns:
            A list of peers addresses (strings).
        """
        # Get all members IPs and remove the current unit IP from the list.
        peer_ips = [ self._get_ip_by_unit(unit) unit in self._peers.units] 
        return peer_ips
```

src/charm.py
```diff
@@ -39,10 +41,124 @@ def __init__(self, *args):
...
+    def _hosts(self) -> set:
+        """List of the current Juju hosts.
+
+        Returns:
+            a set containing the current Juju hosts
+                with the names in the k8s pod name format
+        """
+        peers = self.model.get_relation(PEER)
+        hosts = [self.unit.name.replace("/", "-")] + [
+            unit.name.replace("/", "-") for unit in peers.units
```
MiaAltieri, line 92: peers doesn't include the current unit.  might need to change this or add the current unit

src/charm.py
```diff
@@ -39,10 +41,124 @@ def __init__(self, *args):
...
+        """Get the IP address of a specific unit."""
+        return self._peers.data[unit].get("private-address")
+
+    @property
+    def _hosts(self) -> set:
+        """List of the current Juju hosts.
+
+        Returns:
+            a set containing the current Juju hosts
+                with the names in the k8s pod name format
```
MiaAltieri, line 88: remove k8s

src/charm.py
```diff
@@ -39,10 +41,124 @@ def __init__(self, *args):
...
+        stop=stop_after_delay(60),
+        wait=wait_fixed(5),
+        reraise=True,
+    )
+    def _change_primary(self) -> None:
+        """Change the primary member of the cluster."""
+        # Inform the first of the remaining available members to not incur the risk
+        # of triggering a switchover to a member that is also being removed.
+        if not self._patroni.are_all_members_ready():
+            logger.info("could not switchover because not all members are ready")
```
MiaAltieri, line 67: maybe defer to wait until all members are ready?

src/charm.py
```diff
@@ -39,10 +41,124 @@ def __init__(self, *args):
...
+            return
+
+        # Try switchover and raise and exception if it doesn't succeed.
+        # If it doesn't happen on time, Patroni will automatically run a fail-over.
+        try:
+            self._patroni.switchover()
+            logger.info("successful switchover")
+        except SwitchoverFailedError as e:
+            logger.error(f"switchover failed with reason: {e}")
```
MiaAltieri, line 76: should something be done when `SwitchoverFailedError `, since this error isn't re-`raised` the tenacity retry won't get triggered

src/cluster.py
```diff
@@ -171,3 +254,41 @@ def _start_patroni(self) -> bool:
         """
         service_start(PATRONI_SERVICE)
         return service_running(PATRONI_SERVICE)
+
+    def switchover(self, candidate: str = None) -> None:
+        """Schedule a switchover to a given candidate member????."""
```
MiaAltieri, line 259: remove ???

src/cluster.py
```diff
@@ -95,6 +122,62 @@ def _get_postgresql_version(self) -> str:
...
+        Returns:
+            True if all members are ready False otherwise. Retries over a period of 10 seconds
+            3 times to allow server time to start up.
+        """
+        # Request info from cluster endpoint
+        # (which returns all members of the cluster and their states).
+        try:
+            for attempt in Retrying(stop=stop_after_delay(10), wait=wait_fixed(3)):
+                with attempt:
+                    r = requests.get(f"http://{self.unit_ip}:8008/cluster")
```
MiaAltieri, line 158:
```suggestion
               
                    cluster_status = requests.get(f"http://{self.unit_ip}:8008/cluster")
```

src/cluster.py
```diff
@@ -95,6 +122,62 @@ def _get_postgresql_version(self) -> str:
...
+
+        Args:
+            unit_name_pattern: whether to convert pod name to unit name
+
+        Returns:
+            primary pod or unit name.
+        """
+        primary = None
+        # Request info from cluster endpoint (which returns all members of the cluster).
+        r = requests.get(f"http://{self.unit_ip}:8008/cluster")
```
MiaAltieri, line 136:
```suggestion
        cluster_status = requests.get(f"http://{self.unit_ip}:8008/cluster")
```

src/charm.py
paulomach: Or put this test on the calling handler, since there's no access to the event from the function.

src/charm.py
```diff
@@ -86,6 +202,13 @@ def _on_leader_elected(self, _) -> None:
         data.setdefault("postgres-password", self._new_password())
         data.setdefault("replication-password", self._new_password())
 
+    def _get_ips_to_remove(self) -> List[str]:
+        """List the IPs that were part of the cluster but departed."""
+        old = self.members_ips
+        current = self._units_ips
+        endpoints_to_remove = list(set(old) - set(current))
```
paulomach, line 209: I'm wondering why don't treat all of this structures as sets. So you avoid this conversion/reconversion

src/cluster.py
```diff
@@ -69,6 +88,14 @@ def _change_owner(self, path: str) -> None:
         # Set the correct ownership for the file or directory.
         os.chown(path, uid=user_database.pw_uid, gid=user_database.pw_gid)
 
+    @property
+    @retry(stop=stop_after_attempt(10), wait=wait_exponential(multiplier=1, min=2, max=10))
+    def cluster_members(self) -> set:
+        """Get the current cluster members."""
+        # Request info from cluster endpoint (which returns all members of the cluster).
+        r = requests.get(f"http://{self.unit_ip}:8008/cluster")
```
paulomach, line 96: `requests.get` can raise a lot of different errors, so should these calls to patroni API be wrapped with `try-except`'s or is generelly safe to go without in this context?

templates/patroni.yml.j2
```diff
@@ -11,6 +11,9 @@ scope: {{ scope }}
 namespace: /db/
 name: {{ member_name }}
 
+log:
+  dir: /var/log/postgresql
```
paulomach, line 15: Was this not default? Is this log rotate? How big can it get?
(I've saw log files filling up disk and disrupting services too many times...)

tox.ini
```diff
@@ -75,7 +75,7 @@ deps =
 commands =
     # Download patroni resource to use in the charm deployment.
     sh -c 'stat patroni.tar.gz > /dev/null 2>&1 || curl "https://github.com/zalando/patroni/archive/refs/tags/v2.1.3.tar.gz" -L -s > patroni.tar.gz'
-    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
+    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} --asyncio-mode=auto
```
paulomach, line 78: Maybe add to pyproject.toml instead?

```toml
[tool.pytest.ini_options]
minversion = "6.0"
log_cli_level = "INFO"
asyncio_mode = "auto"
```

paulomach (approved):
Minor comments, but looks good. Will check PR#11 for complete context

src/charm.py
zmraul: Yeah, maybe raise it to the event handler and set the application to BlockedStatus?

tox.ini
marceloneppel: Changed in https://github.com/canonical/postgresql-operator/pull/10/commits/4bdb47ed4156478d5f67dfa98e8b67c60a5fabaf.

templates/patroni.yml.j2
marceloneppel: No. It was not default. When the directory is added to the configuration, Patroni will retain 4 25MB logs by default. There is a note about that in https://patroni.readthedocs.io/en/latest/SETTINGS.html.

src/charm.py
marceloneppel: Good suggestion. Changed in https://github.com/canonical/postgresql-operator/pull/10/commits/bf34427e76fade29e4297b0dc5ae018257b0b69c.

src/cluster.py
marceloneppel: Variable renamed in https://github.com/canonical/postgresql-operator/pull/10/commits/0afd3c85851e1dc5c9e3350dd29cbc130b28c056.

src/cluster.py
marceloneppel: Variable renamed in https://github.com/canonical/postgresql-operator/pull/10/commits/0afd3c85851e1dc5c9e3350dd29cbc130b28c056.

src/cluster.py
marceloneppel: Fixed in https://github.com/canonical/postgresql-operator/pull/10/commits/bd497b27badc78c4db1d0de1b8eb5e956f9769de. Thanks!

src/charm.py
marceloneppel: I fixed that and reorganized the code structure in https://github.com/canonical/postgresql-operator/pull/10/commits/b92efb42f7a7fd1e79bfd418afae2e9a828fa46f. The status of the app should not change to be blocked as Patroni will do an automatic failover if the switchover does not happen for some reason.

src/charm.py
marceloneppel: The event should not be deferred because it is the storage detaching event. I moved this check to the calling handler in the next PR and removed it from here in https://github.com/canonical/postgresql-operator/pull/10/commits/b92efb42f7a7fd1e79bfd418afae2e9a828fa46f.

src/charm.py
marceloneppel: Good catch. Fixed in https://github.com/canonical/postgresql-operator/pull/10/commits/61273256b0210debc93689b8cf62157923190cd6.

src/charm.py
marceloneppel: Changed the method name and docstring in https://github.com/canonical/postgresql-operator/pull/10/commits/e75180a372f5ee3e7e8dca0fee27c099122e68e6 to make it clear that is only for peer units.

src/charm.py
marceloneppel: This description is great. Added in https://github.com/canonical/postgresql-operator/pull/10/commits/5f644d3a857874236d27ab2e0615846184f9fc8c.

src/charm.py
marceloneppel: I changed the name of the method and its description to make it more clear. Thanks for pointing this Mia. Changed on https://github.com/canonical/postgresql-operator/pull/10/commits/0d03e9f03e3cb51d1e2cd7353ff315489c110b3f. It is needed to get only the ips of the units that are part of the cluster (the current members).

src/charm.py
marceloneppel: The current unit is being added in the first line (L91).

src/cluster.py
marceloneppel: Yeah, you are right, Paulo. I added some checks to handle that in the next PR commit (https://github.com/canonical/postgresql-operator/pull/11/commits/9939234da34968bc425dc7b2d41e76b47302724a).

src/charm.py
MiaAltieri: SNEAKY I did NOT see that

src/cluster.py
```diff
@@ -95,6 +122,62 @@ def _get_postgresql_version(self) -> str:
...
+        # Request info from cluster endpoint (which returns all members of the cluster).
+        cluster_status = requests.get(f"http://{self.unit_ip}:8008/cluster")
+        for member in cluster_status.json()["members"]:
+            if member["role"] == "leader":
+                primary = member["name"]
+                if unit_name_pattern:
+                    # Change the last dash to / in order to match unit name pattern.
+                    primary = "/".join(primary.rsplit("-", 1))
+                break
+        return primary
```
MiaAltieri, line 144: This might be cleaner. But tbh I don't really care, feel free to keep what you have
```suggestion
                    primary = "/".join(primary.rsplit("-", 1))
                    return primary
```

MiaAltieri (approved):
LTTM (looks terrific to me 🌟 )

src/cluster.py
marceloneppel: It looks cleaner. It just probably wouldn't work when there is no primary (it should happen only if there is a problem in a cluster, like a previous issue that I saw in the k8s charm when two clusters have the same name and all members become replicas). In that case it would return None (the default primary variable value) if the already implemented code is kept.

paulomach (approved):
lgtm

PR #11, opened 2022-05-05
Scaling and replication

marceloneppel:
# Issue
This is a PR that was on hold to copy the feedback that improved the PR from https://github.com/canonical/postgresql-k8s-operator/pull/9.

It is part of the scale up/down logic that is needed to avoid adding or removing more than one member to a PostgreSQL cluster if another member is still syncing data after it has joined.

In long:
* To avoid overload of the network during copy of data between the replicas that are being added to the cluster during a scale up or downtime in the case of a scale down, the charm needs to add and remove on member at a time from the cluster so.
* If the member is the current primary, we need to trigger a switchover before the unit is removed, to make avoid downtime in the cluster.

# Solution
* Wait for all units to be ready (with `running` state in the Patroni REST API) before adding one more unit to the cluster.
* Trigger a switchover if the departing is the current primary.
* Remove the departing units one at a time (it should be checked that all members are ready before removing the unit).

# Context
* The list of current IP addresses is tracked on application relation data to avoid reaching Patroni REST API every time this information is needed.
* Replicas are started pointing to the IP of the primary (to avoid they thinking they need to bootstrap a new separated cluster), when it's ok to add them to the cluster (one at a time).
* During scale up, the status of the new member is not checked, only the status of the current members (they must be running/ready). This is done this way because Patroni cannot be started without pointing a leader, otherwise it'll create a new cluster in the replica, and when started in this way it has some changed of the state, like stopped and creating replica, so for now it's better to not check these status, as they may conflict with the status that we have when it really becomes part of the existing cluster.
* Helper methods used by the methods implemented on this PR were created on https://github.com/canonical/postgresql-operator/pull/10.

# Testing
* Unit and integrations tests were updated to fix the error that happened with the new features.
* The new tests are implemented on https://github.com/canonical/postgresql-operator/pull/14.

# Release Notes
* Implement the correct scaling up and down mechanism and replication of data.


src/charm.py
```diff
@@ -75,6 +121,89 @@ def _change_primary(self) -> None:
...
+        """
+        try:
+            # Compare set of Patroni cluster members and Juju hosts
+            # to avoid the unnecessary reconfiguration.
+            if self._patroni.cluster_members == self._hosts:
+                return
+
+            logger.info("Reconfiguring cluster")
+            self.unit.status = MaintenanceStatus("reconfiguring cluster")
+            self.unit.status = MaintenanceStatus("reconfiguring cluster")
```
WRFitch, line 177: Duplicate line

tests/unit/test_charm.py
```diff
@@ -101,24 +126,36 @@ def test_on_leader_elected(self):
...
-        # Mock the generated superuser password.
+        # Mock the passwords.
         _get_postgres_password.return_value = "fake-postgres-password"
+        _replication_password.return_value = "fake-replication-password"
 
         # Mock cluster start success values.
         _bootstrap_cluster.side_effect = [False, True]
 
         # Test for a failed cluster bootstrapping.
+        # TODO: test replicas start.
```
WRFitch, line 157: Is this for this PR, or will it be resolved in a future one?

WRFitch (approved):
LGTM!

src/cluster.py
```diff
@@ -48,34 +44,54 @@ class SwitchoverFailedError(Exception):
...
         # Render the configuration files and start the cluster.
-        self._change_owner(STORAGE_PATH)
-        self._render_patroni_yml_file(
-            cluster_name, member_name, superuser_password, replication_password
-        )
+        self.configure_patroni_on_unit(replica)
+        return self.start_patroni()
+
+    def configure_patroni_on_unit(self, replica: bool = False):
+        """Configure Patroni (configuration files and service) on the unit."""
```
paulomach, line 88: nit: add arguments to docstring (not sure the effects of `replica` argument)

paulomach (approved):
Minor comments, lgtm

tests/unit/test_charm.py
marceloneppel: It is for the future. Added the JIRA issue in https://github.com/canonical/postgresql-operator/pull/11/commits/d2529905daa730a8960c697da611a2e396d8fcf7,

src/charm.py
marceloneppel: Good catch! Thanks. I fixed it in https://github.com/canonical/postgresql-operator/pull/11/commits/73adbd589e3c044a3b19bd846df129db14c68e4c.

src/cluster.py
marceloneppel: Added in https://github.com/canonical/postgresql-operator/pull/11/commits/d2910b257fdfe713c0aa918ef5c470f439751997.

src/charm.py
```diff
@@ -54,6 +67,39 @@ def _on_get_primary(self, event: ActionEvent) -> None:
         except RetryError as e:
             logger.error(f"failed to get primary with error {e}")
 
+    def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
+        """The leader removes the departing units from the list of cluster members."""
+        # Allow leader to update hosts if it isn't leaving.
+        if not self.unit.is_leader() or event.departing_unit == self.unit:
```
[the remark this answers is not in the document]
marceloneppel: If it happens, the list of members will be fixed in the leader elected event.

src/charm.py
```diff
@@ -42,9 +50,14 @@ def __init__(self, *args):
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.leader_elected, self._on_leader_elected)
         self.framework.observe(self.on.get_primary_action, self._on_get_primary)
+        self.framework.observe(self.on[PEER].relation_changed, self._on_peer_relation_changed)
+        self.framework.observe(self.on[PEER].relation_departed, self._on_peer_relation_departed)
+        self.framework.observe(self.on.pgdata_storage_detaching, self._on_pgdata_storage_detaching)
         self.framework.observe(self.on.start, self._on_start)
         self.framework.observe(self.on.get_initial_password_action, self._on_get_initial_password)
-        self._cluster = Patroni(self._unit_ip)
+        self._cluster_name = self.app.name
```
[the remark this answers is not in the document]
marceloneppel: At least for now, the fixed name based on the application name is not an issue. Maybe we can see in the future if there are use cases that need a random cluster name.

WRFitch (approved):

paulomach (approved):
lgtm

PR #12, opened 2022-07-04
Add legacy relations implementation

marceloneppel:
# Issue
This PostgreSQL charm should support legacy relations from the [original charm](https://launchpad.net/postgresql-charm).

# Solution
* Implement both legacy relations, db and db-admin, following the logic from the original charm.

# Context
* `metadata.yaml` defines both relations.
* `requirements.txt` adds an additional dependency used to make it easier to build PostgreSQL string that will be stored in the `master` and `standbys` fields of the relations databags.
* `src/charm.py` has calls to update the endpoints (master and standbys) in the legacy relations.
* `src/constants.py` contains the constants with the legacy relations names to be used in other files.
* `src/relations/db.py` implements the relation logic similar to what was done in the [k8s charm](https://github.com/canonical/postgresql-k8s-operator/blob/main/src/relations/db.py). The main difference is that in this charm we also need to put the relation data in the unit databag (in the k8s charm it is only needed to add the data in the application databag).
* `tests/unit/test_db.py` implements the tests for those relations methods.
* `tox.ini` only add a missing dependency for integration tests.
* Note: as in https://github.com/canonical/postgresql-operator/pull/17, there is a temporary workaround for https://bugs.launchpad.net/juju/+bug/1979811 in src/relations/db.py.

# Testing
* Unit tests for the relation code were added.
* Integration tests were implement in the following PRs:
  * https://github.com/canonical/postgresql-operator/pull/18
  * https://github.com/canonical/postgresql-operator/pull/19

# Release Notes
* Add legacy db and db-admin relations.


src/relations/db.py
```diff
@@ -0,0 +1,325 @@
...
+
+        try:
+            # Creates the user and the database for this specific relation if it was not already
+            # created in a previous relation changed event.
+            user = f"relation-{event.relation.id}"
+            password = unit_relation_databag.get("password", new_password())
+            self.charm.postgresql.create_user(user, password, self.admin)
+            self.charm.postgresql.create_database(database, user)
+            postgresql_version = self.charm.postgresql.get_postgresql_version()
+        except (
```
shayancanonical, line 118: is it better to explicitly create a helper that checks if a user and database exists instead of using `try/except`?

WRFitch (approved):

src/relations/db.py
marceloneppel: In this specific case, the errors wont happen if the user or database already exists. It is handled by the PostgreSQL library to check that and just [alter](https://github.com/canonical/postgresql-operator/blob/main/lib/charms/postgresql_k8s/v0/postgresql.py#L126) privileges in the user or database if they already exist. The errors will be raised if another unexpected thing happens.

paulomach (approved):
lgtm, tests are great.
This is a design choice, but i think it's semantically better to split the user adding for the relation into a `relation_joined` handler, and leave the databag updates to `config_changed` handler. That's a nit though, feel free to dismiss it.

marceloneppel:
Thanks @WRFitch and @paulomach for the suggestion. It seems good to separate those two steps and move one of them to the relation joined event handler. I'll evaluate it and probably open another PR to change that. Thanks again!

PR #15, opened 2022-07-18
Add new relation

marceloneppel:
# Issue
The new PostgreSQL VM charm should support a new relation following the pattern from [this relation interface](https://github.com/canonical/charm-relation-interfaces/tree/main/database).

# Solution
* Implement a new relation using the [data platform provides library](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/database_provides.py).

# Context
* `lib/charms/data_platform_libs/v0/database_provides.py` was imported from [Data Platform Libs](https://github.com/canonical/data-platform-libs) repository and don't need to be reviewed.
* `lib/charms/postgresql_k8s/v0/postgresql.py` was imported from the k8s charm and don't need to be reviewed.
* In `src/relations/postgresql_provider.py` the relation handlers are implemented. There is a `on database requested` handler instead of the often used `on relation changed` handler. It is using the Data Platform Libs provides library. There is also a `on relation broken` handler that removes the previously created database user when the relation is gone.
* `README.md` was updated with some initial instructions (the charm is not published yet).
* Relation names were also fixed.
* Some calls to `update_endpoints` are being made to handle the most part of primary changes and update the relation databags (automatic failover still needs an observer or some other mechanism to update the primary in the relation databags).

# Testing
* Unit tests for the postgresql provider were added.
* Existing tests were fixed.
* Integration tests were implement in the [next PR](https://github.com/canonical/postgresql-operator/pull/16).

# Release Notes
* Add new  PostgreSQL relation.

PR #16, opened 2022-07-18
New relations tests

marceloneppel:
# Issue
* The new PostgreSQL relation implemented in https://github.com/canonical/postgresql-operator/pull/15 needs integrations tests.

# Solution
* Replicate the tests already implemented in the k8s charm.

# Context
* Files copied from [Data Platform Libs](https://github.com/canonical/data-platform-libs) repository that don't need to be reviewed:
  * `tests/integration/new_relations/application-charm/*`

# Testing
* The integrations tests from `tests/integration/new_relations/test_new_relations.py` were copied from the [k8s charm](https://github.com/canonical/postgresql-k8s-operator/blob/main/tests/integration/new_relations/test_new_relations.py). Some things were modified:
  * The last two tests (`test_no_read_only_endpoint_in_standalone_cluster` and `test_read_only_endpoint_in_scaled_up_cluster`) were moved to the beggining of the file (aftet the deployment of the charms used in the test).
  * `test_no_read_only_endpoint_in_standalone_cluster` couldn't be run in the end because it needs a scale down to only 1 unit, which is not possible from 2 units using the raft mechanism from Patroni (for that to work, we would need an additional unit running Patroni raft controller instead of Patroni and PostgreSQL).

# Release Notes
* Add new  PostgreSQL relation integration tests.


tests/integration/new_relations/test_new_relations.py
```diff
@@ -0,0 +1,297 @@
...
+    # Connect to the database using the read-only endpoint.
+    with psycopg2.connect(connection_string) as connection, connection.cursor() as cursor:
+        # Read some data.
+        cursor.execute("SELECT data FROM test;")
+        data = cursor.fetchone()
+        assert data[0] == "some data"
+
+        # Try to alter some data in a read-only transaction.
+        with pytest.raises(psycopg2.errors.ReadOnlySqlTransaction):
+            cursor.execute("DROP TABLE test;")
```
MiaAltieri, line 156: very thorough

tests/integration/new_relations/test_new_relations.py
```diff
@@ -0,0 +1,297 @@
...
+    # Get the connection strings to connect to both databases.
+    first_database_connection_string = await build_connection_string(
+        ops_test, APPLICATION_APP_NAME, FIRST_DATABASE_RELATION_NAME
+    )
+    second_database_connection_string = await build_connection_string(
+        ops_test, APPLICATION_APP_NAME, SECOND_DATABASE_RELATION_NAME
+    )
+
+    # Assert the two application have different relation (connection) data.
+    assert first_database_connection_string != second_database_connection_string
```
MiaAltieri, line 297: might be worth adding a test that scales up/down PG and then verifies that the relation data is properly updated and can still perform CRUD ops etc. 

Specifically add 2 units, remove original units so that PG has none of the units it was originally deployed with. Then verify that application can still connect to PG.

^ this suggestion is not for the endpoint tests which you mention cannot do do this scaling due to patroni. But instead for ensuring that relation departed/joined events update the connection URI and other relation data. (Can be done in a future PR)

MiaAltieri (commented):
LGTM once those tests start passing, I have a suggestion regarding tests too :)

WRFitch (approved):

tests/integration/new_relations/test_new_relations.py
marceloneppel: Good suggestion, Mia. I implemented the new test (with fixes figured out when implementing it) in the [next PR](https://github.com/canonical/postgresql-operator/pull/17). It is still a draft because I am checking why scaling down works locally but not when the test is running on Github CI.

I also merged some missing changes from the previous PR and the tests of this PR as working now.

MiaAltieri (approved):

paulomach (approved):
very nice - certainly going to draw inspiration from this tests

PR #17, opened 2022-08-05
Relation data update tests

marceloneppel:
# Issue
There is a mising test for the new relations described in https://github.com/canonical/postgresql-operator/pull/16#discussion_r937627679:

> might be worth adding a test that scales up/down PG and then verifies that the relation data is properly updated and can still perform CRUD ops etc.
> 
> Specifically add 2 units, remove original units so that PG has none of the units it was originally deployed with. Then verify that application can still connect to PG.
> 
> ^ this suggestion is not for the endpoint tests which you mention cannot do do this scaling due to patroni. But instead for ensuring that relation departed/joined events update the connection URI and other relation data.

# Solution
* Add the requested integration test.
* After adding the test, I figured out that it was not working because of some details of the operator implementation. So, it was needed to fix them by adding some retries in the check of the primary of the cluster, specially after a failed switchover, when there may be some delay before electing a new primary (because of some configurations about the failover timeouts).
* Also, there was a problem related to the charm incorrectly deleting a user from a relation if it is scaled down (because the charm receives a relation broken event in that case and the user removal is done in that event handler). This was also fixed through a temporary workaround.

# Context
* `lib/charms/postgresql_k8s/v0/postgresql.py` has some changes to handle the correct deletion of users that have names containing hyphen and other character that need to be escaped before used in PostgreSQL statements.
* `src/charm.py` has some changes to get the primary endpoint with tenacity (sometimes after a failed switchover the cluster members list show all of them as replicas and no primary, which is solved after some seconds). Also, the update endpoints call is beind done only if there is a primary.
* `src/cluster.py' has additional checks for cluster members during a failed switchover scenario.
* `src/relations/postgresql_provider.py` was updated to have a temporary workaround for https://bugs.launchpad.net/juju/+bug/1979811. It does not solve all the situations that trigger that event, but at least the worst cases.
* `tests/integration/new_relations/test_new_relations.py` implements the test mentioned in the issue section of this PR.
* The unit tests from `tests/unit/test_charm.py` and `tests/unit/test_postgresql_provider.py` were updated based on the changes in the operator source code.
* Note: the failover timeouts were not reduced to avoid possible split brain scenarios. This should be improved in the future, to improve how fast the checks for a primary in the cluster are done.

# Testing
* Unit tests were updated.
* The additional integration test was implemented.

# Release Notes
* Fix primary endpoint retrieval.
* Fix endpoints update.
* Add integration test to check endpoints update after scaling up/down.

# Future work
* Try to improve the way failed switchover are handled.
* Sync failover timeouts with primary endpoint tenacity configurations.


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -150,12 +150,12 @@ def delete_user(self, user: str) -> None:
                 with self._connect_to_database(
                     database
                 ) as connection, connection.cursor() as cursor:
-                    cursor.execute(f"REASSIGN OWNED BY {user} TO postgres;")
-                    cursor.execute(f"DROP OWNED BY {user};")
+                    cursor.execute(sql.SQL("REASSIGN OWNED BY {} TO postgres;").format(sql.Identifier(user)))
```
WRFitch, line 153: nice

tests/integration/new_relations/test_new_relations.py
```diff
@@ -295,3 +295,66 @@ async def test_an_application_can_request_multiple_databases(ops_test: OpsTest,
...
+async def test_relation_data_is_updated_correctly_when_scaling(ops_test: OpsTest):
+    """Test that relation data, like connection data, is updated correctly when scaling."""
+    # Retrieve the list of current database unit names.
+    units_to_remove = [unit.name for unit in ops_test.model.applications[DATABASE_APP_NAME].units]
+
+    # Add two more units.
+    await ops_test.model.applications[DATABASE_APP_NAME].add_units(2)
+    await ops_test.model.wait_for_idle(
+        apps=[DATABASE_APP_NAME], status="active", timeout=1000, wait_for_exact_units=4
+    )
```
WRFitch, line 309: `ops_test.fast_forward` might help here

WRFitch (approved):

src/relations/postgresql_provider.py
```diff
@@ -109,10 +122,24 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
         if (
             "cluster_initialised" not in self.charm._peers.data[self.charm.app]
             or not self.charm._patroni.member_started
+            or not self.charm.primary_endpoint
         ):
             event.defer()
```
MiaAltieri, line 127: I assume you defer since you're waiting for one of these conditions to become true:
```
"cluster_initialised" not in self.charm._peers.data[self.charm.app]
or not self.charm._patroni.member_started
or not self.charm.primary_endpoint
```
In order for some of these to become true some other events need to run (ie the start hook) deferring here can cause this event to continuously defer making it take much longer for the start hook to get on the queue before this event re-queues. This can lead to downtime/poor performance. Instead I suggest you set a flag and trigger and event that will re-call `on_relation_broken`.

Mykola did something similar in his k8s charm [here](https://github.com/delgod/mongodb-k8s-operator/blob/ad178a7e2406f3b805ac73d5dfb7927b123c6deb/lib/charms/mongodb_libs/v0/mongodb_provider.py#L60-L63) and instead of re-triggering the event, he creating a function for himself to call once his condition was met by the start hook, see [here](https://github.com/delgod/mongodb-k8s-operator/blob/ad178a7e2406f3b805ac73d5dfb7927b123c6deb/src/charm.py#L142)

tests/integration/new_relations/test_new_relations.py
```diff
@@ -295,3 +295,66 @@ async def test_an_application_can_request_multiple_databases(ops_test: OpsTest,
...
+    connection.close()
+
+    # Remove the relation and test that its user was deleted
+    # (by checking that the connection string doesn't work anymore).
+    await ops_test.model.applications[DATABASE_APP_NAME].remove_relation(
+        f"{DATABASE_APP_NAME}:database", f"{APPLICATION_APP_NAME}:{FIRST_DATABASE_RELATION_NAME}"
+    )
+    await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=1000)
+    with pytest.raises(psycopg2.OperationalError):
+        psycopg2.connect(primary_connection_string)
```
MiaAltieri, line 360: AWESOME test

tests/unit/test_postgresql_provider.py
```diff
@@ -153,12 +174,37 @@ def test_on_database_requested(
...
+            RelationDepartedEvent, "departing_unit", new_callable=PropertyMock
+        ) as mock_departing_unit:
+            # Test that no flag is set in peer relation data if the remote unit
+            # is the unit that is being removed.
+            mock_departing_unit.return_value = self.harness.model.get_unit("application/0")
+            self.harness.remove_relation_unit(self.rel_id, "application/0")
+            peer_relation_data = self.harness.get_relation_data(self.peer_rel_id, self.unit)
+            self.assertDictEqual(peer_relation_data, {})
+
+            # Test that a flag for preventing user deletion is set in peer relation data
```
MiaAltieri, line 190: quick question do you have an integration test for verifying that your "flag preventing user deletion" works?

MiaAltieri (commented):
Looks great! Very thorough tests really awesome work! Just some suggestions :)

tests/unit/test_postgresql_provider.py
marceloneppel: Yes Mia! It is [here](https://github.com/canonical/postgresql-operator/blob/d139e248a5b490f442b0d9d1ecc10f7516cbd56a/tests/integration/test_db.py#L114) (part of the other PR: https://github.com/canonical/postgresql-operator/pull/18). When the two units are destroyed in that line, the deletion of the user would be triggered in them even we still have two more units in the relation with the application. So, with the flag, when the two units are destroyed, the user is not deleted (it is deleted when the relation is really broken between the database and application).

tests/integration/new_relations/test_new_relations.py
marceloneppel: Good suggestion. Added in https://github.com/canonical/postgresql-operator/pull/17/commits/391aaa27327ff9d768c2ca34b336fd6695b6fcd0.

src/charm.py
```diff
@@ -208,8 +224,19 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):
             event.defer()
             return
 
-        self.postgresql_client_relation.update_endpoints()
-        self.unit.status = ActiveStatus()
+        # If the unit is not the leader, just set an ActiveStatus.
+        if not self.unit.is_leader():
+            self.unit.status = ActiveStatus()
```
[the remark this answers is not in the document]
marceloneppel: Yeah. Currently it is safe to set this ActiveStatus (after a BlockedStatus is set it blocks the execution with a return or a missing piece that makes Patroni able to start, preventing the code to reach this point), but probably in the future, when new features are added and can generate additional BlockedStatus, this should be revisited.

tests/integration/new_relations/test_new_relations.py
marceloneppel: Thanks so much for the feedback!

src/relations/postgresql_provider.py
marceloneppel: Great suggestion Mia! I refactored the code a little bit in https://github.com/canonical/postgresql-operator/pull/17/commits/8bda284c7e8c7bc3d6631ed5aa3b7fc2858989c7 to initially move the user deletion logic to something similar to what you commented. In a future PR I will need to also move the user creation and update.

tests/unit/test_postgresql_provider.py
MiaAltieri: Awesome!!

MiaAltieri (approved):
Really awesome Marcelo!! 🤩

PR #18, opened 2022-08-11
Mailman3 Core tests

marceloneppel:
# Issue
* The legacy db relation implemented in https://github.com/canonical/postgresql-operator/pull/12 needs integrations tests.

# Solution
* Implement a test relating PostgreSQL with [Mailman3 Core charm](https://charmhub.io/mailman3-core).

# Context
* `tests/integration/conftest.py` contains a fixture to build the PostgreSQL charm that is used in this new and also in the existing integration tests.
* `tests/integration/helpers.py` has some helper functions to check whether databases and users were create after a relation is established, a function to get relation data related to connection to the database and also a function to deploy the application and relate it to PostgreSQL.
* `tests/integration/test_charm.py` had some small changes related to variable/constants and removing a fixture that was moved to tests/integration/conftest.py.
* `tests/integration/test_db.py` implements the tests that are described in the next section of this PR description.
* `tox.ini` has a new dependency needed to connect to the Mailman3 Core API.

# Testing
* The integration test that was implemented tests the relation between PostgreSQL and Mailman3 Core.
* Then it tests that Mailman3 Core is configured to use PostgreSQL instead of SQLite as the database.
* It also tests that, after the relation, it is possible to connect to Mailman3 Core API and do some CRUD operations.

# Release Notes
* Add  legacy db relation integration tests.


MiaAltieri (approved):
So thorough looks awesome Marcelo!! 🤩

minor suggestions only :)

WRFitch (approved):

shayancanonical (approved):

PR #21, opened 2022-08-16
Change charm database user

marceloneppel:
# Issue
* Database charms need a database user with a name that is easy to identify. This user shouldnt be used by charms related to the database charm.

# Solution
* Change the default `postgres` superuser to a user called `operator`.

# Context
* Changed the `postgres` superuser to `operator`. Patroni uses that for some operations too.
* Patroni uses another user for replication, so no issues because of changing the superuser. 
* This was already done in a similar way in https://github.com/canonical/postgresql-k8s-operator/pull/25.
* Changes in `lib/charms/postgresql_k8s/v0/postgresql.py` will be submitted to the k8s charm later.
* The default `postgres` user needs to be created because some applications (not charms), like [Landscape Server](https://bugs.launchpad.net/landscape/+bug/1987689), need that user in the database for some operations.
* The action used to retrieve the charm user password was renamed. That action will also be used to retrieve other system users passwords later.

# Testing
* Unit and integration tests were updated.
* Manually tested the charm to check that the right user is created inside PostgreSQL.

# Release Notes
* Change charm user to a user called `operator`.


WRFitch (approved):

shayancanonical (approved):
lgtm!

PR #23, opened 2022-08-25
Rework secrets

marceloneppel:
# Issue
* Secrets like passwords and TLS certificates should be moved to Juju secrets in the future.
* Currently, passwords are being set and retrieved from app relation data in multiple places, what can lead to a big refactor later (even more when we add TLS certificates to the charm).

# Solution
* Move secret implementation logic to specific functions that can be easily refactored in the future to use Juju secrets instead of app relation data.

# Context
* Unit peer relation property was also added for future use cases.
* Some constants were created to avoid the same strings in multiple places.
* This was already done in a very similar way in https://github.com/canonical/postgresql-k8s-operator/pull/27.
* Checking again [Will's comment](https://github.com/canonical/postgresql-k8s-operator/pull/27#discussion_r953201947) in the other PR, I saw that it makes sense for the other relation data that the charm uses (like list of endpoints). So an [issue](https://github.com/canonical/postgresql-operator/issues/24) was created to move all operations related to the peer relation to another file.

# Testing
* Additional unit tests were added.
* Tested using integration tests (to ensure the new implementation works the same way).

# Release Notes
* Add specific functions to set and get secrets.


WRFitch (approved):

paulomach (approved):
lgtm

PR #25, opened 2022-08-25
Password rotation

marceloneppel:
# Issue
* Database charms should enable the human operator to rotate any system user (users that are need for the correct operation of the charm and/or the workload, like a superuser, a replication user, etc.) password.

# Solution
* Create an action to set system users passwords and propagate that password to all the other units.

# Context
* This was already done in a very similar way in https://github.com/canonical/postgresql-k8s-operator/pull/26.
* Users created through relations (legacy or new relations) are not covered by this mechanism. Those users will have the password rotated after a sequence of a broken relation and a new relation being established.
* When updating a password using the `set-password` action, that password need to be update in the Patroni configuration (which also needs to be reloaded). The configuration is reloaded in the leader unit first and later on the other units (through relation changed event; that event already handles Patroni configuration changes and its reload process). There is no downtime in that processes.
* The new integration test from `tests/integration/test_password_rotation.py` rotate the two system users password and checks that they are correctly updated in all the units (which is checked after restarting Patroni; it would trigger an connection error in the Patroni process if the password is not updated).

# Testing
* Unit and integration tests were updated based in the new actions.
* Additional unit and integration tests were added to validate the password rotation mechanism.

# Release Notes
* Add password rotation mechanism for all system users (users used by the charm/workload and that were not created through relations).


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -196,3 +200,27 @@ def list_users(self) -> Set[str]:
...
+            username: the user to update the password.
+            password: the new password for the user.
+
+        Raises:
+            PostgreSQLUpdateUserPasswordError if the password couldn't be changed.
+        """
+        try:
+            with self._connect_to_database() as connection, connection.cursor() as cursor:
+                cursor.execute(
+                    sql.SQL("ALTER USER {} WITH ENCRYPTED PASSWORD '" + password + "';").format(
```
shayancanonical, line 217: minor suggestion (feel free to ignore): is it possible to use fstrings instead of `.format`?

shayancanonical (approved):
lgtm!!

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: `.format` is needed to correctly escape and handle invalid characters in the user name. But I can improve it to user a Placeholder object from psycopg2 to avoid using the plus signal to concat the password. I will do it in the library in the k8s repository along with other improvements and then update it here.

Thanks for the questions and the suggestion Shayan!

src/charm.py
```diff
@@ -612,8 +623,73 @@ def _on_start(self, event) -> None:
...
+        if username not in SYSTEM_USERS:
+            event.fail(
+                f"The action can be run only for users used by the charm:"
+                f" {', '.join(SYSTEM_USERS)} not {username}"
+            )
+            return
+
+        password = new_password()
+        if "password" in event.params:
+            password = event.params["password"]
```
zmraul, line 658: Similar to before:
```suggestion
        password = event.params.get("password", new_password())
```

zmraul (approved):
Looks good!

src/charm.py
marceloneppel: Thanks for this improvement, Raúl!

PR #28, opened 2022-09-06
TLS relations files

marceloneppel:
# Issue
* PostgreSQL charm needs to have a way to enable TLS in the connections between instances and from the client applications to the PostgreSQL server.

# Solution
* This PR starts to solve that issue by adding some files needed to interact with https://github.com/canonical/tls-certificates-operator (to later, in the [next PR](https://github.com/canonical/postgresql-operator/pull/30), request TLS certificates and load them).

# Context
* This PR consists on `REPLICATING` what was already reviewed on https://github.com/canonical/postgresql-k8s-operator/pull/35 to the PostgreSQL VM charm.
* Files that were `IMPORTED` and can be `IGNORED`:
  * lib/charms/rolling_ops/v0/rollingops.py
  * lib/charms/postgresql_k8s/v0/postgresql.py
  * lib/charms/postgresql_k8s/v0/postgresql_tls.py
  * lib/charms/tls_certificates_interface/v1/tls_certificates.py
* charmcraft.yaml and requirements.txt only contains dependencies needed for lib/charms/tls_certificates_interface/v1/tls_certificates.py.
* metadata.yaml contains two new relations that are needed (one for triggering rolling restarts after TLS is enabled/disabled and the other to request/renew TLS certificates).

# Testing
* Integration tests were added in the [third PR](https://github.com/canonical/postgresql-operator/pull/31) of this sequence.

# Release Notes
* Add TLS relations files (to handle TLS certificates request and rolling restarts).


lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -0,0 +1,194 @@
+# Copyright 2022 Canonical Ltd.
```
WRFitch, line 1: This library has been imported and can also be ignored, right?

WRFitch (approved):

lib/charms/postgresql_k8s/v0/postgresql_tls.py
marceloneppel: Yes, good catch! I forgot to add it to the pr description. Almost everything in the PR consists of imported files.

charmcraft.yaml
```diff
@@ -11,4 +11,9 @@ bases:
         channel: "20.04"
 parts:
   charm:
-    charm-binary-python-packages: [psycopg2-binary==2.9.3]
+    build-packages:
+      - libffi-dev
+      - libssl-dev
+      - rustc
+      - cargo
+    charm-binary-python-packages: [jsonschema, psycopg2-binary==2.9.3]
```
shayancanonical, line 19: thoughts on using `-` for a list (to stay consistent with the way we're listing build packages?

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -129,14 +137,18 @@ def create_user(
...
+                )
                 if extra_user_roles:
                     user_definition += f' {extra_user_roles.replace(",", " ")}'
                 if cursor.fetchone() is not None:
                     statement = "ALTER ROLE {}"
                 else:
                     statement = "CREATE ROLE {}"
-                cursor.execute(sql.SQL(statement + user_definition + ";").format(sql.Identifier(user)))
+                cursor.execute(
+                    sql.SQL(statement + user_definition + ";").format(sql.Identifier(user))
```
shayancanonical, line 150: minor nit: can we use f-strings here?

shayancanonical (approved):

charmcraft.yaml
marceloneppel: Good suggestions, it is better to stay consistent with that. Updated on https://github.com/canonical/postgresql-operator/pull/28/commits/790e369813b6d0b86e8fe418750f091bc2b7b911.

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: Yes, we can. Updated on https://github.com/canonical/postgresql-operator/pull/28/commits/790e369813b6d0b86e8fe418750f091bc2b7b911.

PR #29, opened 2022-09-07
fix departing unit subtleties

rwcarlsen:
ops 1.5.2 introduced a fix to the harness' handling for departing unit tracking for events.  This uncovered a subtle error in the tests here. remove_relation was triggering the "departing" relation data key to be inserted into the local charm's data because the local charm had been added to the relation units list (which is not really how that is intended to be used).  This was causing the delete_user method to be called earlier than intended and caused assertion failures in the on_relation_broken test.  The fix is to not add the local unit to the relation units.  But then the relation departed test that checks for the "departing" data key wasn't working.  So for now, we can add the local unit to the relation in that test only.

### DIFF
--- a/tests/unit/test_postgresql_provider.py
+++ b/tests/unit/test_postgresql_provider.py
@@ -41,7 +41,6 @@ class TestPostgreSQLProvider(unittest.TestCase):
         # Define some relations.
         self.rel_id = self.harness.add_relation(RELATION_NAME, "application")
         self.harness.add_relation_unit(self.rel_id, "application/0")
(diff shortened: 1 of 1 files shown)


marceloneppel (approved):

rwcarlsen:
Integration failure.  Seems unlikely to be caused by this (test code) change?

marceloneppel:
> Integration failure. Seems unlikely to be caused by this (test code) change?

Yeah, them definitely are not related to the change. I will investigate further what is causing that. Thanks for the fix @rwcarlsen!

PR #32, opened 2022-09-12
Install pgBackRest and add initial backup and rewind settings

marceloneppel:
# Issue
* Jira issue: [DPE-513](https://warthogs.atlassian.net/browse/DPE-513)
* To enable backups later in the PostgreSQL charm, a tool responsible for backups must first be installed.
* PostgreSQL needs some initial settings for backups and rewind operations (synchronisation of the data between a primary and the replicas after a replica is promoted to primary).
* We also need to ensure that a rewind operations (the ones using the [pg_rewind](https://www.postgresql.org/docs/current/app-pgrewind.html) tool) are done using connections encrypted by TLS.

# Solution
* Install [pgBackRest](https://pgbackrest.org).
* Add the following settings for backup operations:

  * achieve_mode=on

  * archive_command=/bin/true

  * wal_level = logical

* Add the following settings for rewind operations:

  * remove_data_directory_on_rewind_failure = true

  * remove_data_directory_on_diverged_timelines = true

Add a user for pg_rewind and test that it uses TLS when connecting to other PostgreSQL instance.

# Context
* This PR is almost a copy of https://github.com/canonical/postgresql-k8s-operator/pull/45.

* The main differences are:
   * A call to reinitialise the replica was added to `src/charm.py` (because at some points in some tests the replica became too far behind the timeline of the primary and this is the recommended way to solve that is to reinitialise the replica data directory).

   * On `src/cluster.py` a call to the Patroni API to retrieve the lag information. Also, timeouts were added to the API calls to avoid long requests that make the cluster slow, specially in the freeze db process test (that was failing sometimes).

# Testing
* Existing unit tests on `tests/unit/test_patroni.py` were updated to match the new settings.
* The extra backup and rewind settings are now added to the integration test that checks the current instance configuration.
* An additional check was added to ensure that `pg_rewind` is using TLS.

# Release Notes
* Install pgBackRest.
* Add initial settings for backup and rewind operations.
* Add user for pg_rewind.
* Add test to ensure pg_rewind is using TLS.

[DPE-513]: https://warthogs.atlassian.net/browse/DPE-513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/cluster.py
```diff
@@ -243,12 +254,36 @@ def member_started(self) -> bool:
...
         return r.json()["state"] == "running"
 
+    @property
+    def member_replication_lag(self) -> str:
+        """Member replication lag."""
+        try:
+            for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
+                with attempt:
+                    cluster_status = requests.get(
+                        f"{self._patroni_url}/cluster",
```
WRFitch, line 274: A lot of these URL magic strings are similar - can you extract them into variables please?

tests/integration/helpers.py
```diff
@@ -299,12 +320,27 @@ async def deploy_and_relate_bundle_with_postgresql(
     await ops_test.model.wait_for_idle(
         apps=[application_name],
         status="active",
-        timeout=1000,
+        timeout=2000,
```
WRFitch, line 323: This isn't really relevant to this ticket, but should adding a relation really take half an hour? It might be worth looking into why this takes so much time

WRFitch (approved):
A couple of nits, but looks great!

tests/integration/helpers.py
marceloneppel: Good catch! I think this change was made just for a local test and I forgot to remove it. Reverted on https://github.com/canonical/postgresql-operator/pull/32/commits/0bb1414c7b9f6f9584826bedba08f5a61fc51f74.

src/cluster.py
marceloneppel: Changed to a constant on [Extract endpoint from URL to constant](https://github.com/canonical/postgresql-operator/pull/32/commits/5adae46693502d3c53d20406138ba035c7836c81).

delgod:
> Cool! LGTM in general. Q: should we really commit "archive_command: /bin/true" in Patroni template? Why not pgbackrest?

it should be switched to pgBackRest only after initialization of the pgBackRest stanza (in other words after storage initialization).

delgod (approved):

PR #33, opened 2022-09-12
TLS on Patroni REST API

marceloneppel:
# Issue
* Jira issue: [DPE-512](https://warthogs.atlassian.net/browse/DPE-512)
* Both Patroni REST API (that is used by this charm for some operations, and also by Patroni itself) and the patronictl command line tool (tool used by Patroni for some operations; it uses Patroni REST API under the hood) need to have the communication to and from them being encrypted.

# Solution
* Enable TLS on both Patroni REST API and patronictl command line tool when TLS is enabled on PostgreSQL (what happens after a relations is established with the TLS Certificates Operator).

# Context
* This PR consists on `REPLICATING` what was already reviewed on https://github.com/canonical/postgresql-k8s-operator/pull/38 to the PostgreSQL VM charm.
* When TLS is enabled, the non encrypted endpoint is not available anymore; the opposite happens when TLS is disabled: the encrypted endpoint is not available anymore. So, with the current implementation, even withouth an integration test, the correct implementation of TLS on the API is validated (because the API is used in several events of the charm and any wrong call would make the charm not deploy correctly).
* Summary of the main changes (you can `FOCUS` your review on them):
  * `tests/integration/helpers.py` and `tests/integration/test_tls.py`: new checks to ensure the API is accepting TLS connections and differently from the k8s charm now we use the TLS CA to verify the server certificate.
* There is a small difference in the way TLS enabled/disabled is passed to  the Patrons object compared to the k8s charm (here peer relation data is used) due to some differences between both charms. This can be updated to be more similar in the future.

# Testing
* The integration test for TLS was updated to check that Patroni REST API has TLS enabled.

# Release Notes
* Add TLS encryption to Patroni REST API and patronictl command line tool.


tests/integration/helpers.py
```diff
@@ -490,6 +517,41 @@ async def check_tls(ops_test: OpsTest, unit_name: str, enabled: bool) -> bool:
...
+    Args:
+        ops_test: The ops test framework instance.
+        unit_name: The name of the unit where Patroni is running.
+        enabled: check if TLS is enabled/disabled
+
+    Returns:
+        Whether TLS is enabled/disabled on Patroni REST API.
+    """
+    unit_address = get_unit_address(ops_test, unit_name)
+    tls_ca = await get_tls_ca(ops_test, unit_name)
```
WRFitch, line 532: Could you return False here `if not tls_ca`?

WRFitch (approved):

tests/integration/helpers.py
marceloneppel: Yes, I'll add it.

src/cluster.py
```diff
@@ -74,6 +76,7 @@ def __init__(
             planned_units: number of units planned for the cluster
             superuser_password: password for the operator user
             replication_password: password for the user used in the replication
+            tls_enabled: whether TLS is enabled
```
MiaAltieri, line 79: nit add verify

tests/integration/helpers.py
```diff
@@ -490,6 +517,47 @@ async def check_tls(ops_test: OpsTest, unit_name: str, enabled: bool) -> bool:
...
+            stop=stop_after_attempt(10), wait=wait_exponential(multiplier=1, min=2, max=30)
+        ):
+            with attempt, tempfile.NamedTemporaryFile() as temp_ca_file:
+                # Write the TLS CA to a temporary file to use it in a request.
+                temp_ca_file.write(tls_ca.encode("utf-8"))
+                temp_ca_file.seek(0)
+
+                # The CA bundle file is used to validate the server certificate when
+                # TLS is enabled, otherwise True is set because it's the default value
+                # for the verify parameter.
```
MiaAltieri, line 550: excellent commenting

MiaAltieri (approved):
Looks awesome

tests/integration/helpers.py
marceloneppel: Adde on https://github.com/canonical/postgresql-operator/pull/33/commits/334a50d1da908dda09563dc10e5eeb70975e8135.

tests/integration/helpers.py
marceloneppel: Thanks Mia!

src/cluster.py
marceloneppel: This specific parameter is related to whether TLS is already enabled or not, different from the tests, where we have parameters related to verify whether TLS is enabled.

PR #35, opened 2022-09-21
Add kill db process test

marceloneppel:

PR #39, opened 2022-10-10
Freeze DB process test

marceloneppel:
# Issue
* Jira issues: [DPE-538](https://warthogs.atlassian.net/browse/DPE-538)
* The PostgreSQL charm should self heal the workload when freezing the DB process (and later unfreezing it).

# Solution
* Add a test that freezes the DB processes (from Patroni and from PostgreSQL), unfreezing it after some time and later check that the workload recovers itself from that situation.

# Context
* This test can also be run on an existing cluster like in the previous implemented test from https://github.com/canonical/postgresql-operator/pull/35.
* An mechanism to get the other instances REST API URL was created on `src/cluster.py` to enable the charm to find the primary (that is used to update the relation application databag with the primary and the replicas list) when the unit that had the process frozen is the leader.

# Testing
* The test was added on `tests/integration/ha_tests/test_self_healing.py`. It tests the charm in a very similar way the [Charmed MongoDB Operator](https://github.com/canonical/mongodb-operator/blob/5ce9e5e9e7cfeda9a060b0403c6c83c4771ec576/tests/integration/ha_tests/test_ha.py#L378) is tested.
* The test freezes both Patroni and PostgreSQL OS process, one at a time.

# Release Notes
* Add freeze db process test.

### DIFF
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -19,6 +19,7 @@ from charms.operator_libs_linux.v1.systemd import (
 )
 from jinja2 import Template
 from tenacity import (
+    AttemptManager,
     RetryError,
     Retrying,
     retry,
@@ -197,14 +198,14 @@ class Patroni:
                         break
         return primary
 
-    def _get_alternative_server_url(self, attempt) -> str:
+    def _get_alternative_server_url(self, attempt: AttemptManager) -> str:
+        """Get an alternative URL from another member each time."""
         if attempt.retry_state.attempt_number > 1:
             url = self._patroni_url.replace(
                 self.unit_ip, list(self.peers_ips)[attempt.retry_state.attempt_number - 2]
             )
         else:
             url = self._patroni_url
-        logger.warning(f"url for get primary: {url}")
         return url
 
     def are_all_members_ready(self) -> bool:
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -162,7 +162,6 @@ class Patroni:
         Returns:
             IP address of the cluster member.
         """
-        ip = None
         # Request info from cluster endpoint (which returns all members of the cluster).
         for attempt in Retrying(stop=stop_after_attempt(len(self.peers_ips) + 1)):
             with attempt:
@@ -170,9 +169,7 @@ class Patroni:
                 cluster_status = requests.get(f"{url}/cluster", verify=self.verify, timeout=10)
                 for member in cluster_status.json()["members"]:
                     if member["name"] == member_name:
-                        ip = member["host"]
-                        break
-                return ip
+                        return member["host"]
 
     def get_primary(self, unit_name_pattern=False) -> str:
         """Get primary instance.
@@ -183,7 +180,6 @@ class Patroni:
         Returns:
             primary pod or unit name.
         """
-        primary = None
         # Request info from cluster endpoint (which returns all members of the cluster).
         for attempt in Retrying(stop=stop_after_attempt(len(self.peers_ips) + 1)):
             with attempt:
@@ -195,8 +191,7 @@ class Patroni:
                         if unit_name_pattern:
                             # Change the last dash to / in order to match unit name pattern.
                             primary = "/".join(primary.rsplit("-", 1))
-                        break
-        return primary
+                        return primary
 
     def _get_alternative_patroni_url(self, attempt: AttemptManager) -> str:
         """Get an alternative REST API URL from another member each time.
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -166,7 +166,7 @@ class Patroni:
         # Request info from cluster endpoint (which returns all members of the cluster).
         for attempt in Retrying(stop=stop_after_attempt(len(self.peers_ips) + 1)):
             with attempt:
-                url = self._get_alternative_server_url(attempt)
+                url = self._get_alternative_patroni_url(attempt)
                 cluster_status = requests.get(f"{url}/cluster", verify=self.verify, timeout=10)
                 for member in cluster_status.json()["members"]:
                     if member["name"] == member_name:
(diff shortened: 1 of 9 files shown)


src/cluster.py
```diff
@@ -163,12 +164,15 @@ def get_member_ip(self, member_name: str) -> str:
...
-        return ip
+        for attempt in Retrying(stop=stop_after_attempt(len(self.peers_ips) + 1)):
+            with attempt:
+                url = self._get_alternative_server_url(attempt)
+                cluster_status = requests.get(f"{url}/cluster", verify=self.verify, timeout=10)
+                for member in cluster_status.json()["members"]:
+                    if member["name"] == member_name:
+                        ip = member["host"]
+                        break
+                return ip
```
MiaAltieri, line 175:
```suggestion
                    if member["name"] == member_name:
                        return member["host"]
```

src/cluster.py
```diff
@@ -181,16 +185,29 @@ def get_primary(self, unit_name_pattern=False) -> str:
...
+                url = self._get_alternative_server_url(attempt)
+                cluster_status = requests.get(f"{url}/cluster", verify=self.verify, timeout=10)
+                for member in cluster_status.json()["members"]:
+                    if member["role"] == "leader":
+                        primary = member["name"]
+                        if unit_name_pattern:
+                            # Change the last dash to / in order to match unit name pattern.
+                            primary = "/".join(primary.rsplit("-", 1))
+                        break
         return primary
```
MiaAltieri, line 199:
```suggestion
                        return primary
```

src/cluster.py
```diff
@@ -181,16 +185,29 @@ def get_primary(self, unit_name_pattern=False) -> str:
...
 
+    def _get_alternative_server_url(self, attempt: AttemptManager) -> str:
+        """Get an alternative URL from another member each time."""
```
MiaAltieri, line 202: what is an alternative URL and why is it needed?

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -83,3 +83,54 @@ async def test_kill_db_process(
...
+        for attempt in Retrying(stop=stop_after_delay(60 * 3), wait=wait_fixed(3)):
+            with attempt:
+                new_primary_name = await get_primary(ops_test, app)
+                assert new_primary_name != primary_name
+
+        # Un-freeze the old primary.
+        await send_signal_to_process(ops_test, primary_name, process, "SIGCONT")
+
+        # Verify that the database service got restarted and is ready in the old primary.
+        assert await postgresql_ready(ops_test, primary_name)
```
MiaAltieri, line 124: add checks that verify:
- that the old primary is now the secondary
- all units are in the same replica set

tests/unit/test_cluster.py
```diff
@@ -13,23 +16,88 @@
...
             "fake-superuser-password",
             "fake-replication-password",
             False,
         )
 
+    def test_get_alternative_server_url(self):
+        # Mock tenacity attempt.
+        retry = tenacity.Retrying()
+        retry_state = tenacity.RetryCallState(retry, None, None, None)
+        attempt = tenacity.AttemptManager(retry_state)
```
MiaAltieri, line 60: VERY COOL

shayancanonical (approved):
looks great!

tests/unit/test_cluster.py
```diff
@@ -40,6 +108,33 @@ def test_get_postgresql_version(self, _from_system):
...
+        # Test using the current server URL.
+        _get_alternative_server_url.side_effect = ["http://server1"]
+        primary = self.patroni.get_primary()
+        self.assertEqual(primary, "postgresql-0")
+
+        # Test requesting the primary in the unit name pattern.
+        _get_alternative_server_url.side_effect = ["http://server1"]
+        primary = self.patroni.get_primary(unit_name_pattern=True)
+        self.assertEqual(primary, "postgresql/0")
+
```
MiaAltieri, line 137: very thorough

MiaAltieri (commented):
wow so clean and very well implemented 🤩! Great job Marcelo! Minor comments only

src/cluster.py
marceloneppel: Sorry Mia. Now I added an improved description. The alternative server URL method is just a way to call the REST API from another cluster member when, for example, the current member DB process is frozen (not able to receive requests). In that case the API is still working in the other members, so the URL from one of them is used. Updated on https://github.com/canonical/postgresql-operator/pull/39/commits/2108c4aff35ceb5707f9f71c3666d45bd1f0e79c.

tests/integration/ha_tests/test_self_healing.py
marceloneppel: Great suggestion. Added both on https://github.com/canonical/postgresql-operator/pull/39/commits/6a1e39bf4d695470e165b6417d9e39ddc4e5f318. I also added those checks on the kill DB process test.

src/cluster.py
marceloneppel: Updated on https://github.com/canonical/postgresql-operator/pull/39/commits/29eae5b6bfaa92227a43f1e77569d9ade31559f5.

src/cluster.py
marceloneppel: Updated on https://github.com/canonical/postgresql-operator/pull/39/commits/29eae5b6bfaa92227a43f1e77569d9ade31559f5.

src/cluster.py
MiaAltieri: fascinating! Great descriptions thanks Marcelo! :)

tests/integration/ha_tests/helpers.py
```diff
@@ -96,19 +128,52 @@ async def get_master_start_timeout(ops_test: OpsTest) -> Optional[int]:
...
+                        role = member["role"]
+
+                # A member that restarted has the DB process stopped may
+                # take some time to know that a new primary was elected.
+                if role == "replica":
+                    return True
+                else:
+                    raise MemberNotUpdatedOnClusterError()
+    except RetryError:
+        return False
```
MiaAltieri, line 174: great function 🤩

MiaAltieri (approved):
really awesome work here Marcelo!

PR #43, opened 2022-10-18
SST test

marceloneppel:


tests/integration/ha_tests/test_self_healing.py
```diff
@@ -168,3 +170,83 @@ async def test_freeze_db_process(
...
+    await start_continuous_writes(ops_test, app)
+
+    # Change the "master_start_timeout" parameter to speed up the fail-over.
+    original_master_start_timeout = await get_master_start_timeout(ops_test)
+    await change_master_start_timeout(ops_test, 0)
+
+    # Copy data dir content removal script.
+    await ops_test.juju(
+        "scp", "tests/integration/ha_tests/clean-data-dir.sh", f"{primary_name}:/tmp"
+    )
```
shayancanonical, line 197: i believe that for this test, we need to remove the data directory while the postgress process is down (we need to extend the systemd restart timeout [like mongodb here](https://github.com/canonical/mongodb-operator/blob/main/tests/integration/ha_tests/helpers.py#L576))

an excerpt of a message from mykola:
```
for SST test (in MySQL):
1) stop pebble/systemd on one member, remove all files in /var/lib/mysql [data directory] (simulate HDD failure)
2) write data to new primary
3) rotate binlog and remove rotated binlog ON ALL alive members. literally remove data written on step 2 (with such we simulate looong period of downtime)
4) run mysql on member from step 1)
```

marceloneppel: Thanks for the additional details, Shayan! I'll update the PR to have a similar approach on PostgreSQL.

WRFitch (approved):

tests/integration/ha_tests/test_self_healing.py
marceloneppel: Hey @shayancanonical! I updated the code to have the right steps that simulate the needed scenario. I haven't changed the systemd service restart timeout as after stopping the service it was not being restarted until I request to start it again. I also added a check to ensure the WAL files (the equivalent to MysQL binlog) are correctly rotated (a new one is created - in fact, more than one new WAL file is kept due to some settings that enabled the old ones to be removed).

shayancanonical (approved):
Looks great!

PR #44, opened 2022-10-26
Fix TLS flag

marceloneppel:
# Issue
* Jira issue: [DPE-820](https://warthogs.atlassian.net/browse/DPE-820)
* A TLS flag was not set when the charm was not ready but is already related to the TLS certificates operator. This commonly happens when the charm is deployed in a bundle together with the TLS certificates operator.
* Also, another problem happened and was related to the charm sending a certificate signing request (CSR) with it's IP address as a DNS name instead of an IP address.

# Solution
* Update the code to correctly set the TLS flag when the workload hasn't started yet, but already has a relation with the TLS certificates operator.

# Context
* You can `FOCUS` on the following files when reviewing:
  * `src/charm.py`: has the change in the restart logic (to match the k8s charm) and also adds the missing change of the TLS flag (that is used internally by the charm to call the Patroni REST API using either HTTP or HTTPS).
  
  * `tests/integration/test_tls.py`: has the updates that Will made to improve the test (simulating what we have on a bundle deployment: no `wait_for_idle`).

* Other files and what has changed on them:

  * `.github/workflows/ci.yaml`: pinned the Juju Python library version due to https://bugs.launchpad.net/juju/+bug/1992833.

  * `lib/charms/postgresql_k8s/v0/postgresql_tls.py`: imported from the k8s charm (it contains the fix for the IP address issue) - it should be ignored in this PR and reviewed on https://github.com/canonical/postgresql-k8s-operator/pull/58 (every change there will be copied to this PR).

  * `lib/charms/tls_certificates_interface/v1/tls_certificates.py`: this is a patched version of the library that is used to solve the IP address issue. It was copied from https://github.com/canonical/opensearch-operator/blob/main/lib/charms/tls_certificates_interface/v1/tls_certificates.py (that patch will be pushed to the right repo in some weeks). Thanks @Mehdi-Bendriss!

# Testing
* The integration test was improved and two new units tests were added.

# Release Notes
* Fix TLS flag value change when the workload hasn't started yet.
* Fix IP address on certificate signing request.

Thanks @WRFitch for starting this fixes.

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql_tls.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql_tls.py
@@ -174,7 +174,7 @@ class PostgreSQLTLS(Object):
             try:
                 ipaddress.ip_address(address)
                 return True
-            except ValueError:
+            except (ipaddress.AddressValueError, ValueError):
                 return False
 
         unit_id = self.charm.unit.name.split("/")[1]
--- a/lib/charms/postgresql_k8s/v0/postgresql_tls.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql_tls.py
@@ -62,7 +62,7 @@ class PostgreSQLTLS(Object):
         super().__init__(charm, "client-relations")
         self.charm = charm
         self.peer_relation = peer_relation
-        self.additional_dns_names = [] if additional_dns_names is None else additional_dns_names
+        self.additional_dns_names = additional_dns_names or []
         self.certs = TLSCertificatesRequiresV1(self.charm, TLS_RELATION)
         self.framework.observe(
             self.charm.on.set_tls_private_key_action, self._on_set_tls_private_key
--- a/lib/charms/postgresql_k8s/v0/postgresql_tls.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql_tls.py
@@ -34,7 +34,7 @@ from cryptography import x509
 from cryptography.x509.extensions import ExtensionType
 from ops.charm import ActionEvent
 from ops.framework import Object
-from ops.pebble import PathError, ProtocolError
+from ops.pebble import ConnectionError, PathError, ProtocolError
 
 # The unique Charmhub library identifier, never change it
 LIBID = "c27af44a92df4ef38d7ae06418b2800f"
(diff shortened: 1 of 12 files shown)


lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -149,35 +153,57 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+        sans = [
             f"{self.charm.app.name}-{unit_id}",
+            self.charm.get_hostname_by_unit(self.charm.unit.name),
             socket.getfqdn(),
             str(self.charm.model.get_binding(self.peer_relation).network.bind_address),
         ]
+        sans.extend(self.additional_dns_names)
+
+        # Separate IP addresses and DNS names.
+        sans_ip = [san for san in sans if is_ip_address(san)]
```
Mehdi-Bendriss, line 192: Interesting approach

Mehdi-Bendriss (approved):
Very good work, Marcelo!

lib/charms/postgresql_k8s/v0/postgresql_tls.py
```diff
@@ -54,11 +55,14 @@
...
 
-    def __init__(self, charm, peer_relation):
+    def __init__(
+        self, charm, peer_relation: str, additional_dns_names: Optional[List[str]] = None
+    ):
         """Manager of PostgreSQL relation with TLS Certificates Operator."""
         super().__init__(charm, "client-relations")
         self.charm = charm
         self.peer_relation = peer_relation
+        self.additional_dns_names = [] if additional_dns_names is None else additional_dns_names
```
MiaAltieri, line 65:
```suggestion
        self.additional_dns_names = additional_dns_names or []
```

src/charm.py
```diff
@@ -833,6 +835,9 @@ def update_config(self) -> None:
         # Update and reload configuration based on TLS files availability.
         self._patroni.render_patroni_yml_file(enable_tls=enable_tls)
         if not self._patroni.member_started:
+            # If Patroni/PostgreSQL has not started yet and TLS relations was initialised,
+            # then mark TLS as enabled.
```
MiaAltieri, line 839:
```suggestion
            # then mark TLS as enabled. This commonly happens when the charm is deployed in a bundle together with the TLS certificates operator.
```

MiaAltieri (approved):
Looks great 2 minor comments

marceloneppel:
Thanks for the reviews, @Mehdi-Bendriss and @MiaAltieri!

PR #45, opened 2022-11-01
Full cluster restart and crash tests

marceloneppel:
# Issue
* Jira issues: [DPE-545](https://warthogs.atlassian.net/browse/DPE-545) and [DPE-550](https://warthogs.atlassian.net/browse/DPE-550)
* The PostgreSQL charm should self heal the workload when restarting the DB process.

# Solution
* Add a test that restarts (using SIGTERM) the Patroni OS processes and later check that the workload recovers itself from that situation.
* Add a test that crashes (using SIGKILL) the Patroni OS processes and later check that the workload recovers itself from that situation.

# Context
* Those tests can also be run on an existing cluster.
* Both tests were implemented in a single test (with the pkill signal as a test parameter).

# Testing
* The test was added on `tests/integration/ha_tests/test_self_healing.py`.

# Release Notes
* Add full cluster restart and crash tests.


tests/integration/ha_tests/test_self_healing.py
```diff
@@ -221,3 +225,107 @@ async def test_restart_db_process(
...
+
+    # Update all units to have a new RESTART_DELAY,  Modifying the Restart delay to 3 minutes
+    # should ensure enough time for all replicas to be down at the same time.
+    for unit in ops_test.model.applications[app].units:
+        await update_restart_delay(ops_test, unit, RESTART_DELAY)
+
+    # Restart all units "simultaneously".
+    await asyncio.gather(
+        *[
+            send_signal_to_process(ops_test, unit.name, process, kill_code="SIGKILL")
```
WRFitch, line 299: It looks like these two tests are very similar, apart from this kill_code. Could you parametrize this?

WRFitch (approved):

tests/integration/ha_tests/test_self_healing.py
marceloneppel: Good catch! I merged them in one single test (parametrized).

tests/integration/ha_tests/helpers.py
```diff
@@ -25,7 +31,37 @@ class MemberNotUpdatedOnClusterError(Exception):
...
+
+async def all_db_processes_down(ops_test: OpsTest, process: str) -> bool:
+    """Verifies that all units of the charm do not have the DB process running."""
+    app = await app_name(ops_test)
+
+    try:
+        for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
+            with attempt:
+                for unit in ops_test.model.applications[app].units:
+                    search_db_process = f'run --unit {unit.name} ps ax | grep "{process} "'
```
shayancanonical, line 49: minor nit: could probably replace `ps ax | grep <process>` with `pgrep <process`

shayancanonical (approved):

PR #47, opened 2022-11-30
Fixes for machines restart

marceloneppel:
# Issue
Jira issue: [DPE-908](https://warthogs.atlassian.net/browse/DPE-908)

When the host VM is rebooted sometimes at least one of the PostgreSQL units get in an error state. The problem happens due to the Patroni service not being restarted correctly and the storage not being mounted (due to https://bugs.launchpad.net/juju/+bug/1999758).

# Solution
Enable the Patroni service to start on reboot and restart the machine if the storage is not correctly attached (the same that was done on https://github.com/canonical/mysql-operator/pull/74).

# Context
The Patroni systemd service was enabled to start at boot on `src/cluster.py`.

On `src/charm.py` a check for the storage was added (`_is_storage_attached `) and a call to reboot the machine (`_reboot_on_detached_storage `). Also, some logic was extracted from the `_on_start` method due to some additional fixes needed in the status of the unit.

# Testing
Tested manually by deploying the charm along with the TLS certificates operator and issued `sudo reboot`. Later checked that everything is working correctly, including the access to the workloads.

Also, added one optional test (`test_restart_machine ` that can be executed by running `RESTART_MACHINE_TEST=true tox -e tls-integration`), that restarts the unit's machine (which also leads to this issue), but is not configured to run on CI due to the issue from 
https://bugs.launchpad.net/juju/+bug/1999758 not being reproducible all the times - even if retrying multiple times in the code - and that the test would fail when the issue is fixed on Juju.

# Release Notes
Fix workload restart when host VM is restarted.

[DPE-908]: https://warthogs.atlassian.net/browse/DPE-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Install dependencies
         run: python3 -m pip install tox
       - name: Run linters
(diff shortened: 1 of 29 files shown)


taurus-forever (approved):
Wow! Nice!

dragomirp (approved):

PR #50, opened 2023-01-09
DPE 1083: Prevent relation extensions

dragomirp:
# Issue
* [DPE-1083](https://warthogs.atlassian.net/browse/DPE-1083)
* The charm should block in case extensions are enabled by a legacy relation

# Solution
* Changed where the charm was looking for if extensions were enabled

# Context
* Essentially the same as the fix for [DPE-1034](https://warthogs.atlassian.net/browse/DPE-1034)
* The Postgresql charm should block in case relation extensions are enabled, but it was looking for them in the relation databags, not in the event data
* Also updated action versions

# Testing

# Release Notes
Fixes bug in preventing relation extensions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Install dependencies
         run: python3 -m pip install tox
       - name: Run linters
@@ -20,13 +20,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Install dependencies
         run: python -m pip install tox
       - name: Run tests
         run: tox -e unit
       - name: Upload Coverage to Codecov
-        uses: codecov/codecov-action@v2
+        uses: codecov/codecov-action@v3
 
   integration-test-lxd-charm:
     name: Integration tests for charm deployment (lxd)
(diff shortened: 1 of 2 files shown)


tests/integration/test_db.py
```diff
@@ -156,3 +158,39 @@ async def test_relation_data_is_updated_correctly_when_scaling(ops_test: OpsTest
...
+            channel="edge",
+            application_name="nextcloud",
+            num_units=APPLICATION_UNITS,
+        )
+        await ops_test.model.wait_for_idle(
+            apps=["nextcloud"],
+            status="blocked",
+            raise_on_blocked=False,
+            timeout=1000,
+        )
```
dragomirp, line 178: This is a third party charm requesting extensions.

dragomirp:
Codecov seems to complain that the patch doesn't change the coverage. How should we deal with such cases?

taurus-forever:
> How should we deal with such cases?

@marceloneppel ? IMHO, enforcing test coverage is not something we should push _now_, IMHO.

taurus-forever (approved):
Tnx!

tests/integration/test_db.py
marceloneppel: Awesome finding!

marceloneppel (approved):
It looks great! @taurus-forever, I agree about not enforcing the coverage now. And also because IMO the specific issue we see in this PR may be a bug on codecov or its action.

In this case it commented `The diff coverage is 0.00%.` but it passed when it commented `The diff coverage is n/a.` (on https://github.com/canonical/postgresql-operator/pull/43#issuecomment-1282332593). I know we have a minor difference between both PRs, but it seems to be a bug on codecov.

PR #55, opened 2023-01-24
Enable full cluster crash and restart test and changes to somewhat stablise the test suite

dragomirp:
# Issue
* [DPE-825](https://warthogs.atlassian.net/browse/DPE-825)
* Full cluster crash and restart tests for Postgresql process

# Solution
* Enable full cluster crash and restart tests for Postgresql process
* Try to make the HA suite more durable

# Context
* Made the cleanup of continuous_writes to retry since it sometimes failed due to read only transaction
* Terminating continuous_writes instead of killing it, so that it will complete the last execution before exiting
* Adding failover when both primary and sync standby are gone (expected behaviour by `charm-integration` suite)

# Testing
* Full cluster crash and restart tests for Postgresql process

# Release Notes

[DPE-825]: https://warthogs.atlassian.net/browse/DPE-825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


requirements.txt
```diff
@@ -1,5 +1,5 @@
 cryptography==37.0.4
```
dragomirp, line 1: I think updating cryptography breaks pip and the installation of Patroni, so I've only updated ops for the time being.

tests/integration/ha_tests/application-charm/src/continuous_writes.py
```diff
@@ -29,7 +37,7 @@ def continuous_writes(connection_string: str, starting_number: int):
         connection.close()
 
     # Continuously write the record to the database (incrementing it at each iteration).
-    while True:
+    while run:
```
dragomirp, line 40: Break the loop when getting a SIGTERM.

tests/integration/ha_tests/application-charm/src/continuous_writes.py
```diff
@@ -53,6 +61,9 @@ def continuous_writes(connection_string: str, starting_number: int):
 
         write_value += 1
 
+    with open("/tmp/last_written_value", "w") as fd:
+        fd.write(str(write_value - 1))
```
dragomirp, line 65: Write the results in a file, for the charm to read.

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -46,14 +47,20 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+                charm, resources={"patroni": "patroni.tar.gz"}, num_units=3
+            )
+            await ops_test.juju("attach-resource", APP_NAME, "patroni=patroni.tar.gz")
+            await ops_test.model.wait_for_idle(status="active", timeout=1000)
+    # Deploy the continuous writes application charm if it wasn't already deployed.
+    if await app_name(ops_test, APPLICATION_NAME) is None:
+        async with ops_test.fast_forward():
+            charm = await ops_test.build_charm("tests/integration/ha_tests/application-charm")
+            await ops_test.model.deploy(charm, application_name=APPLICATION_NAME)
+            await ops_test.model.wait_for_idle(status="active", timeout=1000)
```
dragomirp, line 63: Moved this out of the fixture, so that the suite will fail from the get go if the app charm fails to build. When in the fixture, tests will continue to execute even if the suite cannot be successful.

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -236,7 +243,7 @@ async def test_restart_db_process(
 
 
 @pytest.mark.ha_self_healing_tests
-@pytest.mark.parametrize("process", [PATRONI_PROCESS])
+@pytest.mark.parametrize("process", DB_PROCESSES)
```
dragomirp, line 246: Enable the restart crash tests for both Patroni and Postgress.

tests/unit/test_charm.py
```diff
@@ -325,7 +325,6 @@ def test_on_start_no_patroni_member(
         patroni,
         _postgresql,
     ):
-
```
dragomirp, line 328: Linting fails otherwise.

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -94,10 +107,12 @@ def _count_writes(self) -> int:
...
-        with psycopg2.connect(
-            self._connection_string
-        ) as connection, connection.cursor() as cursor:
-            cursor.execute("DROP TABLE continuous_writes;")
+        for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
+            with attempt:
+                with psycopg2.connect(
+                    self._connection_string
+                ) as connection, connection.cursor() as cursor:
+                    cursor.execute("DROP TABLE continuous_writes;")
```
dragomirp, line 115: Sometimes tests failed due to this happening in a read only transaction. Retrying to make certain the table will be dropped.

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -128,33 +143,26 @@ def _start_continuous_writes(self, starting_number: int) -> None:
...
+            for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(5)):
                 with attempt:
-                    with psycopg2.connect(
-                        self._connection_string
-                    ) as connection, connection.cursor() as cursor:
-                        cursor.execute("SELECT MAX(number) FROM continuous_writes;")
-                        last_written_value = int(cursor.fetchone()[0])
-                    connection.close()
+                    with open(LAST_WRITTEN_FILE, "r") as fd:
+                        last_written_value = int(fd.read())
```
dragomirp, line 163: Read what continuous_writes reported instead of querying the DB.

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -128,33 +143,26 @@ def _start_continuous_writes(self, starting_number: int) -> None:
...
 
         # Stop the process.
-        proc = subprocess.Popen(["pkill", "--signal", "SIGKILL", "-f", "src/continuous_writes.py"])
-
-        # Wait for process to be killed.
-        proc.communicate()
+        os.kill(int(self.app_peer_data[PROC_PID_KEY]), signal.SIGTERM)
 
-        self._stored.continuous_writes_pid = None
+        del self.app_peer_data[PROC_PID_KEY]
```
dragomirp, line 156: `os.kill()` failed when using the stored state PID, so copied over the logic from the mongodb-k8s and mysql-k8s charms. This might have also caused issues with the suite, if the PID being killed was wrong.

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -88,17 +101,17 @@ def _count_writes(self) -> int:
         ) as connection, connection.cursor() as cursor:
             cursor.execute("SELECT COUNT(number) FROM continuous_writes;")
             count = cursor.fetchone()[0]
-        connection.close()
```
dragomirp, line 91: With should be closing the connection anyway.

tests/integration/ha_tests/test_self_healing.py
marceloneppel: Nice! Just as a suggestion: having only one call to `await ops_test.model.wait_for_idle(status="active", timeout=1000)` at the end outside both `if` blocks may speedup this part of the tests.

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -128,33 +141,26 @@ def _start_continuous_writes(self, starting_number: int) -> None:
...
-                        self._connection_string
-                    ) as connection, connection.cursor() as cursor:
-                        cursor.execute("SELECT MAX(number) FROM continuous_writes;")
-                        last_written_value = int(cursor.fetchone()[0])
-                    connection.close()
+                    with open(LAST_WRITTEN_FILE, "r") as fd:
+                        last_written_value = int(fd.read())
         except RetryError as e:
-            logger.exception(e)
+            logger.exception("Unable to query the database", exc_info=e)
```
marceloneppel, line 163: Shouldn't we change this message to something related to reading the file instead of querying the database?

tests/integration/ha_tests/application-charm/src/charm.py
marceloneppel: Great idea!

tests/integration/ha_tests/application-charm/src/charm.py
marceloneppel: I remembered about a note saying `leaving contexts doesn't close the connection` on https://www.psycopg.org/docs/connection.html#connection.

marceloneppel (commented):
Great updates on the code, Dragomir! I left a comment about the connections not being closed when exiting the `with` blocks.

marceloneppel (approved):
Super! LGTM!

taurus-forever (approved):
Nice work. I like it. You still need to make CI happy. LGTM!

templates/patroni.yml.j2
```diff
@@ -48,11 +48,15 @@ bootstrap:
     loop_wait: 10
     retry_timeout: 10
     maximum_lag_on_failover: 1048576
+    synchronous_mode: true
+    synchronous_mode_strict: false
```
dragomirp, line 52: Strict mode doesn't seem to work with a single unit while doing rolling ops. We'll need to enable/disable based on the size of the cluster.

tests/integration/ha_tests/helpers.py
```diff
@@ -73,6 +73,37 @@ async def app_name(ops_test: OpsTest, application_name: str = "postgresql") -> O
...
+        with attempt:
+            has_standby = False
+            cluster = get_patroni_cluster(unit_ip)
+            for member in cluster["members"]:
+                print
+                if member["role"] == "sync_standby":
+                    has_standby = True
+                    break
+            if not has_standby:
+                raise ValueError("No standby")
```
dragomirp, line 99: Faking strict synchronous mode, once the cluster is deployed. Should be removed once we have proper logic for that in the charm.

tests/integration/ha_tests/helpers.py
```diff
@@ -128,29 +159,40 @@ async def change_wal_settings(
...
-        for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
-            with attempt:
-                with psycopg2.connect(
-                    connection_string
-                ) as connection, connection.cursor() as cursor:
-                    cursor.execute("SELECT COUNT(number) FROM continuous_writes;")
-                    count = cursor.fetchone()[0]
-                connection.close()
-    except RetryError:
-        return -1
```
dragomirp, line 153: This is usually called within a retrying loop already.

tests/integration/ha_tests/helpers.py
```diff
@@ -263,7 +305,7 @@ def is_replica(ops_test: OpsTest, unit_name: str) -> bool:
 
                 # A member that restarted has the DB process stopped may
                 # take some time to know that a new primary was elected.
-                if role == "replica":
+                if role != "leader":
```
dragomirp, line 308: Could also be a sync standby as well.

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -72,12 +88,23 @@ def _on_start(self, _) -> None:
         """Only sets an Active status."""
         self.unit.status = ActiveStatus()
 
-    def _on_database_created(self, _) -> None:
-        """Event triggered when a database was created for this application."""
-        self._start_continuous_writes(1)
-
```
dragomirp, line 78: We are already starting manually in every test.

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -72,12 +88,23 @@ def _on_start(self, _) -> None:
...
+
+        if not self.app_peer_data.get(PROC_PID_KEY):
+            return None
+
+        with open(CONFIG_FILE, "w") as fd:
+            fd.write(self._connection_string)
+            os.fsync(fd)
+
+        try:
+            os.kill(int(self.app_peer_data[PROC_PID_KEY]), signal.SIGKILL)
```
dragomirp, line 104: Tried SIGHUPing the script, however `psycopg2` hangs the session if connection is lost during execution (freeze postgesql process test) and I was unable to figure out the configuration necessary to make it timeout. Haven't tried with `psycopg3`, it might be worthwhile to investigate.

tests/integration/ha_tests/conftest.py
```diff
@@ -21,23 +21,18 @@
...
+        action = (
+            await ops_test.model.applications[APPLICATION_NAME]
+            .units[0]
+            .run_action("clear-continuous-writes")
+        )
+        await action.wait()
 
 
-@pytest.fixture()
+@pytest.fixture(scope="module")
```
dragomirp, line 35: We are using it for all tests anyway.

templates/patroni.yml.j2
delgod: I think `synchronous_mode_strict` should be `false`.
The reason: we still want to have cluster work when the charm shuts down or restarts standbys.

dragomirp: > I think `synchronous_mode_strict` should be `false`. The reason: we still want to have cluster work when the charm shuts down or restarts standbys.

I've been testing the HA suite mostly with strict mode enabled, since I have encountered data loss without it. I will retry to see if the latest changes would be at least somewhat stable without it and report back, but I think that Patroni offers the most guarantees on data consistency with strict mode enabled. @taurus-forever @marceloneppel What do you think the correct way to go would be here?

marceloneppel: Dragomir commented with me on MM about trying to do a failover on the observer process that we recently added to the charm to fix other issues. It may help. If not, I think we should investigate a little bit deeper one specific situation that I found before synchronous mode was enabled and maybe persists now when strict mode is not enabled and lead to the failures Dragomir commented.

On the Patroni [docs](https://patroni.readthedocs.io/en/latest/replication_modes.html#synchronous-mode) we can see `When it is absolutely necessary to guarantee that each write is stored durably on at least two nodes, enable synchronous_mode_strict in addition to the synchronous_mode. This parameter prevents Patroni from switching off the synchronous replication on the primary when no synchronous standby candidates are available.`.

It’s not clear to me if Patroni turns off the synchronous replication due to no replicas or if it’s due to another issue. The situation that I commented above and happened before turning synchronous mode on was that it seemed that there was no synchronous replica even when we requested synchronous replication to be enabled.

In summary, I think we can research if synchronous replication is really working on synchronous mode without strict mode and fix if there is any issue. I saw some other operators not using strict mode, so it seems to be worth to try to avoid it. One last option would be to enable strict mode (if synchronous replications is working, but for some reason it still loses transactions - I think it should’t happen) and add some code to disable it in the observer (this will increase the complexity of the charm, of course), but this is only after we confirm that synchronous replication is working.

src/charm.py
```diff
@@ -520,6 +523,26 @@ def _on_cluster_topology_change(self, _):
...
+        if (
+            "cluster_initialised" not in self._peers.data[self.app]
+            or not self._patroni.member_started
+        ):
+            logger.debug("Early exit _on_cluster_topology_failover: awaiting for cluster to start")
+            return
+
+        logger.info("No primary or standby detected. Failing over")
+
+        self._patroni.failover()
```
dragomirp, line 545: To restore functionality expected by the `charm-integration` suite

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -72,12 +89,23 @@ def _on_start(self, _) -> None:
...
-        """Event triggered when a database was created for this application."""
-        self._start_continuous_writes(1)
-
     def _on_endpoints_changed(self, _) -> None:
         """Event triggered when the read/write endpoints of the database change."""
+        if self._connection_string is None:
+            return
+
+        if not self.app_peer_data.get(PROC_PID_KEY):
+            return None
```
marceloneppel, line 98: Do you really need to return `None` at this point?

marceloneppel (approved):
Looks amazing! Great work, Dragomir!

tests/integration/ha_tests/application-charm/src/charm.py
dragomirp: If there's no `continuous_writes` running, because it was stopped or because it has crashed, we shouldn't be restarting it.

marceloneppel (approved):

taurus-forever (approved):
This is just superb!

src/cluster.py
```diff
@@ -400,6 +381,21 @@ def switchover(self) -> None:
...
+    def failover(self) -> None:
+        """Trigger a failover."""
+        # Try to trigger the failover.
+        for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
+            with attempt:
+                r = requests.post(
+                    f"{self._patroni_url}/failover",
+                    json={"candidate": self.member_name},
+                    verify=self.verify,
+                )
```
delgod, line 393: Can we please set `synchronous_node_count` to minority, instead of promoting unsync replica as primary from the charm code?
```
synchronous_node_count = self.app.planned_units() // 2
```

delgod (changes requested):

src/cluster.py
dragomirp: Unfortunately, it doesn't seem like a minority count of sync standbys solves the instability in the `charm-integration` test. Since this PR has taken a while, I propose reverting to the default `synchronous_node_count` (1) and without automatic failover, break up `test_scale_down_and_up` into smaller tests and marking the parts that break as `unstable` until we have more clarity about how we expect things to behave. @marceloneppel @taurus-forever Would that be satisfactory for everyone?

src/charm.py
```diff
@@ -227,6 +224,8 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
             event.defer()
             return
 
+        self._patroni.update_synchronous_node_count(len(self._units_ips))
```
marceloneppel, line 227: `UpdateSyncNodeCountErrorr` (or the `RetryError`) is not being handled here and in the other call.

marceloneppel (approved):
Looks excellent!

taurus-forever (approved):
Nice!

tests/integration/ha_tests/application-charm/src/continuous_writes.py
```diff
@@ -17,22 +37,14 @@ def continuous_writes(connection_string: str, starting_number: int):
...
-        connection.close()
+    read_config_file()
 
     # Continuously write the record to the database (incrementing it at each iteration).
-    while True:
+    while run:
         try:
             with psycopg2.connect(connection_string) as connection, connection.cursor() as cursor:
                 connection.autocommit = True
+                cursor.execute("SET synchronous_commit = 'remote_apply';")
```
delgod, line 47: I think this line is not needed in the test 😞 it should be controlled on postgresql side 😞

tests/integration/ha_tests/application-charm/src/charm.py
delgod: feel free to ignore
```suggestion
            return
```

tests/integration/ha_tests/application-charm/src/charm.py
```diff
@@ -91,18 +119,64 @@ def _count_writes(self) -> int:
...
+            event.set_results({"result": "False"})
+            logger.exception("Unable to stop writes to drop table", exc_info=e)
+            return
+
+        try:
+            with psycopg2.connect(
+                self._connection_string
+            ) as connection, connection.cursor() as cursor:
+                cursor.execute("DROP TABLE IF EXISTS continuous_writes;")
+                event.set_results({"result": "True"})
```
delgod, line 140: feel free to ignore
```suggestion
            event.set_results({"result": "True"})
```

delgod (changes requested):

delgod (approved):

PR #53, opened 2023-01-17
Updating libjuju to latest version for 2

dragomirp:
# Issue
* [DPE-1081](https://warthogs.atlassian.net/browse/DPE-1081)
* Old lib juju is causing warnings in integration tests execution

# Solution
* update to latest version of libjuju for 2 (2.9.38.1)

# Context
* Only one test seemed to break
* Also updating flake8 to the latest version expected to work
* Also updating ops requirement to latest stable (1.5.4)

# Testing

# Release Notes

[DPE-1049]: https://warthogs.atlassian.net/browse/DPE-1049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DPE-1081]: https://warthogs.atlassian.net/browse/DPE-1081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #57, opened 2023-01-24
Fix unit tests

dragomirp:
# Issue
* [PR#53](https://github.com/canonical/postgresql-operator/pull/53) silently broke the unit tests due to running CI on older ops version (1.5.4)

# Solution
* Fix issues resulting from the ops update

# Context
* One of the unit tests added in [PR#53](https://github.com/canonical/postgresql-operator/pull/53) doesn't seem to be possible anymore with the test harnesss relation broken functionality. It may be possible to reproduce it using mock events and calling the `_on_relation_broken()` directly
* Also updating postgres_tls

# Testing

# Release Notes

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -250,11 +250,8 @@ class DbProvides(Object):
             self.charm._has_blocked_status
             and self.charm.unit.status.message == EXTENSIONS_BLOCKING_MESSAGE
         ):
-            if "extensions" in event.relation.data.get(
-                event.app, {}
-            ) or "extensions" in event.relation.data.get(event.unit, {}):
-                if not self._check_for_blocking_relations(event.relation.id):
-                    self.charm.unit.status = ActiveStatus()
+            if not self._check_for_blocking_relations(event.relation.id):
+                self.charm.unit.status = ActiveStatus()
 
     def update_endpoints(self, event: RelationChangedEvent = None) -> None:
         """Set the read/write and read-only endpoints."""
--- a/tests/unit/test_db.py
+++ b/tests/unit/test_db.py
@@ -42,7 +42,6 @@ class TestDbProvides(unittest.TestCase):
         # Define some relations.
         self.rel_id = self.harness.add_relation(RELATION_NAME, "application")
         self.harness.add_relation_unit(self.rel_id, "application/0")
-        self.harness.add_relation_unit(self.rel_id, self.unit)
         self.peer_rel_id = self.harness.add_relation(PEER, self.app)
         self.harness.add_relation_unit(self.peer_rel_id, self.unit)
         self.harness.update_relation_data(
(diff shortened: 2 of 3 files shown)


src/relations/db.py
```diff
@@ -250,11 +250,8 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
             self.charm._has_blocked_status
             and self.charm.unit.status.message == EXTENSIONS_BLOCKING_MESSAGE
         ):
-            if "extensions" in event.relation.data.get(
-                event.app, {}
-            ) or "extensions" in event.relation.data.get(event.unit, {}):
```
dragomirp, line 255: ops 1.5.4 test harness complains if we access the relationship databag during `_on_relation_broken`. Since we enter this block only when we block due to enabled extensions, it should be OK to remove.

tests/unit/test_db.py
```diff
@@ -42,7 +42,6 @@ def setUp(self):
         # Define some relations.
         self.rel_id = self.harness.add_relation(RELATION_NAME, "application")
         self.harness.add_relation_unit(self.rel_id, "application/0")
-        self.harness.add_relation_unit(self.rel_id, self.unit)
```
dragomirp, line 45: Causes warnings when executing the unit tests.

tests/unit/test_db.py
```diff
@@ -229,52 +228,16 @@ def test_on_relation_broken_extensions_unblock(
             postgresql_mock.delete_user = PropertyMock(return_value=None)
             self.harness.model.unit.status = BlockedStatus("extensions requested through relation")
             with self.harness.hooks_disabled():
-                self.rel_id = self.harness.add_relation(RELATION_NAME, "application")
```
dragomirp, line 232: Causes the relationship id to mismatch in the test harness.

tests/unit/test_db.py
```diff
@@ -229,52 +228,16 @@ def test_on_relation_broken_extensions_unblock(
...
-                self.harness.update_relation_data(
-                    second_rel_id,
-                    "application2",
-                    {"database": DATABASE, "extensions": ["test"]},
-                )
-
-            # Break the relation before the database is ready.
-            self.harness.remove_relation(self.rel_id)
-            self.assertTrue(isinstance(self.harness.model.unit.status, BlockedStatus))
-
```
dragomirp, line 277: Unfortunately it seems like this kind of test is no longer possible, since a [runtime error](https://github.com/canonical/operator/blob/1.5.4/ops/testing.py#L1446) will be raised by the test harness even if we are accessing remote side data for other relations.

tests/unit/test_db.py
taurus-forever: Just curios: is it a relation to itself? o_0

tests/unit/test_db.py
taurus-forever: Up 2 @marceloneppel but it looks strange. Is it runtime error in `ops`?

taurus-forever (commented):
1) we need to fix tests for sure
2) deleting a valuable test part is strange... can we address (or at least report and workaround) runtime error?

tests/unit/test_db.py
dragomirp: I guess it was added to have both a local and remote units in the relation, but the harness method seems to be for simulating a remote unit only.

dragomirp:
>     2. deleting a valuable test part is strange... can we address (or at least report and workaround) runtime error?
The test was added with [PR#53](https://github.com/canonical/postgresql-operator/pull/57) which broke the unit tests, so it is a new test. I think I should be able to rewrite the test by calling `_on_relation_broken()` directly instead of relying on the test harness, but that may take some time and slow down fixing the unit tests on main.

tests/unit/test_db.py
dragomirp: Readded the test.

marceloneppel (approved):
LGTM! I agree that we should keep the test, just fixing it. Thanks for fixing it @dragomirp! It seems that the ops code has some divergence with its own test code. We need to understand better and possibly open a bug.

taurus-forever (approved):
Yeah! A way better!

PR #58, opened 2023-01-25
Add observer that updates endpoints on relation

marceloneppel:
# Issue
Jira ticket: [DPE-790](https://warthogs.atlassian.net/browse/DPE-790)
When the PostgreSQL charm is related to the Landscape charm and the PostgreSQL primary is stopped, the switchover happens but the relation data is not updated (it's only updated on update status hook, but it may not happen as fast as needed), making Landscape break.

# Solution
Add an observer process like the one developed on https://github.com/canonical/observability-libs/pull/10 to detect when the roles of the cluster members changed and change the primary and replicas in the relation data.

# Context
Some fixes were needed in the call to the Patroni REST API URL to retry in the http or https endpoint when the current leader unit has it's DB process frozen (like in the freeze DB process test), because only the leader can update the endpoints in the relation data, but when it has it's DB process frozen it needs to check other members REST API to get the primary (and they may be using TLS or not - different members can have different TLS settings while the cluster is restarting to enable TLS).

Also, the login on the db relation changed a bit to make it possible that a PostgreSQL replica is able to update the endpoints in the relation data when it's the leader unit and erase it's unit data (to avoid confusing other charms using the legacy relation).

I'll open a bug on the Landscape charm because it fails when the PostgreSQL primary machine is stopped (even when using the legacy PostgreSQL charm), but it works well again when a switchover is made (automatically or manually - it was not working before).

A method called `_update_relation_endpoints ` was created to encapsulate alll the calls relating to update endpoints in the relations.

# Testing
Some new checks were added to `tests/integration/test_db_admin.py`.

# Release Notes
Add observer that updates endpoints on relation

[DPE-790]: https://warthogs.atlassian.net/browse/DPE-790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -20,13 +20,15 @@ from ops.charm import (
     RelationDepartedEvent,
 )
 from ops.framework import Object
-from ops.model import BlockedStatus, Relation, Unit
+from ops.model import ActiveStatus, BlockedStatus, Relation, Unit
 from pgconnstr import ConnectionString
 
 from constants import DATABASE_PORT
 from utils import new_password
 
 logger = logging.getLogger(__name__)
+
+EXTENSIONS_BLOCKING_MESSAGE = "extensions requested through relation"
 
 
 class DbProvides(Object):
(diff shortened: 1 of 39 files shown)


src/relations/db.py
```diff
@@ -134,6 +134,12 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
             # created in a previous relation changed event.
             user = f"relation-{event.relation.id}"
             password = unit_relation_databag.get("password", new_password())
+
+            # Store the user, password and database name in the secret store to be accessible by
+            # non-leader units when the cluster topology changes.
+            self.charm.set_secret("app", user, password)
+            self.charm.set_secret("app", f"{user}-database", database)
```
taurus-forever, line 141: I thought secrets are available for Juju 3.0+, TIL.

taurus-forever (approved):
Great approach!

src/relations/db.py
dragomirp: > I thought secrets are available for Juju 3.0+, TIL.

Those are convenience wrapper methods within the charms, that currently just put things into databags. In the future they will handle whether to use secrets for juju 3 or databags for juju 2

dragomirp (approved):

PR #67, opened 2023-02-14
Upgrade to jammy

marceloneppel:
# Issue
Jira issue: https://warthogs.atlassian.net/browse/DPE-1413
The charm is still on Focal.

# Solution
Upgrade all the bases and all the deployments to use Jammy.

# Context
The Landscape bundle test was disabled due to incompatibilities. Ticket to manage those incompatibilities: https://warthogs.atlassian.net/browse/DPE-1414

A lot of changes in the tests related to the database app name constant are due to removing multiple series in the tests.

# Testing
Used the existing tests to confirm that the charm works on Jammy.

# Release Notes
Upgrade charm to Jammy.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,7 @@ jobs:
           - charm-integration
           - database-relation-integration
           - db-relation-integration
-          - db-admin-relation-integration
+#          - db-admin-relation-integration
           - ha-self-healing-integration
           - password-rotation-integration
           - tls-integration
--- a/tests/integration/new_relations/test_new_relations.py
+++ b/tests/integration/new_relations/test_new_relations.py
@@ -10,7 +10,7 @@ import pytest
 import yaml
 from pytest_operator.plugin import OpsTest
 
-from tests.integration.helpers import scale_application
+from tests.integration.helpers import CHARM_SERIES, scale_application
 from tests.integration.new_relations.helpers import (
     build_connection_string,
     check_relation_data_existence,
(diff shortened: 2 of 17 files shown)


.github/workflows/ci.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
           - charm-integration
           - database-relation-integration
           - db-relation-integration
-          - db-admin-relation-integration
+#          - db-admin-relation-integration
```
dragomirp, line 51: Might be better to mark it as `unstable`

dragomirp (approved):

.github/workflows/ci.yaml
marceloneppel: I tried with it before, but the test failed. I think it's because there is only one test in that file.

taurus-forever: @carlcsaposs-canonical can you please check this case and improve unstable concept... commenting tests makes them invisible. 

@marceloneppel please create a bugreport ticket to re-enable the test. So we can schedule it in a pulse.

tests/integration/new_relations/test_new_relations.py
```diff
@@ -42,20 +42,21 @@ async def test_deploy_charms(ops_test: OpsTest, application_charm, database_char
...
                 application_name=APPLICATION_APP_NAME,
                 num_units=2,
+                series=CHARM_SERIES,
             ),
             ops_test.model.deploy(
                 database_charm,
                 resources={"patroni": "patroni.tar.gz"},
                 application_name=DATABASE_APP_NAME,
                 num_units=1,
-                trust=True,
```
taurus-forever, line 51: JFMI, why trust is no longer necessary?

taurus-forever (approved):
Thank you!

.github/workflows/ci.yaml
marceloneppel: Created: https://warthogs.atlassian.net/browse/DPE-1414

tests/integration/new_relations/test_new_relations.py
marceloneppel: It was not needed before. Probably it was incorrectly copied from the k8s charm tests.

PR #81, opened 2023-03-14
Use charmed postgresql snap

dragomirp:
# Issue
* [DPE-1424](https://warthogs.atlassian.net/browse/DPE-1424)
* Switch the charm to use charmed-postgresql snap
* [DPE-1514](https://warthogs.atlassian.net/browse/DPE-1514)
* Setting the workload version

# Solution

# Context
* `apt` and `systemd` libraries are replaced with `snap`
* Tweaks had to be made to the HA suite
* Setting the workload version based on the snapped postgresql version

# Testing

# Release Notes
* Start using the charmed-postgresql snap

[DPE-1424]: https://warthogs.atlassian.net/browse/DPE-1424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DPE-1514]: https://warthogs.atlassian.net/browse/DPE-1514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #93, opened 2023-03-29
Enable landscape scalable bundle test

marceloneppel:
# Issue
Jira ticket: [DPE-1414](https://warthogs.atlassian.net/browse/DPE-1414)
The Landscape tests was marked as unstable because the version that was tested was not compatible with SCRAM authentication (the default on PostgreSQL 14 - the version we got when switched to jammy on https://github.com/canonical/postgresql-operator/pull/67).

# Solution
Use the latest revision of the Landscape charm (through the Landscape Scalable bundle) and adjust the tests.

# Context
* In `src/charm.py`
  * The added APT packages are for Landscape to work (they are replaced by the ones from the snap on https://github.com/canonical/postgresql-operator/pull/81).
  * `_on_config_changed ` was removed to avoid users adding packages to the charm unit.
* In `src/relations/db.py`:
  * If no database name is provided, create one using the application name (this is needed for latest Landscape charm).
  * Update connection data in all the units' databags (remove it from the application databag, because it breaks Landscape).
  * Move the connection data to the update endpoints method, to update the data also in the replicas.
* In `tests/integration/helpers.py`:
  * Fixed some code to be able to deploy the latest revision of the bundle.
  * Added a helper function to get the Landscape admin user credentials (in the latest revision we don't have the charm action anymore, so it's needed to get those credentials from the database).
* In `tests/integration/test_db.py`:
  * Fixed the code to correctly handle the timing of some events (like the removal of the database user, that is being done in the `oversee_users` method called on the update status hook, which takes some time).
* In `tests/integration/test_db_admin.py`:
  * Fixed the test to match the behaviour from the latest revision of the bundle/charm.

# Testing
Fixed the existing tests (unit and integration) to match the new implementation.

# Release Notes
Enable landscape scalable bundle test.

[DPE-1414]: https://warthogs.atlassian.net/browse/DPE-1414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/relations/db.py
```diff
@@ -159,15 +157,9 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
         # is connecting to this database will receive a "database gone" event from the
         # old PostgreSQL library (ops-lib-pgsql) and the connection between the
         # application and this charm will not work.
-        allowed_subnets = self._get_allowed_subnets(event.relation)
-        allowed_units = self._get_allowed_units(event.relation)
-        for databag in [application_relation_databag, unit_relation_databag]:
+        for databag in [unit_relation_databag]:
```
dragomirp, line 160: I don't think we need to loop here any more.

marceloneppel: Good catch. I forgot to remove it. I'm going to fix it.

tests/unit/test_charm.py
```diff
@@ -691,5 +703,5 @@ def test_on_cluster_topology_change_clear_blocked(
 
         _update_relation_endpoints.assert_called_once_with()
         _update_certificate.assert_called_once_with()
-        _primary_endpoint.assert_called_once_with()
+        self.assertEqual(_primary_endpoint.call_count, 2)
```
dragomirp, line 706:
```suggestion
        self.assertEqual(_primary_endpoint.call_count, 2)
        _primary_endpoint.assert_called_with()
```

dragomirp (approved):
LGTM. Added a few nitpicks.

delgod (approved):

PR #95, opened 2023-03-30
Enable backup on replicas and hide replica from PgBouncer while doing backup

marceloneppel:
# Issue
Jira ticket: [DPE-1271](https://warthogs.atlassian.net/browse/DPE-1271)

It's not possible to run backups on replicas, only in the primary.

Also, when performing a backup, the unit is accessible by PgBouncer and other applications (those who can face performance issues while the backup is being created).

# Solution
Start a snap service that runs `pgbackrest server`, which is a TLS server that is used by the replicas to communicate with pgBackRest in the primary and request that missing WAL files are sent to the repository (S3) when the replica starts a backup process (so it backups the most updated data).

Also, while doing a backup in a replica, create some `pg_hba` reject rules to drop new connections to the replica.

This is the same that was done on https://github.com/canonical/postgresql-k8s-operator/pull/114.

# Context
The pgBackRest server is a requirement to backup from replicas. If TLS is not enabled, it's allowed to backup only from the primary.

In `src/backups.py`:
* `_is_primary_pgbackrest_service_running ` is needed to confirm that the primary unit started the TLS server because in the replicas the service will fail if it cannot connect to the primary TLS server.
* `_change_connectivity_to_database ` changes the connectivity to the database from external host (pg_hba reject rule).
* `start_stop_pgbackrest_service ` starts or stops the pgbackrest TLS server, depending on whether TLS is enabled (relation to TLS certificates operator) and there are replicas.

In `templates/pgbackrest.conf.j2`:
* Some TLS server related configs were added.

In `tests/integration/test_tls.py`:
* TLS test was enabled again.

# Testing
Changed the existing backup and restore integration test to backup from the replica.

Tested manually the connection (refused) to the replica while doing the backup on it.

# Release Notes
Enable backup on replicas.
Hide replica from PgBouncer while doing backup.

[DPE-1271]: https://warthogs.atlassian.net/browse/DPE-1271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/backups.py
```diff
@@ -573,6 +605,40 @@ def _retrieve_s3_parameters(self) -> Tuple[Dict, List[str]]:
...
+            return False
+
+        snap_cache = snap.SnapCache()
+        charmed_postgresql_snap = snap_cache["charmed-postgresql"]
+        if not charmed_postgresql_snap.present:
+            logger.error("Cannot start/stop service, snap is not yet installed.")
+            return False
+
+        # Stop the service if TLS is not enabled or there are no replicas.
+        if not self.charm.is_tls_enabled or len(self.charm._peer_members_ips) == 0:
```
delgod, line 630: do we need pgbackrest running if cluster size==1?
We should allow backups if it is 1 member cluster

delgod (approved):

src/backups.py
marceloneppel: No, we don't need. The service is needed only when the user wants to backup from a replica (cluster size > 1).

PR #114, opened 2023-04-14
Test network cut

marceloneppel:
## Issue
Jira ticket: [DPE-541](https://warthogs.atlassian.net/browse/DPE-541)
PostgreSQL should self heal after cutting the network connection and reestablishing it.

## Solution
* Add a test that adds a eth0 device to the unit to block the netwrok connection to and from a PostgreSQL unit and later restore the network traffic also checking that no writes (done by a continuous writes application) were missed.

In `src/charm.py`:
* Added logic to update the Patroni cluster with the new IP of the unit that was previously isolated from the cluster. 

In `tests/integration/ha_tests/conftest.py`:
* Added loop wait Patroni setting fixture to fix instability on some tests (specially full cluster restart and crash tests).

In `tests/integration/ha_tests/helpers.py`:
* Create the `is_cluster_updated ` helper function to avoid duplicate of the logic from the checks that are needed most of the tests. Improving the checks on https://github.com/canonical/postgresql-operator/pull/116 as this current PR became huge.
* Also added the helper functions needed to perform the network cut operation.

In `tests/integration/ha_tests/test_self_healing.py`:
* Fixed some instabilities in the other tests and added the network cut test.

[DPE-541]: https://warthogs.atlassian.net/browse/DPE-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


renovate.json
```diff
@@ -24,6 +24,9 @@
     }, {
       "matchPackageNames": ["juju"],
       "allowedVersions": "<3.0.0"
+    }, {
+      "matchPackageNames": ["packaging"],
+      "allowedVersions": "<23.0.1"
```
dragomirp, line 29: I don't think we need this if we are pinning in `pyproject.toml`

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -252,16 +232,21 @@ async def test_full_cluster_restart(
...
+        assert await are_all_db_processes_down(
+            ops_test, process
+        ), "Not all units down at the same time."
+    finally:
+        if process == PATRONI_PROCESS:
+            awaits = []
+            for unit in ops_test.model.applications[app].units:
+                awaits.append(update_restart_condition(ops_test, unit, ORIGINAL_RESTART_CONDITION))
+            await asyncio.gather(*awaits)
+        await change_patroni_setting(ops_test, "loop_wait", initial_loop_wait)
```
dragomirp, line 245: Shouldn't the fixture clear this either way?

dragomirp (approved):

renovate.json
marceloneppel: I'm going to remove it. Thanks for catching it.

delgod (approved):

PR #122, opened 2023-04-24
Dpe 549 full recreate cluster

dragomirp:


tests/integration/ha_tests/helpers.py
```diff
@@ -551,6 +561,7 @@ async def reused_storage(ops_test: OpsTest, unit_name) -> bool:
     await run_command_on_unit(
         ops_test,
         unit_name,
-        "grep 'Database cluster state: in archive recovery' /var/snap/charmed-postgresql/common/var/log/patroni/patroni.log",
+        "grep -E 'Database cluster state: in archive recovery|Database cluster state: shut down' "
```
dragomirp, line 564: The first cluster's primary seems to be in `Database cluster state: shut down`

tests/integration/ha_tests/test_restore_cluster.py
```diff
@@ -0,0 +1,130 @@
...
+
+charm = None
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest) -> None:
+    """Build and deploy three unit of PostgreSQL."""
+    wait_for_apps = False
+    # It is possible for users to provide their own cluster for HA testing. Hence, check if there
+    # is a pre-existing cluster.
```
dragomirp, line 39: This is a potentially destructive test, does this assumption hold true here as well?

marceloneppel (approved):
LGTM!

tests/integration/ha_tests/test_restore_cluster.py
delgod: let's do not allow running this test with existing installations.

tests/integration/ha_tests/test_restore_cluster.py
```diff
@@ -0,0 +1,130 @@
...
+            await ops_test.model.wait_for_idle(status="active", timeout=1000)
+
+
+async def test_cluster_restore(ops_test):
+    """Recreates the cluster from storage volumes."""
+    app = await app_name(ops_test)
+    if storage_type(ops_test, app) == "rootfs":
+        pytest.skip(
+            "re-use of storage can only be used on deployments with persistent storage not on rootfs deployments"
+        )
```
delgod, line 63: Maybe we should skip earlier, before the first cluster creation

delgod (changes requested):

tests/integration/ha_tests/helpers.py
dragomirp: I tried checking based on role, but it was unstable in CI, so I split the functions based on usage.

delgod (approved):

marceloneppel (approved):

PR #124, opened 2023-04-25
Add network cut test (without unit IP change)

marceloneppel:
## Issue
Jira ticket: [DPE-541](https://warthogs.atlassian.net/browse/DPE-541)
PostgreSQL should self heal after cutting the network connection and reestablishing it (when the network cut doesn't cause a change in the unit IP address).

## Solution
Limit the network connection of the unit to 0 kbit of egress traffic and 1 kbit of ingress traffic (the minimum allowed), and by reducing the priority of the traffic sent to the unit to the lowest.

In `tests/integration/ha_tests/helpers.py`:
* Added the helper functions needed to perform the network cut/restore operations (they don't cause an unit IP change).

In `tests/integration/ha_tests/test_self_healing.py`:
* Added the network cut test (the one where the unit IP doesn't change).

[DPE-541]: https://warthogs.atlassian.net/browse/DPE-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/ha_tests/helpers.py
+++ b/tests/integration/ha_tests/helpers.py
@@ -215,6 +215,26 @@ def cut_network_from_unit(machine_name: str) -> None:
     subprocess.check_call(cut_network_command.split())
 
 
+def cut_network_from_unit_without_ip_change(machine_name: str) -> None:
+    """Cut network from a lxc container (without causing the change of the unit IP address).
+
+    Args:
+        machine_name: lxc container hostname
+    """
+    override_command = f"lxc config device override {machine_name} eth0"
+    try:
+        subprocess.check_call(override_command.split())
+    except subprocess.CalledProcessError:
+        # Ignore if the interface was already overridden.
+        pass
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.egress=0kbit"
+    subprocess.check_call(limit_set_command.split())
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.ingress=1kbit"
+    subprocess.check_call(limit_set_command.split())
+    limit_set_command = f"lxc config set {machine_name} limits.network.priority=10"
+    subprocess.check_call(limit_set_command.split())
+
+
 async def fetch_cluster_members(ops_test: OpsTest):
     """Fetches the IPs listed by Patroni as cluster members.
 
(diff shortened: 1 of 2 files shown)


dragomirp (approved):

tests/integration/ha_tests/helpers.py
```diff
@@ -215,6 +215,26 @@ def cut_network_from_unit(machine_name: str) -> None:
...
+    """
+    override_command = f"lxc config device override {machine_name} eth0"
+    try:
+        subprocess.check_call(override_command.split())
+    except subprocess.CalledProcessError:
+        # Ignore if the interface was already overridden.
+        pass
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.egress=0kbit"
+    subprocess.check_call(limit_set_command.split())
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.ingress=1kbit"
```
delgod, line 232: maybe
```suggestion
    limit_set_command = f"lxc config device set {machine_name} eth0 limits.ingress=0kbit"
```

delgod, line 232: maybe
```suggestion
    limit_set_command = f"lxc config device set {machine_name} eth0 limits.ingress=0kbit"
```

tests/integration/ha_tests/helpers.py
```diff
@@ -498,6 +518,22 @@ def restore_network_for_unit(machine_name: str) -> None:
...
+    Args:
+        machine_name: lxc container hostname
+    """
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.egress="
+    subprocess.check_call(limit_set_command.split())
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.ingress="
+    subprocess.check_call(limit_set_command.split())
+    limit_set_command = f"lxc config set {machine_name} limits.network.priority="
+    subprocess.check_call(limit_set_command.split())
+    override_command = f"lxc config device remove {machine_name} eth0"
```
delgod, line 533: hm why we remove interface?

delgod (changes requested):

tests/integration/ha_tests/helpers.py
marceloneppel: For ingress, 0kbit is not allowed. 1kbit is the minimum allowed value. I did some manual tests (along with the automated test from the PR) and the connections to the database or Patroni fail at the other side (other unit or client application) as the response to the connections is not returned due.

marceloneppel: Same as above.

tests/integration/ha_tests/helpers.py
marceloneppel: Hmm, it's not needed. It's something that I used to test only. I'll remove it.

marceloneppel: Removed on https://github.com/canonical/postgresql-operator/pull/124/commits/4233128a2b838a3f42326e102af5ad984fc8d8ed.

PR #158, opened 2023-06-12
[DPE-1915] Backup and restore log messages

marceloneppel:
## Issue
We are not logging the success or failure of both the `create-backup` and `restore` actions in a way that is easy for other applications to parse the logs related to success or failure on those operations.

## Solution
Copied from https://github.com/canonical/postgresql-k8s-operator/pull/181.

Log failures (as errors) and successes (as info) to the debug log. The failures have prefix `Backup failed`/`Restore failed`, the successes have prefix `Backup successful`/`Restore successful`.

The `Restore successful` message can be logged only after Patroni starts (which happens some time after the execution of the `restore` action).


taurus-forever (approved):
Nice!

dragomirp (approved):

PR #167, opened 2023-06-20
Internal changes to the charm shouldn't be released

dragomirp:
## Issue
Charm is published to charmhub on each push to main even if there are no changes to the charm itself

## Solution
Try to limit the amount of releases based on the content of the push

### DIFF
--- a/tests/integration/test_db.py
+++ b/tests/integration/test_db.py
@@ -288,6 +288,7 @@ async def test_sentry_db_blocked(ops_test: OpsTest, charm: str) -> None:
         )
 
 
+@pytest.mark.unstable
 async def test_weebl_db(ops_test: OpsTest, charm: str) -> None:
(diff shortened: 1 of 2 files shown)


tests/integration/test_db.py
```diff
@@ -288,6 +288,7 @@ async def test_sentry_db_blocked(ops_test: OpsTest, charm: str) -> None:
         )
 
 
+@pytest.mark.unstable
 async def test_weebl_db(ops_test: OpsTest, charm: str) -> None:
```
dragomirp, line 292: A new weebl revision was released last week and it fails relating with our charms. Marking the test as unstable and created [DPE-2159](https://warthogs.atlassian.net/browse/DPE-2159) to investigate.

taurus-forever (approved):

marceloneppel (approved):

PR #169, opened 2023-06-22
[DPE-2090] Add aliases

dragomirp:
## Issue
Users don't know that psql and patronictl are installed together with the snap

## Solution
Adding aliases manually for the snapped application.

https://github.com/canonical/operator-libs-linux/pull/94 would allow us to do the same from the charm lib

Also updated charm libs


marceloneppel (approved):
Cool!

taurus-forever (approved):

PR #170, opened 2023-06-22
[DPE-2090] aliases with lib

dragomirp:
## Issue
Lib feature to create aliases was merged

## Solution
Switch alias creation to use the lib


marceloneppel (approved):
Nice!

PR #187, opened 2023-07-17
[DPE-1768] Minor version upgrades

marceloneppel:
## Issue
We cannot safely upgrade the charm.

## Solution
Implemented upgrade module with unit test coverage using the library from https://github.com/canonical/data-platform-libs/pull/77.

Due to https://github.com/pgbackrest/pgbackrest/issues/2007, there is no info about running backups in the json output from `pgbackrest info`, so we provided a workaround for that in `src/backups.py`, which sets a Patroni tag to inform that to the other cluster members (which is checked in `src/patroni.py`).

### Manual Testing (with same local built charm)
```shell
git clone git@github.com:canonical/postgresql-operator.git
cd postgresql-operator
git checkout dpe-1768-minor-version-upgrade

charmcraft pack
# Deploy local charm
juju deploy ./*.charm -n 3

# After units settle, run pre-upgrade-check to build update-stack.
juju run-action postgresql/leader pre-upgrade-check --wait

# Refresh command - we are updating to the same charm, which will pass version checking.
juju refresh --path ./*.charm postgresql
```

While reviewing, the following files can be ignored (they are either imported or generated):
* `lib/charms/data_platform_libs/v0/upgrade.py`
* `requirements.txt`
* `tests/integration/ha_tests/application-charm/requirements.txt`
* `tests/integration/new_relations/application-charm/requirements.txt`

We disabled the hashes temporarily on all the `requirements.txt` files due to [canonical/charmcraft#1179] (so we can pack the charms now). I will be reverted later.


src/dependency.json
```diff
@@ -0,0 +1,8 @@
+{
+  "charm": {
```
delgod, line 2: same as here https://github.com/canonical/postgresql-k8s-operator/pull/199#discussion_r1275760205

src/upgrade.py
```diff
@@ -0,0 +1,145 @@
...
+        # Wait until the database initialise.
+        self.charm.unit.status = WaitingStatus("waiting for database initialisation")
+        try:
+            for attempt in Retrying(stop=stop_after_attempt(6), wait=wait_fixed(10)):
+                with attempt:
+                    # Check if the member hasn't started or hasn't joined the cluster yet.
+                    if (
+                        not self.charm._patroni.member_started
+                        or self.charm.unit.name.replace("/", "-")
+                        not in self.charm._patroni.cluster_members
```
delgod, line 103: same as here - https://github.com/canonical/postgresql-k8s-operator/pull/199#discussion_r1274848982

delgod (changes requested):

src/dependency.json
marceloneppel: My bad. I uploaded this update to another [branch](https://github.com/canonical/postgresql-operator/pull/191/commits/6ad056268d7c845e5063189f3d77f55c2aa0d48e) last week. I updated it in this branch/PR on https://github.com/canonical/postgresql-operator/pull/187/commits/735b65412a97b432cd6eaa77768069023ad8cd4f.

src/upgrade.py
marceloneppel: Like above, I updated the wrong branch. This was updated on https://github.com/canonical/postgresql-operator/pull/187/commits/735b65412a97b432cd6eaa77768069023ad8cd4f in this branch/PR.

dragomirp (approved):

PR #193, opened 2023-07-28
[DPE-1763] Juju3 peer secrets

dragomirp:
## Issue
Charm doesn't use secrets for internal sensitive values

## Solution
Use Juju secrets instead of peer data if available


dragomirp:
Mostly a copy of the K8S implementation

dragomirp:
Added [DPE-2325](https://warthogs.atlassian.net/browse/DPE-2325) to move the secrets during upgrade charm once we have both implementation.

delgod (approved):

marceloneppel (approved):

Issue #182, opened 2023-07-12
SyntaxError when trying to create an admin user with data-integrator relation 


gustavosr98:
## Steps to reproduce
```bash
juju deploy postgresql
juju deploy data-integrator postgresql-admin --to 1
juju config postgresql-admin database-name=maas extra-user-roles=admin
juju relate  postgresql postgresql-admin
```

## Expected behavior
Relation to work

## Actual behavior
```bash 
➜  ~ juju status
Model    Controller  Cloud/Region         Version  SLA          Timestamp
default  lxd         localhost/localhost  2.9.42   unsupported  15:58:08-05:00

App               Version  Status   Scale  Charm            Channel    Rev  Exposed  Message
postgresql        14.7     blocked      1  postgresql       14/stable  288  no       Failed to initialize database relation
postgresql-admin           blocked      1  data-integrator  stable      11  no       Please relate the data-integrator with the desired product

Unit                 Workload  Agent  Machine  Public address  Ports  Message
postgresql-admin/0*  blocked   idle   1        10.46.90.163           Please relate the data-integrator with the desired product
postgresql/1*        blocked   idle   1        10.46.90.163           Failed to initialize database relation

Machine  State    Address       Inst id        Series  AZ  Message
1        started  10.46.90.163  juju-f9c688-1  jammy       Running
```

## Versions
Operating system: Jammy
Juju CLI: 2.9.43
Juju agent: 2.9.42
LXD: 5.1

Charm versions:
- postgresql 14/stable rev 288
- data-integrator stable rev 11

## Log output
```bash
# [..]
controller-0: 15:39:42 INFO juju.worker.provisioner entering provisioner task loop; using provisioner pool with 16 workers
controller-0: 15:39:42 INFO juju.worker.provisioner provisioning in zones: [black]
controller-0: 15:40:34 INFO juju.worker.provisioner provisioning in zones: [black]
controller-0: 15:40:34 INFO juju.worker.provisioner found machine pending provisioning id:1, details:1
controller-0: 15:40:36 INFO juju.worker.provisioner trying machine 1 StartInstance in availability zone black
controller-0: 15:40:50 INFO juju.worker.provisioner started machine 1 as instance juju-f9c688-1 with hardware "arch=amd64 cores=0 mem=0M", network config [], volumes [], volume attachments map[], subnets to zones [], lxd profiles []
[...]

PR #199, opened 2023-08-04
[DPE-2235] Add admin extra user role

marceloneppel:
## Issue
There is no `admin` extra user role that can be requested by client application charms.

## Solution
Add only read permissions to the special `admin` role. The permission to create anything in the public schema of the `postgres` database was removed. Copied from https://github.com/canonical/postgresql-k8s-operator/pull/201.

Also add to it the permission to create and drop databases (except the system database `postgres`).

Note: some libs were updated and an adjustment was made to `test_relation_data_is_updated_correctly_when_scaling` to avoid frequent errors.

Fixes https://github.com/canonical/postgresql-operator/issues/182.


dragomirp (approved):

taurus-forever (approved):
Excellent!

PR #200, opened 2023-08-04
[MISC]  More resilient topology observer

dragomirp:
## Issue
A bug was reported in the public MM that the legacy relation for VM doesn't fire the hook to update the primary. It seems to me that the topology observer has crashed at some point and changes in the topology are no longer passed on to relations.

## Solution
Check that the observer is still alive in  _on_update_status and restart it if it os not. This is more of a POC fix to see if it will resolve the bug. If it is indeed the issue at hand, we can revisit with a better solution.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1154,6 +1154,9 @@ class PostgresqlOperatorCharm(CharmBase):
 
         self._set_primary_status_message()
 
+        # Restart topology observer if it is gone
+        self._observer.start_observer()
+
     def _handle_processes_failures(self) -> bool:
         """Handle Patroni and PostgreSQL OS processes failures.
 
--- a/src/cluster_topology_observer.py
+++ b/src/cluster_topology_observer.py
@@ -54,12 +54,16 @@ class ClusterTopologyObserver(Object):
 
     def start_observer(self):
         """Start the cluster topology observer running in a new process."""
-        if (
-            not isinstance(self._charm.unit.status, ActiveStatus)
-            or self._charm._peers is None
-            or "observer-pid" in self._charm._peers.data[self._charm.unit]
-        ):
+        if not isinstance(self._charm.unit.status, ActiveStatus) or self._charm._peers is None:
             return
+        if "observer-pid" in self._charm._peers.data[self._charm.unit]:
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -1154,6 +1154,9 @@ def _on_update_status(self, _) -> None:
 
         self._set_primary_status_message()
 
+        # Restart topology observer if it is gone
+        self._observer.start_observer()
+
```
dragomirp, line 1159: We should most probably consider something like systemd timer unit in the future.

src/cluster_topology_observer.py
```diff
@@ -54,12 +54,16 @@ def __init__(self, charm: CharmBase):
...
+        if not isinstance(self._charm.unit.status, ActiveStatus) or self._charm._peers is None:
             return
+        if "observer-pid" in self._charm._peers.data[self._charm.unit]:
+            # Double check that the PID exists
+            pid = int(self._charm._peers.data[self._charm.unit]["observer-pid"])
+            try:
+                os.kill(pid, 0)
+                return
+            except OSError:
+                pass
```
dragomirp, line 66: Check that the pid is actually there.

dragomirp:
MM bugreport: https://chat.charmhub.io/charmhub/pl/bxb6wpoyn7bii8wikmoc8i73eo

tests/unit/test_charm.py
```diff
@@ -652,6 +656,7 @@ def test_on_update_status_after_restore_operation(
         _update_relation_endpoints,
         _handle_workload_failures,
         _set_primary_status_message,
+        _,
```
taurus-forever, line 659: _start_observer ?

taurus-forever (approved):
Do we have a log file for observer?

tests/unit/test_charm.py
dragomirp: It's the mock from above.

dragomirp:
> Do we have a log file for observer?

Should be writing to `/var/log/cluster_topology_observer.log` but I didn't think to ask for a copy.

marceloneppel (approved):

PR #211, opened 2023-08-25
[DPE-2038] Basic alerts

dragomirp:
## Issue
No alert rules set when integrating with COS

## Solution
Add some basic prometheus rules and an empty dir for loki rules


paulomach (approved):

taurus-forever (approved):
Another hint for AlerRules: Backup/Restore failed (we have recently created a Loki IMHO prefix for this cases).

PR #214, opened 2023-08-28
[DPE-2467] Socket connection

dragomirp:
## Issue
Default configuration prevents socket connections for the operator user from within the charm machine

## Solution
* Allow socket connections
* Restrict the access to Patroni configuration

### DIFF
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -99,6 +99,7 @@ postgresql:
   pgpass: /tmp/pgpass
   pg_hba:
     - local all backup peer map=operator
+    - local all operator password
     - local all monitoring password
     {%- if not connectivity %}
     - {{ 'hostssl' if enable_tls else 'host' }} all all 0.0.0.0/0 reject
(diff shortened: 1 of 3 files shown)


tests/unit/test_cluster.py
```diff
@@ -273,7 +273,7 @@ def test_render_patroni_yml_file(self, _, _render_file, _get_postgresql_version)
         _render_file.assert_called_once_with(
             "/var/snap/charmed-postgresql/current/etc/patroni/patroni.yaml",
             expected_content,
-            0o644,
+            0o600,
```
dragomirp, line 276: Should we keep the group permissions here?

templates/patroni.yml.j2
```diff
@@ -99,6 +99,7 @@ postgresql:
   pgpass: /tmp/pgpass
   pg_hba:
     - local all backup peer map=operator
+    - local all operator password
```
dragomirp, line 102: This is plaintext password auth. It should be fine for socket connections and in line with how monitoring connects. Alternatively we can demand md5 or scram.

tests/unit/test_cluster.py
marceloneppel: From what I remember, we don't need that.

marceloneppel (approved):

taurus-forever (approved):
Tnx! BTW, are you going to allow socket connection for K8s as well?

PR #220, opened 2023-09-01
[DPE-2495] Ensure that snap_daemon home exists

dragomirp:
## Issue
We expect that a home is created and set for snap_daemon to run specific commands (pgbackrest, patronictl), but creating it during cluster initialisation may not happen fast enough and may cause race conditions.

## Solution
Move the snap_daemon setup to the _on_install hook for now.


taurus-forever (approved):
Thank you! LGTM.

PR #223, opened 2023-09-05
[DPE-1935][DPE-1754] Postgresql logs

dragomirp:
## Issue
Postgresql's logs are not split into a separate location

## Solution
Split the Postgresql logs and add the log rotation configuration from the spec.


src/grafana_dashboards/postgresql-metrics.json
```diff
@@ -3033,7 +3033,7 @@
   "style": "dark",
   "tags": [
     "postgres",
-    "db",
```
dragomirp, line 3036: Typo

templates/patroni.yml.j2
dragomirp, line 1: Copy of the configuration in the spec.

dragomirp, line 1: Mostly copied from the spec.

taurus-forever (approved):
Tnx!

PR #225, opened 2023-09-11
[MISC] Secondary units stuck on configuring exporter

dragomirp:
## Issue
Starting the Postgresql exporter is frequently failing on replica units, likely due to LXD update.

## Solution
Trigger the exporter start in on peer data change, after Patroni and PG are started.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -137,10 +137,10 @@ jobs:
             echo "mark_expression=" >> $GITHUB_OUTPUT
           else
             echo Skipping unstable tests
-            echo "mark_expression=not unstable" >> $GITHUB_OUTPUT
+            echo "mark_expression=and not unstable" >> $GITHUB_OUTPUT
           fi
       - name: Run integration tests
-        run: tox run -e ${{ matrix.tox-environments }}-${{ env.libjuju }} -- -m 'not not${{ env.libjuju }} and ${{ steps.select-tests.outputs.mark_expression }}' --keep-models
+        run: tox run -e ${{ matrix.tox-environments }}-${{ env.libjuju }} -- -m 'not not${{ env.libjuju }} ${{ steps.select-tests.outputs.mark_expression }}' --keep-models
         env:
           AWS_ACCESS_KEY: "${{ secrets.AWS_ACCESS_KEY }}"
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -137,10 +137,10 @@ jobs:
             echo "mark_expression=" >> $GITHUB_OUTPUT
           else
             echo Skipping unstable tests
-            echo "mark_expression=not unstable" >> $GITHUB_OUTPUT
+            echo "mark_expression=and not unstable" >> $GITHUB_OUTPUT
```
dragomirp, line 140: This seems to have broken the scheduled tests.

src/backups.py
```diff
@@ -411,6 +411,9 @@ def _initialise_stanza(self) -> None:
 
     @property
     def _is_primary_pgbackrest_service_running(self) -> bool:
+        if not self.charm.primary_endpoint:
+            logger.warning("Failed to contact pgBackRest TLS server: no promary endpoint")
+            return False
```
dragomirp, line 416: [Sometimes](https://github.com/canonical/postgresql-operator/actions/runs/6157787902/job/16709350681) there's an unhandled exception due to subprocess None parameter.

src/grafana_dashboards/postgresql-metrics.json
dragomirp, line 1: To bring in line with DPE-2588

src/charm.py
```diff
@@ -947,14 +950,6 @@ def _on_start(self, event: StartEvent) -> None:
 
         self.unit.set_workload_version(self._patroni.get_postgresql_version())
 
-        try:
-            # Set up the postgresql_exporter options.
-            self._setup_exporter()
-        except snap.SnapError:
-            logger.error("failed to set up postgresql_exporter options")
-            self.unit.status = BlockedStatus("failed to set up postgresql_exporter options")
```
dragomirp, line 955: There will no longer be a blocked status over this, but the call in `_on_peer_relation_changed()` will retry if not successful.

marceloneppel (approved):
Thanks!

src/backups.py
taurus-forever: `s/promary/primary/`?

taurus-forever (approved):
LGTM, but one typo in text message.

PR #229, opened 2023-09-20
DPE-2617 Basic profile support

paulomach:
## Issue

Profiles support missing from PG

Same as PG K8s [PR#245](https://github.com/canonical/postgresql-k8s-operator/pull/254)

### DIFF
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -64,6 +64,11 @@ bootstrap:
         {%- endif %}
         archive_mode: on
         wal_level: logical
+        {%- if pg_parameters %}
+        {%- for key, value in pg_parameters.items() %}
+        {{key}}: {{value}}
+        {%- endfor -%}
+        {% endif %}
 
   {%- if restoring_backup %}
   method: pgbackrest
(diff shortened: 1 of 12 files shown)


templates/patroni.yml.j2
```diff
@@ -64,6 +64,11 @@ bootstrap:
         {%- endif %}
         archive_mode: on
         wal_level: logical
+        {%- if pg_parameters %}
+        {%- for key, value in pg_parameters.items() %}
+        {{key}}: {{value}}
+        {%- endfor -%}
+        {% endif %}
```
dragomirp, line 71: I think if we pass an empty dict as a default `pg_parameters` we can remove the outer if, but that can be changed later.

dragomirp (approved):

templates/patroni.yml.j2
paulomach: I'm think I've tested this, but will recheck

dragomirp: > I'm think I've tested this, but will recheck

Not worth rerunning CI over it.

PR #232, opened 2023-09-26
[MISC] Copy fixes of DPE-2626 and DPE-2627 from k8s

dragomirp:
Copy fixes for stuck blocking and multiple attempts to create the s3 bucket.


taurus-forever (approved):
Tnx!

marceloneppel (approved):
Thanks Dragomir!

PR #235, opened 2023-09-29
[DPE-2290] [DPE-2681] Upgrade from 14/stable and add integration tests

marceloneppel:
## Issue
We cannot upgrade from `14/stable`.

There are no integration tests for the upgrade process.

## Solution
Add logic to populate the upgrade stack field if we're upgrading from a upgrade unsupported charm revision.

The monitoring user and its credentials are being created when upgrading from `14/stable` (which doesn't have that user).

This also fixes the upgrade from a single unit cluster, updates the workload version, refreshes the snap when upgrading from a revision without upgrade support and creates the admin extra user role when it was not created yet in the previous revision.

Add tests that check if the upgrade succeeds, by using some assertions related to the upgrade lib and ensure that no writes from an application are lost during the upgrade process.


src/upgrade.py
```diff
@@ -77,6 +80,49 @@ def log_rollback_instructions(self) -> None:
...
+                self.charm.unit.is_leader()
+                and dependencies is not None
+                and "snap" not in json.loads(dependencies)
+            ):
+                fixed_dependencies = json.loads(dependencies)
+                fixed_dependencies["snap"] = {
+                    "dependencies": {},
+                    "name": "charmed-postgresql",
+                    "upgrade_supported": "^14",
+                    "version": "14.9",
```
taurus-forever, line 103: one more place to update on snap bump? :-(

taurus-forever (approved):
Exceptional! Thank you!

src/upgrade.py
dragomirp: I don't think this should move, as far as I understand, it is for charm revs that don't have it defined.

dragomirp (approved):

src/upgrade.py
marceloneppel: Exactly as Dragomir said. It's here only for the revisions missing the `snap` field in the dependency model.

PR #239, opened 2023-10-05
[DPE-1781] Configuration support

marceloneppel:
## Issue
There is no way to provide custom configurations to the workload (PostgreSQL database).

## Solution
Add to `patroni.yaml` the requested configurations with their default values.

Expose the allowed configurations through charm config options (and add validators based on the allowed values).

`max_connections` and `max_prepared_transactions` can be changed only through the Patroni API, so they are changed in the charm code differently from the other configurations.

Also, added a call to `time.sleep(10)` to give some time to Patroni to reload the configurations before it's possible to know which ones need a database restart. Previously, sometimes a change in `shared_buffers` wouldn't trigger a restart.


taurus-forever (approved):
LGTM.

JFMI, why don't we set defaults in config.yaml?

marceloneppel:
> LGTM.
> 
> JFMI, why don't we set defaults in config.yaml?

No specific reason. I think it's a good idea to put them. I'll update the code.

marceloneppel:
> LGTM.
> 
> JFMI, why don't we set defaults in config.yaml?

I added the default values and the feedback from the K8S charm PR to https://github.com/canonical/postgresql-operator/pull/239/commits/efc7c8fe5a98f571ca8aeeaa5ad97b522b760184.

delgod (approved):

dragomirp (approved):

PR #244, opened 2023-10-09
DPE-2655 profile limit option

paulomach:
port of k8s [pr#283](https://github.com/canonical/postgresql-k8s-operator/pull/283)


src/charm.py
```diff
@@ -1361,10 +1367,13 @@ def _is_workload_running(self) -> bool:
     def update_config(self, is_creating_backup: bool = False) -> bool:
         """Updates Patroni config file based on the existence of the TLS files."""
         enable_tls = all(self.tls.get_tls_files())
+        limit_memory = None
+        if self.config.profile_limit_memory:
+            limit_memory = self.config.profile_limit_memory * 10**6
```
marceloneppel, line 1372: Should we use `1024 * 1024` instead of `10**6` as the config description states `MegaBytes`?

paulomach: I think MB to B is `10**6` or am I wrong?

marceloneppel (approved):

src/charm.py
dragomirp: Does the docs specify if it's MB or MiB?

dragomirp (approved):

src/charm.py
paulomach: Spec is MB, and pg config knows how to handle different suffixes. Charm calculations being done in bytes

PR #261, opened 2023-10-23
[DPE-1993] Patroni COS

dragomirp:
Add Patroni COS dashboard and configuration.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -157,7 +157,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         return [
             {
                 "metrics_path": "/metrics",
-                "static_configs": [{"targets": ["localhost:8008"]}],
+                "static_configs": [{"targets": [f"{self._unit_ip}:8008"]}],
                 "tls_config": {"insecure_skip_verify": True},
                 "scheme": "https" if self.is_tls_enabled else "http",
             }
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,7 +5,9 @@
 """Charmed Machine Operator for the PostgreSQL database."""
 import json
 import logging
+import os
 import subprocess
+import time
 from typing import Dict, List, Optional, Set
 
 from charms.data_platform_libs.v0.data_models import TypedCharmBase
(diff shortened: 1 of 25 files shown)


src/charm.py
```diff
@@ -142,12 +142,27 @@ def __init__(self, *args):
...
         self._grafana_agent = COSAgentProvider(
             self,
-            metrics_endpoints=[
-                {"path": "/metrics", "port": METRICS_PORT},
+            metrics_endpoints=[{"path": "/metrics", "port": METRICS_PORT}],
+            scrape_configs=self.patroni_scrape_config,
+            refresh_events=[
+                self.on[PEER].relation_changed,
+                self.on.secret_changed,
+                self.on.secret_remove,
```
dragomirp, line 150: `certificates` rel events don't guarantee that the scheme will switch, so we update on events that can potentially change or remove the certs

templates/patroni.yml.j2
```diff
@@ -19,7 +19,7 @@ log:
   file_size: 600
 
 restapi:
-  listen: '{{ self_ip }}:8008'
+  listen: '0.0.0.0:8008'
```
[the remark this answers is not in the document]
taurus-forever: I think it will produce issues with [juju spaces](https://juju.is/docs/juju/juju-spaces) once we are ready.
Anyway LGTM for now. 

@delgod any security concerns here?

src/charm.py
```diff
@@ -142,12 +142,27 @@ def __init__(self, *args):
...
             ],
             log_slots=[f"{POSTGRESQL_SNAP_NAME}:logs"],
         )
 
+    def patroni_scrape_config(self) -> List[Dict]:
+        """Generates scrape config for the Patroni metrics endpoint."""
+        return [
+            {
+                "metrics_path": "/metrics",
+                "static_configs": [{"targets": ["localhost:8008"]}],
```
taurus-forever, line 160: why not {{ self_ip }} here instead of localhost? :-)

taurus-forever (approved):

src/charm.py
dragomirp: It should work, but last time we discussed that, I think COS preferred to only scrape localhost. I'll ping them on MM to recheck.

dragomirp: Sounds like it's OK to do that. Retested locally and all seems to be in order.

marceloneppel (approved):
LGTM!

PR #262, opened 2023-10-25
add plpgsql extension

TakoB222:


marceloneppel (approved):
Hi @TakoB222! Thanks! Do you or @taurus-forever have a ticket on Jira or other place where we can track the request of changes like this one?

delgod (approved):

7annaba3l:
@delgod and @marceloneppel : For the record, we do not need a parameter to control the activation of the extension (let's keep it activated by default), for now. We might reconsider if someone is asking us to do so.

PR #263, opened 2023-10-25
[DPE-2789] update data-platform-workflows

dragomirp:
Update to latest workflows and refactor CI to more closely resemble MySQL's.


.github/workflows/ci.yaml
```diff
@@ -22,45 +22,40 @@ on:
...
+    name: Build charm
+    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v5.1.2
+    permissions:
+      actions: write  # Needed to manage GitHub Actions cache
 
-  integration-test:
+  gh-hosted-integration-test:
     strategy:
       fail-fast: false
-      max-parallel: 6
```
dragomirp, line 58: Do we still want this?

tests/integration/conftest.py
```diff
@@ -1,82 +1,9 @@
...
 from pytest_operator.plugin import OpsTest
 
-from tests.integration.helpers import construct_endpoint
-
-AWS = "AWS"
-GCP = "GCP"
-
-
-@pytest.fixture(scope="module")
-async def cloud_configs(ops_test: OpsTest) -> None:
```
dragomirp, line 20: Moved to backups

renovate.json
dragomirp, line 1: Most custom rules are dropped for now, will try to reenable what we need at a later date.

tox.ini
dragomirp, line 1: Poetry and shellcheck need to be installed separately.

.github/workflows/ci.yaml
```diff
@@ -69,22 +64,21 @@ jobs:
...
-            free-disk-space: true
-    name: ${{ matrix.tox-environments }} | ${{ matrix.agent-versions }}
+          - juju-snap-channel: "3.1/stable"
+            agent-version: "3.1.6"
+            libjuju-version:  "3.2.2"
+            exclude-mark:  "juju2"
+          - juju-snap-channel: "2.9/stable"
+            agent-version: "2.9.45"
+            libjuju-version:  "2.9.45.0"
+            exclude-mark:  "juju3"
```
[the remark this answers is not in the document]
dragomirp: Some charms used in tests don't work on Juju 3. Without a mark in include, they either need to be split off in a separate suite to be excluded or have logic that skips them in the tests themselves.

pyproject.toml
```diff
@@ -13,57 +13,68 @@ repository = "https://github.com/canonical/postgresql-operator"
...
-pytest = "7.4.2"
-pytest-operator = "0.29.0"
-juju = "2.9.44.1 || 3.2.0.1"  # renovate libjuju
-psycopg2 = {version = "^2.9.5", extras = ["binary"]}
-
-[build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+pytest = "^7.4.0"
+pytest-github-secrets = {git = "https://github.com/canonical/data-platform-workflows", tag = "v5.0.0", subdirectory = "python/pytest_plugins/github_secrets"}
```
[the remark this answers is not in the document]
dragomirp: Bumped.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
dragomirp: I'd imagine for the same reasons as `libjuju-version` and `agent-version`. Matrix keys should be flat and this is  specific to the value of `juju-snap-channel`.

.github/workflows/ci.yaml
marceloneppel: I think it's not needed based on Alex comments some time ago and also because we should move to self hosted runners later (which should have more runners available as I heard recently).

marceloneppel (approved):
LGTM! Thanks Dragomir!

marceloneppel (approved):

PR #285, opened 2023-11-15
[DPE-2904] Switch to self hosted runners

dragomirp:
Switch to Canonical hosted runners.


pyproject.toml
```diff
@@ -57,24 +57,24 @@ coverage = {extras = ["toml"], version = "^7.3.2"}
 pytest = "^7.4.0"
 pytest-asyncio = "^0.21.1"
 jsonschema = "^4.19.1"
-psycopg2 = {version = "^2.9.9", extras = ["binary"]}
+psycopg2-binary = "^2.9.9"
```
dragomirp, line 60: No libpq on the runner.

tests/integration/ha_tests/conftest.py
```diff
@@ -7,7 +7,8 @@
 from pytest_operator.plugin import OpsTest
 from tenacity import Retrying, stop_after_delay, wait_fixed
 
-from tests.integration.ha_tests.helpers import (
+from ..helpers import run_command_on_unit
+from .helpers import (
```
dragomirp, line 11: Relative imports seems to be the easiest way to get the test collector to play game.

tox.ini
dragomirp, line 1: Do we need the individual testenvs? They are no longer used by CI.

tests/integration/conftest.py
```diff
@@ -10,3 +10,17 @@ async def charm(ops_test: OpsTest):
...
+    """Skip test if it is not on juju2."""
+    if hasattr(ops_test.model, "list_secrets"):
+        pytest.skip("Test can't run on Juju3")
+
+
+@pytest.fixture(scope="module")
+def juju3(ops_test):
+    """Skip test if it is not on juju3."""
+    if not hasattr(ops_test.model, "list_secrets"):
+        pytest.skip("Test can't run on Juju2")
```
dragomirp, line 26: Skip fixures for juju version to replace the custom marks.

.github/workflows/ci.yaml
dragomirp, line 1: It doesn't seem to be a simple way to use a matrix with a reusable workflow, so I've added a two separate tasks to run the workflow for juju2 and juju3.

tests/integration/markers.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+
+import pytest
+
+from .juju_ import juju_major_version
+
+juju2 = pytest.mark.skipif(juju_major_version != 2, reason="Requires juju 2")
+juju3 = pytest.mark.skipif(juju_major_version != 3, reason="Requires juju 3")
```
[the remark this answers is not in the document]
dragomirp: > imo, I would keep this and `juju_.py` identical to mysql-operator

The Mysql one imports ops, which IMHO shouldn't be installed by the integration tests at all.

> I don't think it makes sense to skip juju >=4 tests when what we care about is yes/no for secrets support

This is used to skip tests that run on a specific juju version, not necessarily due to lack of secrets, but also due to client charms having issues with the newer version. E.g. at the moment livepatch only works on juju 2. When we start testing juju 4, client charms may not follow the same timetable as us, in the mean time we should still be able to verify that the tests work on older versions.

marceloneppel (commented):
@carlcsaposs-canonical, is https://github.com/canonical/data-platform-workflows/pull/94 related to https://github.com/canonical/mysql-operator/pull/354/files?

marceloneppel:
> > @carlcsaposs-canonical, is [canonical/data-platform-workflows#94](https://github.com/canonical/data-platform-workflows/pull/94) related to https://github.com/canonical/mysql-operator/pull/354/files?
> 
> No, mysql-operator#354 uses github-hosted runners (with the reusable integration workflow). dpw#94 adds a self-hosted runner reusable integration workflow that has the same API as the github-hosted integration workflow
> 
> does that answer what you were asking?

Cool. Thanks for the explanation, Carl!

marceloneppel (approved):

marceloneppel (approved):

taurus-forever (approved):
1) LGTM.
2) Please update CONTRIBUTION.md guilde to represent new way to run the test(s):
```
tox run -e integration -- 'tests/integration/ha_tests/test_restore_cluster.py' --group='1' -m 'not unstable' --model test
```

taurus-forever (dismissed):
LGTM, but `tox run -e integration # integration tests` is still wrong in CONTRIBUTION.md, isn't it?

dragomirp:
> LGTM, but `tox run -e integration # integration tests` is still wrong in CONTRIBUTION.md, isn't it?

I tried and `tox -e integration` still collects all the tests. We should most probably add a HOWTO run individual suites, but that's out of scope IMHO.

taurus-forever (approved):

taurus-forever (approved):
Let's the party begin!

PR #291, opened 2023-11-23
[DPE-3039] Update Snap

dragomirp:
Updates from PPAs:
* Patroni 3.1.2
* Pgbackrest 2.48

### DIFF
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -43,6 +43,8 @@ logger = logging.getLogger(__name__)
 
 PG_BASE_CONF_PATH = f"{POSTGRESQL_CONF_PATH}/postgresql.conf"
 
+RUNNING_STATES = ["running", "streaming"]
+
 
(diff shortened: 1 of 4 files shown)


src/cluster.py
```diff
@@ -43,6 +43,8 @@
 
 PG_BASE_CONF_PATH = f"{POSTGRESQL_CONF_PATH}/postgresql.conf"
 
+RUNNING_STATES = ["running", "streaming"]
```
dragomirp, line 46: [Patroni 3.0.4](https://patroni.readthedocs.io/en/latest/releases.html#version-3-0-4) exposes the status of replicas using `pg_stat_get_wal_receiver()` so the replicas are usually in streaming state.

marceloneppel (approved):

taurus-forever (approved):

PR #298, opened 2023-11-29
[DPE-2740] Handle tables ownership

marceloneppel:
## Issue
Some charms like Discourse-K8S need to have the ownership of the tables and other objects in a database to correctly run its migrations if a database from a previous Discourse installation is used, otherwise it will not complete the migrations and will throw an error. Also, the existing Discourse K8S test was marked as `unstable` in the past because that charm was migrated from the legacy to the modern relation interface.

## Solution
This is a port of https://github.com/canonical/postgresql-k8s-operator/pull/334.
`lib/charms/postgresql_k8s/v0/postgresql.py` is copied from the K8S charm.

If the relation between PostgreSQL and Discourse K8S (or any other charm), continue with the same behaviour, but after another instance of the charm is related to the PostgreSQL again, requesting the same database, check if there is no other application related and that previously requested the same database. If so, change the ownership of the objects in the database to the new user created for this relation. If there is already another application related to this database, just give permission to the database object, as we are already doing.

I manually tested these changes in a cross-model relation between this charm and Discourse-K8S.

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -22,7 +22,9 @@ import logging
 from typing import Dict, List, Optional, Set, Tuple
 
 import psycopg2
+from ops.model import Relation
 from psycopg2 import sql
+from psycopg2.sql import Composed
 
 # The unique Charmhub library identifier, never change it
 LIBID = "24ee217a54e840a598ff21a079c3e678"
@@ -32,7 +34,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 20
+LIBPATCH = 21
 
 INVALID_EXTRA_USER_ROLE_BLOCKING_MESSAGE = "invalid role(s) for extra user roles"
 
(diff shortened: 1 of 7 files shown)


dragomirp (approved):

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -308,6 +308,55 @@ def enable_disable_extensions(self, extensions: Dict[str, bool], database: str =
...
+UNION SELECT 2 AS index,'ALTER SEQUENCE '|| sequence_schema || '."' || sequence_name ||'" OWNER TO {};' AS statement
+FROM information_schema.sequences WHERE NOT sequence_schema IN ('pg_catalog', 'information_schema')
+UNION SELECT 3 AS index,'ALTER FUNCTION '|| nsp.nspname || '."' || p.proname ||'"('||pg_get_function_identity_arguments(p.oid)||') OWNER TO {};' AS statement
+FROM pg_proc p JOIN pg_namespace nsp ON p.pronamespace = nsp.oid WHERE NOT nsp.nspname IN ('pg_catalog', 'information_schema')
+UNION SELECT 4 AS index,'ALTER VIEW '|| schemaname || '."' || viewname ||'" OWNER TO {};' AS statement
+FROM pg_catalog.pg_views WHERE NOT schemaname IN ('pg_catalog', 'information_schema')) AS statements ORDER BY index) LOOP
+      EXECUTE format(r.statement);
+  END LOOP;
+END; $$;"""
+                ).format(
```
taurus-forever, line 333: This is a bit read-only... and looks crazy... LGTM, but we should think on background about possible simplification here.

taurus-forever (approved):

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: Thanks Alex. I agree. I created https://warthogs.atlassian.net/browse/DPE-3090 for that.

PR #301, opened 2023-11-29
[DPE-2904] Reenable backup tests and revert to reusable workflow

dragomirp:
Backup tests were missed when switching to self-hosted runners and are failing likely due to a proxy issue.

This PR re-enables all missed tests and switches back to the reusable workflow.


marceloneppel (approved):
LGTM for now if we need more investigation or help from IS.

taurus-forever (approved):

PR #302, opened 2023-11-30
[DPE-3077] Block roles in legacy interface

dragomirp:
Block the charm if the legacy interface requests roles

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -39,7 +39,7 @@ ROLES_BLOCKING_MESSAGE = (
 class DbProvides(Object):
     """Defines functionality for the 'provides' side of the 'db' relation.
 
-    or roles    Hook events observed:
+    Hook events observed:
         - relation-changed
         - relation-departed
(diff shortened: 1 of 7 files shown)


src/relations/db.py
```diff
@@ -31,11 +31,15 @@
...
+ROLES_BLOCKING_MESSAGE = (
+    "roles requested through relation, use postgresql_client interface instead"
+)
+
 
 class DbProvides(Object):
     """Defines functionality for the 'provides' side of the 'db' relation.
 
-    Hook events observed:
+    or roles    Hook events observed:
```
marceloneppel, line 42: Is this `or roles` text supposed to be here? It's seems weird.

marceloneppel (approved):

marceloneppel (approved):

taurus-forever (approved):

taurus-forever:
@dragomirp do we need the same commit for K8s charm?

Issue #259, opened 2023-10-23
Certificates integration raises ValueError: This extension has already been set.


Martwall:
## Steps to reproduce

```shell
juju deploy postgresql
juju deploy self-signed-certificates --channel edge
juju integrate self-signed-certificates postgresql
```
## Expected behavior
The integration hooks finishes successfully. 

## Actual behavior

The relation-changed hook errors. The CertificateCreationRequestEvent raises a ValueError exception.

## Versions

Operating system: Ubuntu 22.04

Juju CLI: 3.1.6-genericlinux-amd64

Juju agent: 14.9

Charm revision: 336

LXD: 5.18

## Log output

Juju debug log: 

```shell
unit-self-signed-certificates-0: 08:32:13 DEBUG unit.self-signed-certificates/0.juju-log certificates:9: ops 2.7.0 up and running.
unit-self-signed-certificates-0: 08:32:13 DEBUG unit.self-signed-certificates/0.juju-log certificates:9: Emitting Juju event certificates_relation_created.
unit-postgresql-0: 08:32:13 DEBUG unit.postgresql/0.juju-log certificates:9: ops 2.7.0 up and running.
unit-postgresql-0: 08:32:13 DEBUG unit.postgresql/0.juju-log certificates:9: Emitting Juju event certificates_relation_created.
unit-self-signed-certificates-0: 08:32:13 INFO juju.worker.uniter.operation ran "certificates-relation-created" hook (via hook dispatching script: dispatch)
unit-self-signed-certificates-0: 08:32:14 DEBUG unit.self-signed-certificates/0.juju-log certificates:9: ops 2.7.0 up and running.
unit-self-signed-certificates-0: 08:32:14 DEBUG unit.self-signed-certificates/0.juju-log certificates:9: Emitting Juju event certificates_relation_joined.
unit-postgresql-0: 08:32:14 INFO juju.worker.uniter.operation ran "certificates-relation-created" hook (via hook dispatching script: dispatch)
unit-self-signed-certificates-0: 08:32:14 INFO juju.worker.uniter.operation ran "certificates-relation-joined" hook (via hook dispatching script: dispatch)
unit-postgresql-0: 08:32:14 DEBUG unit.postgresql/0.juju-log certificates:9: ops 2.7.0 up and running.
unit-self-signed-certificates-0: 08:32:14 DEBUG unit.self-signed-certificates/0.juju-log certificates:9: ops 2.7.0 up and running.
[...]

PR #321, opened 2024-01-12
[DPE-2830] Update TLS lib and PostgreSQL TLS lib

marceloneppel:
## Issue
As shown in https://github.com/canonical/postgresql-operator/issues/259, it's not possible to relate the PostgreSQL charms with the Self Signed Certificates operator.

## Solution
Update the TLS library to V2 and remove the addition of the BasicContraints critical extension from the certificate request (done int the PostgreSQL TLS library from the K8S charm).


taurus-forever (approved):
Tnx! Now we can switch to self-signed-operator! Tnx!

dragomirp (approved):

PR #329, opened 2024-01-24
[DPE-2904] Switching to self-hosted runners

dragomirp:
Switching to latest dpw and self hosted runners.

Backup test seem to work fine now.


taurus-forever (changes requested):
Nice. Let's see how self-hosted stable for our needs.
LGTM, but apply changes requested by Carl (+1 there).

marceloneppel (approved):

taurus-forever (approved):

PR #336, opened 2024-01-29
[DPE-3422] Switch to self signed certificates

marceloneppel:
## Issue
Only a single TLS test uses the Self Signed Certificates operator on Juju 3.

## Solution
Use the same approach implemented on [PgBouncer K8S](https://github.com/canonical/pgbouncer-k8s-operator/pull/191) (conditionally use TLS Certificates Operator or Self Signed Certificates operator, depending on the Juju version).


taurus-forever (approved):
Nice!

PR #340, opened 2024-01-31
[DPE-3380] Handle S3 relation in primary non-leader unit

marceloneppel:
## Issue
The stanza can only be initialised in the PostgreSQL primary unit if it's the Juju leader unit.

Otherwise, the unit fails with an error, as shown in https://github.com/canonical/postgresql-k8s-operator/issues/370.

## Solution
Ported from https://github.com/canonical/postgresql-k8s-operator/pull/375.

Allow the non-leader unit (when it's the PostgreSQL primary) to initialise the stanza.

For that, all checks for the unit being the Juju leader unit were changed to check whether the unit is the PostgreSQL primary.

Also, some coordination is now made in the relation databags to share the stanza name and the `init-pgbackrest` flag.

The current flow doesn't change if the PostgreSQL primary unit is also the Juju leader unit.

If they are different units, then:

- The stanza name and the `init-pgbackres` flag are set in the primary unit relation databag.
- The leader unit copies those values to the application relation databag (to be used later for other actions in backups).
- The `pgbackrest check` command is executed in the primary unit, which clears the `init-pgbackrest` flag from the unit relation databag.
- The leader unit clears the `init-pgbackrest` flag from the application relation databag.
- The primary unit clears the stanza name from the unit relation databag.

Both unit and integration tests were added/updated to ensure we keep this flow intact.

Also, the fix from https://github.com/canonical/postgresql-k8s-operator/pull/273 was added through this PR, as it was not applied to the VM charm.


dragomirp (approved):

taurus-forever (approved):
Great!

PR #346, opened 2024-02-06
[DPE-3257] Fix network cut tests

marceloneppel:
## Issue
The two network cut tests are being skipped from CI due to instabilities on them.

## Solution
Enable the tests again on CI.

Fix the causes of the issues in the tests:

- Fix the removal of the old IP of the unit that was separated from the network of the other units. Sometimes, the old IP was already removed in the leader-elected hook before the leader was able to reconfigure the cluster.
- In the tests, the IP address of the unit that was separated from the network of the other units was retrieved through `juju exec` because the IP reported by `juju status` might be outdated sometimes.
- Added a check for readiness in the Patroni API and added a retry to the attempt to connect to the database after the network connection was restored ([DPE-3489](https://warthogs.atlassian.net/browse/DPE-3489) was created to investigate why it was failing).

[DPE-3489]: https://warthogs.atlassian.net/browse/DPE-3489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taurus-forever (approved):
Excellent! Thank you!

dragomirp (approved):

PR #347, opened 2024-02-06
deployment test "zero-units"

BalabaDmintri:
Test Zero-units: continuous writes ON, deploy 3 units, check, scale to 0 units, check no errors, scale to 3, check

### DIFF
--- a/tests/integration/ha_tests/test_self_healing.py
+++ b/tests/integration/ha_tests/test_self_healing.py
@@ -5,6 +5,7 @@ import asyncio
 import logging
 
 import pytest
+from pip._vendor import requests
 from pytest_operator.plugin import OpsTest
 from tenacity import Retrying, stop_after_delay, wait_fixed
 
@@ -14,7 +15,7 @@ from ..helpers import (
     get_machine_from_unit,
     get_password,
     get_unit_address,
-    run_command_on_unit,
+    run_command_on_unit, scale_application,
 )
 from .conftest import APPLICATION_NAME
 from .helpers import (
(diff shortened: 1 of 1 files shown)


tests/integration/ha_tests/test_self_healing.py
```diff
@@ -540,3 +541,78 @@ async def test_network_cut_without_ip_change(
...
+
+    # Scale the database to one unit.
+    logger.info("scaling database to one unit")
+    await add_unit_with_storage(ops_test, storage=primary_storage, app=APP_NAME)
+    logger.info("checking whether writes are increasing")
+    await are_writes_increasing(ops_test)
+
+    # Scale the database to three units.
+    for store_id in storage_id_list:
+        await add_unit_with_storage(ops_test, storage=store_id, app=APP_NAME)
```
taurus-forever, line 617: @marceloneppel 

JFMI, should re use ops lib directly as in helper it refers to [this](https://github.com/BalabaDmintri/postgresql-operator/blob/9d7bfed48288486872c6f26b7e6892a5e9f2e605/tests/integration/ha_tests/helpers.py#L742) which is resolved:
```
    Note: this function exists as a temporary solution until this issue is resolved:
    https://github.com/juju/python-libjuju/issues/695
```

taurus-forever (changes requested):
@BalabaDmintri  Thank you for the contribution. We need to add a ticket number and PR body describing What?Where?Why? better.

Re:PR. It is overall good start!!! We need to cover all scaling cases:

re-scaling-back with and without drive. E.g. the default hostpath storage will remove drive, so re-scaling will start on empty disk. In the same time btrfs will keep disk available and re-scaling will be with a disk (could be default disk or manually provided). Could be wrongly provided...  

I mean we have those cases:
 * user wants to restore the same storage: `remove-unit postgresql/3 && && add-unit -n 2` should reuse old storage as user didn't specify it (it can be new disk, e.g. hostpath). => should be OK, long SST/clone process for 2nd+ node, but 0->1 is new DB init as no disk.
 * user wants to provide specific **correct** storage => should be OK automatically,  just fast resilvering.
 * user made mistake and specified **wrong** storage (another cluster) => charm blocked. do NOT corrupt foreign disks!!!
 * user made mistake and specified **wrong** storage (empty/corrupter/no psql data on it) => block to be on a safe side. force user to remove disk and re-scale without any disks.
 * ...

I believe we should test all those cases above. 

P.S. Charm should be safe and block bootstrap if found foreign cluster on disk, etc (probable improvement here required).

tests/integration/ha_tests/test_self_healing.py
marceloneppel: Yes. We should remove the workaround and use the methods provided by the lib.

marceloneppel (changes requested):

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -540,3 +541,78 @@ async def test_network_cut_without_ip_change(
...
+            resp = requests.get(f"http://{unit_ip}:8008")
+            assert resp.status_code != 200, f"status code = {resp.status_code}, message = {resp.text}"
+        except requests.exceptions.ConnectionError as e:
+            assert True, f"unit host = http://{unit_ip}:8008, all units shutdown"
+        except Exception as e:
+            assert False, f"{e} unit host = http://{unit_ip}:8008, something went wrong"
+
+    # Scale the database to one unit.
+    logger.info("scaling database to one unit")
+    await add_unit_with_storage(ops_test, storage=primary_storage, app=APP_NAME)
```
delgod, line 611: After the unit starts, we should check if the data on the storage has been actually restored.

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -540,3 +541,78 @@ async def test_network_cut_without_ip_change(
...
+    logger.info("checking whether writes are increasing")
+    await are_writes_increasing(ops_test)
+
+    # Scale the database to three units.
+    for store_id in storage_id_list:
+        await add_unit_with_storage(ops_test, storage=store_id, app=APP_NAME)
+    await check_writes(ops_test)
```
delgod, line 618: After 2nd and 3rd units start, it is needed to check that data on them is restored from WAL (not via backup/restore).
Maybe @dragomirp or @marceloneppel know how to check this

delgod (changes requested):

tests/integration/ha_tests/test_self_healing.py
dragomirp: [ha_test.helpers.reused_replica_storage()](https://github.com/canonical/postgresql-operator/blob/main/tests/integration/ha_tests/helpers.py#L838) and [ha_test.helpers.reused_full_cluster_recovery_storage()](https://github.com/canonical/postgresql-operator/blob/main/tests/integration/ha_tests/helpers.py#L852) should do the trick.

tests/integration/ha_tests/test_self_healing.py
dragomirp: IIRC this is only available in libjuju 3

taurus-forever:
Discussed this today on the sync. For the history:
* our Moto: Never destroy customer data!
* The goal for this test is to block the charm IF user made a mistake and re-scale application with wrong storage. IF user want to re-deploy application using foreign storage: it is possible (with warning in logs), but re-scaling up with foreign storage must be impossible.

Some example in my mind:
* deploy 3 units, generate data, create backup, generate data.
* scale to 2 unit, change data, scale-up to 3 units (with dirty storage of 3rd unit) => cluster member should join well 
* scale to 2 unit, restore backup, scale-up to 3 units (with dirty storage of 3rd unit) => cluster member should be blocked as patroni will see the data but will not be able to join the cluster. Fix: scale down, remove 3rd storage, scale-up (automated re-silvering used). All OK.
* scale to 2 units, change 3rd storage to foreign cluster, scale-up => unit must be blocked as foreign data detected. Better safe then sorry.
* ...

BalabaDmitri:
@marceloneppel Can you run actions

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -543,3 +558,119 @@ async def test_network_cut_without_ip_change(
...
+    await are_writes_increasing(ops_test)
+
+    # Connect to the database.
+    # Create test data.
+    logger.info("connect to DB and create test table")
+    await create_test_data(connection_string)
+
+    # Test to check the use of different versions postgresql.
+    # Release of a new version of charm with another version of postgresql,
+    # it is necessary to implement a test that will check the use of different versions of postgresql.
```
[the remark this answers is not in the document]
BalabaDmitri: @marceloneppel I tried this option but all revisions are contains Postgresql  version 14.11. 

Remove the logic with different versions of Postgres from this PR and create another PR with this logic ?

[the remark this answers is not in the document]
BalabaDmitri: @marceloneppel added issue #457

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -543,3 +558,119 @@ async def test_network_cut_without_ip_change(
...
+    logger.info("checking whether writes are increasing")
+    await are_writes_increasing(ops_test)
+
+    logger.info("check test database data")
+    await validate_test_data(connection_string)
+
+    logger.info("database scaling up to two units using third-party cluster storage")
+    new_unit = await add_unit_with_storage(
+        ops_test, app=app, storage=second_storage, is_blocked=True
+    )
```
taurus-forever, line 636: nit: IMHO, it worth to check: we are blocked with the right message (foreign disk).

taurus-forever (approved):
LGTM, this is something we were missing. Tnx!

BalabaDmitri:
@dragomirp Can you check

dragomirp:
Hi, @BalabaDmitri, can you resync with main again, to retrigger the CI and to get some fixes. Sorry for asking again.

PR #351, opened 2024-02-12
[DPE-3559] Stabilise restore cluster test

marceloneppel:
## Issue
The restore cluster test was failing a lot of times due to the `replication` user password not being updated, which led to follower units not being able to connect to the primary to get the data.

## Solution
Update all users' passwords in the test.

Also, use this opportunity to change the blocked status with the message `no primary in the cluster` to a waiting status because it's resolved after some time (no need for user action).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -89,7 +89,7 @@ from utils import new_password
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


taurus-forever (approved):
Nice catch! I have just a curiosity question: why it failed sometimes only IF password mismatch was invariant. LGTM

PR #355, opened 2024-02-13
[DPE-3567] Enable the Landscape test

marceloneppel:


marceloneppel:
Closed because this uses an older version of the Landscape charm, which isn't the one that people are deploying.

PR #357, opened 2024-02-15
Cover missing test cases for internal users

TakoB222:
### Issue
- Authentification should be enabled by default always. No anonymous/guest 
access ever;
- get-password and set-password functions should be created for internal 
users;

### DIFF
--- a/tests/integration/test_password_rotation.py
+++ b/tests/integration/test_password_rotation.py
@@ -4,6 +4,7 @@
 import json
 import time
 
+import psycopg2
 import pytest
 from pytest_operator.plugin import OpsTest
 
@@ -12,8 +13,11 @@ from .helpers import (
     CHARM_SERIES,
     METADATA,
     check_patroni,
+    db_connect,
     get_leader_unit,
     get_password,
+    get_primary,
+    get_unit_address,
     restart_patroni,
     set_password,
 )
@@ -106,7 +110,6 @@ async def test_password_rotation(ops_test: OpsTest):
     assert new_rewind_password == await get_password(ops_test, any_unit_name, "rewind")
     assert rewind_password != new_rewind_password
 
-
     # Restart Patroni on any non-leader unit and check that
     # Patroni and PostgreSQL continue to work.
     restart_time = time.time()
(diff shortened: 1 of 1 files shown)


marceloneppel (approved):

tests/integration/test_password_rotation.py
```diff
@@ -39,13 +43,17 @@ async def test_deploy_active(ops_test: OpsTest):
         await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1500)
 
 
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
```
taurus-forever, line 46: @dragomirp Q: is it necessary here? AFAIK, it is set for group 1 in line 28 above.

tests/integration/test_password_rotation.py
```diff
@@ -122,6 +159,16 @@ async def test_empty_password(ops_test: OpsTest) -> None:
     assert password == "None"
 
 
+@pytest.mark.group(1)
+async def test_db_connection_with_empty_password(ops_test: OpsTest):
+    """Test that user can't connect with empty password."""
+    primary = await get_primary(ops_test, f"{APP_NAME}/0")
+    address = get_unit_address(ops_test, primary)
+    with pytest.raises(psycopg2.Error):
+        with db_connect(address, "") as connection:
```
taurus-forever, line 168: This will test user `operator` only. Should we test lack of connection without password using user `postgresql` and the rest from the list https://charmhub.io/postgresql/docs/e-users
@marceloneppel what do you think, is it necessary?

taurus-forever (commented):
1) Thank you! Nice improvement for our tests!
2) I would improve it a bit, e.g. compare the passwords to make sure they are different.
+ some minor questions to Data team.

tests/integration/test_password_rotation.py
marceloneppel: I think it's not necessary to test with the other users.

tests/integration/test_password_rotation.py
dragomirp: No, one self hosted mark per suite should be sufficient.

dragomirp:
Hi, @TakoB222, please sync up with main and remove the excess self-hosted test mark.

dragomirp:
/canonical/self-hosted-runners/run-workflows dc417ecfacb35ae68d38fe1faa021c3380fa8a5d

dragomirp (approved):

taurus-forever (approved):
Thank you Artem.

I would also test here:
 * compare initial passwords for all users. Ensure all are different. => auto-generation works properly.
 * compare the value after changing password. At the moment we:
 1) check `"password" in result.keys()` => some password returned
 2) ensure `superuser_password != new_superuser_password`  => `ensure something has been changed`. We do NOT check: `the new password charm returns matches the password we requested to set`. Isn't it?

taurus-forever:
Merging as all reviewers ACKed the PR. Please create the separate PR to address my comments above (if they are valid).
Thank you for the contribution!

PR #361, opened 2024-02-15
[DPE-3591] Fix shared buffers validation

marceloneppel:
## Issue
The validation for the `shared_buffers` parameter was not correct. The calculus of the maximum allowed shared buffers value was not correctly converting the available memory for that to the same unit as the `shared_buffers` parameter that the user informs.

Also, the `effective_cache_size` parameter did not consider the user-provided `shared_buffers` value.

One last issue: sometimes, the restart of the PostgreSQL database was not triggered because Patroni hadn't finished reloading the configuration (even after 10 seconds).

## Solution
Fix the calculus of the maximum allowed shared buffers value to be in the same unit as the user-provided value.

Considered the user-provided shared buffers value for the effective cache size parameter calculus.

An improved retry mechanism was added to check whether a restart is needed after a configuration change.

Unit tests for the library file are being updated to the [K8S charm](https://github.com/canonical/postgresql-k8s-operator/pull/396), which owns the library.

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -579,9 +579,13 @@ END; $$;"""
                 f"Shared buffers config option should be at most 40% of the available memory, which is {shared_buffers_max_value_in_mb}MB"
             )
         if profile == "production":
-            # Use 25% of the available memory for shared_buffers.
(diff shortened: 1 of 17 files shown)


dragomirp (approved):

taurus-forever (approved):
Nice!

PR #366, opened 2024-02-22
[MISC] Check for peer relation when getting the primary endpoint

dragomirp:
Try to speed up and stabilise the tests. Based of the shared buffers PR.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1458,7 +1458,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
 
         # Try to connect
         if not self._can_connect_to_postgresql:
-            logger.debug("Early exit update_config: Cannot connect to Postgresql")
+            logger.warning("Early exit update_config: Cannot connect to Postgresql")
             return False
         self._validate_config_options()
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -1395,10 +1395,8 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
     @property
     def _can_connect_to_postgresql(self) -> bool:
         try:
-            for attempt in Retrying(stop=stop_after_delay(30), wait=wait_fixed(3)):
-                with attempt:
-                    assert self.postgresql.get_postgresql_timezones()
-        except RetryError:
+            assert self.postgresql.get_postgresql_timezones()
+        except Exception:
             logger.debug("Cannot connect to database")
             return False
         return True
--- a/src/charm.py
+++ b/src/charm.py
@@ -1021,10 +1021,6 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             event.defer()
             return
 
-        if not self.primary_endpoint:
-            logger.debug("Primary endpoint not generated")
-            self.postgresql.primary_host = self.postgresql.current_host
-
         # Create the default postgres database user that is needed for some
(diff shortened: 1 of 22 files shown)


src/charm.py
```diff
@@ -299,6 +300,9 @@ def postgresql(self) -> PostgreSQL:
     @property
     def primary_endpoint(self) -> Optional[str]:
         """Returns the endpoint of the primary instance or None when no primary available."""
+        if not self._peers:
+            logger.debug("primary endpoint early exit: Peer relation not joined yet.")
+            return None
```
dragomirp, line 305: If the peer relation is not yet joined, don't try to find the primary.

src/charm.py
```diff
@@ -1036,6 +1053,10 @@ def _start_primary(self, event: StartEvent) -> None:
             logger.exception(e)
             self.unit.status = BlockedStatus("Failed to create postgres user")
             return
+        except PostgreSQLListUsersError:
+            logger.warning("Deferriing on_start: Unable to list users")
+            event.defer()
+            return
```
dragomirp, line 1059: Skipping the wait in primary_endpoint could lead o exceptions here.

src/charm.py
```diff
@@ -1424,6 +1456,10 @@ def update_config(self, is_creating_backup: bool = False) -> bool:
             logger.debug("Early exit update_config: Patroni not started yet")
             return False
 
+        # Try to connect
+        if not self._can_connect_to_postgresql:
+            logger.debug("Early exit update_config: Cannot connect to Postgresql")
+            return False
```
dragomirp, line 1462: If the node's db is not yet up, we can't really check those values. Next update status should run this again.

src/charm.py
```diff
@@ -308,7 +312,20 @@ def primary_endpoint(self) -> Optional[str]:
...
+                        # TODO figure out why peer data is not available
+                        if (
+                            primary_endpoint
+                            and len(self._units_ips) == 1
+                            and len(self._peers.units) > 1
+                        ):
+                            logger.warning(
+                                "Possibly incoplete peer data: Will not map primary IP to unit IP"
+                            )
+                            return primary_endpoint
```
dragomirp, line 324: On juju 3 we don't seem to be getting the peer data used to get the IP addr when a new unit is joining (e.g. when scaling in the `test_db` suite).

src/charm.py
```diff
@@ -308,7 +312,20 @@ def primary_endpoint(self) -> Optional[str]:
...
+                        logger.debug(
+                            "primary endpoint early exit: Primary IP not in cached peer list."
+                        )
+                        primary_endpoint = None
```
dragomirp, line 328: `_units_ips` is a property and will not update on retry. The Patroni response might change while retrying, but IMHO it's better to leave that to the topology observer script.

tests/integration/test_db_admin.py
```diff
@@ -34,7 +34,6 @@
 
 
 @pytest.mark.group(1)
-@pytest.mark.skip(reason="DB Admin tests are currently broken")
```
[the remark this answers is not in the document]
dragomirp: I reenabled it here to reduce the amount CI runs.

src/charm.py
dragomirp: Changed to warning.

taurus-forever (approved):
Discussed with Dragomir, merging this to have progress with tests stabilisation otherwise shared_buffers test cannot be merged.

PR #374, opened 2024-02-28
Add `test_config_parameters` test

TakoB222:
### Issue
- Charge each charm config option using clusterized app and make sure changes are 
applied without error (and without downtime if possible):
  - check values boundaries (e.g. try 42 where 1-10 are allowed)
  - check types (e.g. try 42 where boolean true/false are expected)
  - check invalid config values (see the “test string” below)


tests/integration/test_config.py
```diff
@@ -0,0 +1,103 @@
...
+from .helpers import (
+    CHARM_SERIES,
+    DATABASE_APP_NAME,
+    get_leader_unit,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
```
dragomirp, line 18:
```suggestion
```
The other test suites are on GH runners.

tests/integration/test_config.py
```diff
@@ -0,0 +1,102 @@
...
+async def test_config_parameters(ops_test: OpsTest) -> None:
+    """Build and deploy one unit of PostgreSQL and then test config with wrong parameters."""
+    # Build and deploy the PostgreSQL charm.
+    async with ops_test.fast_forward():
+        charm = await ops_test.build_charm(".")
+        await ops_test.model.deploy(
+            charm,
+            num_units=1,
+            series=CHARM_SERIES,
+            constraints="arch=arm64",
```
dragomirp, line 29:
```suggestion
```
No arm runners yet.

src/charm.py
```diff
@@ -862,8 +862,16 @@ def _on_config_changed(self, _) -> None:
...
             logger.debug("Early exit on_config_changed: upgrade in progress")
             return
-        # update config on every run
-        self.update_config()
+
+        try:
+            # update config on every run
+            self.update_config()
+        except ValueError as e:
+            self.unit.status = BlockedStatus(f"Configuration Error: {str(e)}")
```
dragomirp, line 870:
```suggestion
            self.unit.status = BlockedStatus(f"Configuration Error. Please check the logs")
            logger.error("Invalid configuration: %s", str(e))
```
Looks like the validation message can be quite long and it's not going to be very readable in the status.

Better to dump it in the logs and inform the user to check that in the status message.

tests/integration/test_config.py
```diff
@@ -0,0 +1,101 @@
...
+        },  # config option is between 0 and 2147483647
+        {
+            "vacuum_autovacuum_freeze_max_age": {"99999", "100000"}
+        },  # config option is between 100000 and 2000000000
+        {
+            "vacuum_autovacuum_vacuum_cost_delay": {"-2", "-1"}
+        },  # config option is between -1 and 100
+        {
+            "vacuum_vacuum_freeze_table_age": {"-1", "0"}
+        },  # config option is between 0 and 2000000000
```
dragomirp, line 83: Sets are not ordered, so when you create a list from them there is no guarantee that the order will be kept. Best to just use lists from the get go.

dragomirp:
Multiple validator decorators will fail to kick in, if both fields are present and one is valid and the other is invalid. Cases like https://github.com/TakoB222/postgresql-operator/blob/configurations/src/config.py#L240-L242 should be changed to something like:
```
    @validator("response_lc_monetary", "response_lc_numeric", "response_lc_time")
```

tests/integration/test_config.py
```diff
@@ -0,0 +1,101 @@
...
+            assert "Configuration Error" in leader_unit.workload_status_message
+
+    config = {}
+    for c in configs:
+        for k, v in c.items():
+            config.update({k: list(v)[1]})
+            break
+
+    await ops_test.model.applications[DATABASE_APP_NAME].set_config(config)
+    await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=1500)
```
dragomirp, line 101: Once `v` is list:
```suggestion
    charm_config = {}
    for config in configs:
        for k, v in config.items():
            logger.info(k)
            charm_config[k] = v[0]
            await ops_test.model.applications[DATABASE_APP_NAME].set_config(charm_config)
            await ops_test.model.wait_for_idle(
                apps=[DATABASE_APP_NAME], status="blocked", timeout=100
            )
            assert "Configuration Error" in leader_unit.workload_status_message
            charm_config[k] = v[1]

    await ops_test.model.applications[DATABASE_APP_NAME].set_config(charm_config)
    await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=100)
```
We should test each invalid value one by one, since otherwise the other values would keep the charm blocked, even if the validation didn't work for one specific value.

We should log the value that is currently tested, so it would be easier to debug in the future.

We don't have to wait for idle long, on_config should be fairly fast.

dragomirp:
Hi, @TakoB222, The Juju 2 test fails on: https://github.com/canonical/postgresql-operator/blob/960c04ebc39789d791ea03a1aa111ba34f8e610f/src/charm.py#L479
We should add the same except clause for ValueError there as in `_on_config_changed`

dragomirp (approved):

taurus-forever (approved):
Thank you @TakoB222 !!!

PR #382, opened 2024-03-02
[MISC] Integration tests tweaks

dragomirp:
* Update charm libs
* Update ruff config
* Drive by tweaks of the integration tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,10 +104,12 @@ target-version = ["py38"]
 [tool.ruff]
 # preview and explicit preview are enabled for CPY001
 preview = true
-explicit-preview-rules = true
 target-version = "py38"
 src = ["src", "."]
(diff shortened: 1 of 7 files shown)


pyproject.toml
dragomirp, line 1: Update ruff configs so it doesn't complain.

tests/integration/new_relations/test_new_relations.py
```diff
@@ -344,6 +344,7 @@ async def test_an_application_can_request_multiple_databases(ops_test: OpsTest):
 
 
 @pytest.mark.group(1)
+@pytest.mark.abort_on_fail
```
dragomirp, line 347: If raft gets stuck the other tests will likely fail as well.

taurus-forever (approved):
Tnx!

PR #395, opened 2024-03-11
[DPE-3593] Only check config values against the DB in on_config_changed

dragomirp:
`_validate_config_options` connecting to the database will be done only in on_config_changed


src/charm.py
```diff
@@ -1507,21 +1507,22 @@ def update_config(self, is_creating_backup: bool = False) -> bool:
     def _validate_config_options(self) -> None:
         """Validates specific config options that need access to the database or to the TLS status."""
         if (
-            self.config.instance_default_text_search_config is not None
+            self.config.instance_default_text_search_config != "pg_catalog.simple"
             and self.config.instance_default_text_search_config
             not in self.postgresql.get_postgresql_text_search_configs()
```
dragomirp, line 1512: We should consider storing the current value in the peer data and only check if it is changed.

marceloneppel (approved):

taurus-forever (approved):
I believe idea of hardcoding option values in charm code is a bad idea, but it is a Marcelo call as PostgreSQL Lead here. I believe it is to create a quick fix for the customer and find a time to invent the proper fix.

BTW, I believe we should check them on config_update event only and defer if PostgreSQL is not running.

src/charm.py
```diff
@@ -862,19 +863,25 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
         else:
             self.unit.status = WaitingStatus(PRIMARY_NOT_REACHABLE_MESSAGE)
 
-    def _on_config_changed(self, _) -> None:
+    def _on_config_changed(self, event) -> None:
         """Handle configuration changes, like enabling plugins."""
         if not self.is_cluster_initialised:
-            logger.debug("Early exit on_config_changed: cluster not initialised yet")
+            logger.debug("Defer on_config_changed: cluster not initialised yet")
+            event.defer()
```
dragomirp, line 870: Deferring, because otherwise we'll never validate the config.

src/charm.py
```diff
@@ -862,19 +863,25 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
...
             return
-
         try:
+            self._validate_config_options()
             # update config on every run
             self.update_config()
+        except psycopg2.OperationalError:
+            logger.debug("Defer on_config_changed: Cannot connect to database")
+            event.defer()
+            return
```
dragomirp, line 884: If can't connect, defer

src/charm.py
```diff
@@ -1504,24 +1510,18 @@ def update_config(self, is_creating_backup: bool = False) -> bool:
     def _validate_config_options(self) -> None:
         """Validates specific config options that need access to the database or to the TLS status."""
         if (
-            self.config.instance_default_text_search_config is not None
-            and self.config.instance_default_text_search_config
+            self.config.instance_default_text_search_config
             not in self.postgresql.get_postgresql_text_search_configs()
         ):
-            raise Exception(
+            raise ValueError(
```
dragomirp, line 1516: Switching to ValueError to resuse the handler for config blocking

src/charm.py
taurus-forever: Curiosity question to simplify the code:
can we improve the ops lib to print DEBUG message if defer() got comment?
It will be only:
```
event.defer("Defer on_config_changed: cluster not initialized yet")
```

taurus-forever (approved):
I like it. The final works to Marcelo. Tnx!

PR #399, opened 2024-03-14
[TEST] Remove pyopenssl dependency

dragomirp:


dragomirp:
Can't remove. Older version is already installed in the system packages

PR #405, opened 2024-03-18
[MISC] Remove skipped legacy tests

dragomirp:


tests/integration/test_db.py
```diff
@@ -365,34 +323,6 @@ async def test_roles_blocking(ops_test: OpsTest, charm: str) -> None:
     )
 
 
-@pytest.mark.group(1)
-@pytest.mark.unstable
-async def test_weebl_db(ops_test: OpsTest, charm: str) -> None:
```
dragomirp, line 370: Weebl charm doesn't deploy with the legacy charm either.

tests/integration/test_db.py
```diff
@@ -185,48 +192,6 @@ async def test_relation_data_is_updated_correctly_when_scaling(ops_test: OpsTest
                 psycopg2.connect(primary_connection_string)
 
 
-@pytest.mark.group(1)
-@pytest.mark.skip(reason="Should be ported and moved to the new relation tests")
-async def test_nextcloud_db_blocked(ops_test: OpsTest, charm: str) -> None:
```
dragomirp, line 190: Moved to new relation test

tests/integration/new_relations/test_new_relations.py
```diff
@@ -362,9 +362,16 @@ async def test_relation_data_is_updated_correctly_when_scaling(ops_test: OpsTest
...
+            unit for unit in units_to_remove if unit != leader_unit.name
+        ])
         await ops_test.model.wait_for_idle(
-            apps=[DATABASE_APP_NAME], status="active", timeout=1500, wait_for_exact_units=2
+            apps=[DATABASE_APP_NAME], status="active", timeout=600, wait_for_exact_units=3
+        )
+        await ops_test.model.applications[DATABASE_APP_NAME].destroy_units(leader_unit.name)
+        await ops_test.model.wait_for_idle(
+            apps=[DATABASE_APP_NAME], status="active", timeout=600, wait_for_exact_units=2
         )
```
dragomirp, line 375: Destroy the original follower unit and then the leader

marceloneppel: That would be a good addition to ensure that the connection details in the relation databag are updated correctly, even if the primary is not the Juju leader.

Another thing to remember: add a note to https://warthogs.atlassian.net/browse/DPE-3684 to add the test back that deletes the leader and the synchronous standby (so we can ensure that if that happens - two units at the same time - the cluster doesn't get stuck).

taurus-forever: I would say those extensive tests are absolutely valid and diverse the ticket in backlog to be implemented. At the moment we are trying to stabilize CI test to move forward.

tests/integration/test_db.py
taurus-forever: Requested migration officially: https://github.com/canonical/postgresql-operator/pull/405

taurus-forever (approved):

marceloneppel (approved):

tests/integration/new_relations/test_new_relations.py
dragomirp: Added a comment to the ticket.

PR #407, opened 2024-03-18
[MISC] Order set of replicas

dragomirp:
Should prevent pgb from restarting on each run of update status

### DIFF
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -296,7 +296,8 @@ class DbProvides(Object):
             return
 
         # List the replicas endpoints.
-        replicas_endpoint = self.charm.members_ips - {self.charm.primary_endpoint}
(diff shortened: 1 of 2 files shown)


src/relations/db.py
```diff
@@ -296,7 +296,8 @@ def update_endpoints(self, relation: Relation = None) -> None:
             return
 
         # List the replicas endpoints.
-        replicas_endpoint = self.charm.members_ips - {self.charm.primary_endpoint}
+        replicas_endpoint = list(self.charm.members_ips - {self.charm.primary_endpoint})
+        replicas_endpoint.sort()
```
dragomirp, line 300: Sets don't guarantee ordering.

marceloneppel (approved):
LGTM! Thanks!

taurus-forever (approved):

PR #420, opened 2024-03-29
Add  ```test_smoke``` test

BON4:
**Issue https://github.com/canonical/postgresql-operator/issues/441**

- Test Application. VM resources is removed after application removal. Install app, remove app (removal test), check no garbage left.
- Test Unit. Re-Installation. Remove unit with `--force --no-wait`, storage should be detached.
- Test Unit. Garbage ignorance. Charm should deploy in dirty environment with garbage storage.
- Test Application. Application should deploy in dirty environment with garbage storage from another application.


marceloneppel:
Hi @BON4! Thanks for the contribution.

Please add a PR description so we can understand what you're proposing.

Thanks.

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,146 @@
...
+    host = get_unit_address(ops_test, primary_name)
+    result = requests.get(f"http://{host}:8008/health")
+    assert result.status_code == 200
+
+    storage_id_str = storage_id(ops_test, primary_name)
+
+    # Check if storage exists after application deployed
+    assert await is_storage_exists(ops_test, storage_id_str)
+
+    await ops_test.model.remove_application(APPLICATION_NAME, block_until_done=True, force=True, no_wait=True)
```
taurus-forever, line 99: AFAIK, this will remove storage (as it is removed by default on LXD with hostpath storage type (default)), so the next text will be deployed again without any storage. Consider to add destroy_storage=False (but I am not sure if it help).

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,146 @@
...
+
+    primary_name = await get_primary(ops_test, ops_test.model.applications[APPLICATION_NAME].units[0].name)
+    host = get_unit_address(ops_test, primary_name)
+    result = requests.get(f"http://{host}:8008/health")
+    assert result.status_code == 200
+
+    # Create test database to check there is no resouces conflicts
+    await create_db(ops_test, APPLICATION_NAME, TEST_DATABASE_RELATION_NAME)
+
+    # Deploy duplicaate charm
```
taurus-forever, line 128: As discussed on the sync today: we are testing Juju here, but we need to test charm.
Juju guaranties that independent applications will have interdependent storages, no need to test it here, IMHO. But you can keep it as already created as it should be fast.

Our goal is to test charm:
 * charm A can start with storage from A (cleanly). Even if there are some garbage lock files on storage.
 * charm A can start with storage from B (with warning in log, about foreign storage, but deploying anyway as it can be an app recovery process).

P.S. the goal of smoke test is a fast result, so try to keep the test fast (in time).

taurus-forever (commented):
As discussed today, please add link to the GH/Jira issue. We always must link commits to bugtracker.

Please add PR description: What? Why? How?

Please improve the test to check charm and not juju :-)

Tnx!

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,173 @@
...
+            attach_storage=[tag.storage(garbadge_storage)]
+        )
+
+        try:
+            await ops_test.model.wait_for_idle(apps=[DUP_APPLICATION_NAME], timeout=500, status="blocked")
+        except (TimeoutError) as e:
+            logger.info(f"Application is not in blocked state. Checking logs...")
+
+        # Since application have postgresql db in storage from external application it should not be able to connect due to new password
+        assert not await check_password_auth(ops_test, ops_test.model.applications[DUP_APPLICATION_NAME].units[0].name)
```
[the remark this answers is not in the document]
BON4: This charm stalls forever in Init:CrashLoopBackOff loop with "waiting, installing agent". As I know, this [commit](https://github.com/canonical/postgresql-operator/commit/a18b1d3f951c02b6a41b1ab041b3856421476e74) will fix this issue, and deployment will be in `blocked` state instead of falling in to forever CrashLoopBackOff due to database from another application.

BON4:
@dragomirp @marceloneppel bumping

dragomirp:
Hi, @BON4, please sign the CLA and sync up with the main branch.

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # Check if storage exists after application deployed
+        assert await is_storage_exists(ops_test, storage_id_str)
+
+        # Create test database to check there is no resources conflicts
+        await create_db(ops_test, APPLICATION_NAME, TEST_DATABASE_RELATION_NAME)
+
+        # Destroy charm
+        await ops_test.model.destroy_unit(
+            primary_name, force=True, destroy_storage=False, max_wait=1500
```
dragomirp, line 109: `destroy_storage` param is only available in python lib juju 3. We should either call the CLI to achieve the same result or make the test run only on juju 3.

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,188 @@
...
+TEST_GARBADGE_STORAGE_NAME = "test_pgdata"
+TEST_DATABASE_RELATION_NAME = "test_database"
+DUP_APPLICATION_NAME = "postgres-test-dup"
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_app_removal(ops_test: OpsTest, charm: str):
```
dragomirp, line 37: IMHO we don't need this test. The default storage pool for LXD, rootfs, will be deleted when released and destroying the volume is not something that the charm handles.

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,188 @@
...
+            APPLICATION_NAME, block_until_done=True, destroy_storage=True
+        )
+
+        # Check if storage removed after application removal
+        assert not await is_storage_exists(ops_test, storage_id_str)
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_app_force_removal(ops_test: OpsTest, charm: str):
```
dragomirp, line 75: IMHO there is no need to test this either. With force removal the charm code shouldn't even fire, so this would be handled by Juju entirely.

marceloneppel (changes requested):
Thanks for the contribution, @BON4! I requested some changes. I also agree with Dragomir's comments.

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,188 @@
...
+            application_name=DUP_APPLICATION_NAME,
+            num_units=1,
+            series=CHARM_SERIES,
+            config={"profile": "testing"},
+            attach_storage=[tag.storage(garbadge_storage)],
+        )
+
+        try:
+            await ops_test.model.wait_for_idle(
+                apps=[DUP_APPLICATION_NAME], timeout=500, status="blocked"
```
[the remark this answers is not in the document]
BON4: It is not yet implemented in operator but this [commit](https://github.com/canonical/postgresql-operator/commit/a18b1d3f951c02b6a41b1ab041b3856421476e74) will bring changes, so that application will fall in blocked state with `"Failed to start postgresql. The storage belongs to a third-party cluster"` message. I added check for operator auth  down below just to double check and in case this test will be merged before the commit for blocked status.
```
# Since application have postgresql db in storage from external application it should not be able to connect due to new password
assert not await check_password_auth(
    ops_test, ops_test.model.applications[DUP_APPLICATION_NAME].units[0].name
)
```

BON4:
@dragomirp @marceloneppel @taurus-forever Hello team, bumping for review and re-run checks, after changes applied.

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # Check that test database is not exists for new unit
+        assert not await check_db(ops_test, APPLICATION_NAME, TEST_DATABASE_RELATION_NAME)
+
+        await ops_test.model.destroy_unit(primary_name, destroy_storage=False, max_wait=1500)
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_app_recoures_conflicts(ops_test: OpsTest, charm: str):
```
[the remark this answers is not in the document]
BON4: FIxed

tests/integration/ha_tests/test_smoke.py
BON4: @taurus-forever suggested to leave `test_app_force_removal` just for the sake of testing the correctness of storage detaching (charm will be removed with force correctly). Since this is smoke test it should be fast.

tests/integration/ha_tests/test_smoke.py
```diff
@@ -0,0 +1,173 @@
...
+        primary_name = await get_primary(ops_test, ops_test.model.applications[APPLICATION_NAME].units[0].name)
+        assert await is_postgresql_ready(ops_test, primary_name)
+
+        storage_id_str = storage_id(ops_test, primary_name)
+
+        # Check if storage exists after application deployed
+        assert await is_storage_exists(ops_test, storage_id_str)
+
+        # Check that test database is not exists for duplicate application 
+        assert not await check_db(ops_test, APPLICATION_NAME, TEST_DATABASE_RELATION_NAME)
```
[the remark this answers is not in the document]
BON4: Fixed

PR #425, opened 2024-04-03
[DPE-2674] Convert test_charm.py to pytest style testing instead of unittest

lucasgameiroborges:
## Motivation

Allow tests to run with `jujuVersion.has_secrets` both enabled and disabled. Move away from `unittest.TestCase`.

## Structure

Tests are largely unaltered, with few exceptions:

- `test_delete_password` got removed and `test_delete_existing_password_secrets` now handles both versions with and without secrets.
- `test_update_certificate` got removed and `test_update_certificate_secrets` now handles both versions with and without secrets.
- `test_migration_from_databag` and `test_migration_from_single_secret` got adapted in order to run only if `has_secrets == True` since they test migration from databag to secrets.

Only two global pytest fixtures are used: `harness` and `_has_secrets`. Especific patches are kept inside their respective tests.

## Testing
```
cd path/to/tests/
tox run -e unit
```

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -47,8 +47,8 @@ def harness():
     harness.cleanup()
 
 
-# This causes every test defined in this file will run 2 times, with
-# the patch of charm.JujuVersion.has_secrets set as True and as False
+# This causes every test defined in this file to run 2 times, each with
+# charm.JujuVersion.has_secrets set as True or as False
 @pytest.fixture(params=[True, False], autouse=True)
 def _has_secrets(request, monkeypatch):
     monkeypatch.setattr("charm.JujuVersion.has_secrets", PropertyMock(return_value=request.param))
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -541,659 +541,595 @@ def test_enable_disable_extensions(harness, caplog):
             new_harness.charm.enable_disable_extensions()
             assert postgresql_mock.enable_disable_extensions.call_count == 1
 
-@patch("charm.PostgresqlOperatorCharm.enable_disable_extensions")
-@patch("charm.snap.SnapCache")
-@patch("charm.Patroni.get_postgresql_version")
 @patch_network_get(private_address="1.1.1.1")
-@patch("charm.PostgreSQLProvider.oversee_users")
-@patch("charm.PostgresqlOperatorCharm._update_relation_endpoints", new_callable=PropertyMock)
-@patch("charm.PostgresqlOperatorCharm.postgresql")
-@patch("charm.PostgreSQLProvider.update_endpoints")
-@patch("charm.PostgresqlOperatorCharm.update_config")
-@patch(
-    "charm.Patroni.member_started",
-    new_callable=PropertyMock,
-)
-@patch("charm.Patroni.bootstrap_cluster")
-@patch("charm.PostgresqlOperatorCharm._replication_password")
-@patch("charm.PostgresqlOperatorCharm._get_password")
-@patch("charm.PostgresqlOperatorCharm._reboot_on_detached_storage")
-@patch("upgrade.PostgreSQLUpgrade.idle", return_value=True)
-@patch(
-    "charm.PostgresqlOperatorCharm._is_storage_attached",
-    side_effect=[False, True, True, True, True],
-)
-def test_on_start(
-    harness,
-    _is_storage_attached,
-    _idle,
-    _reboot_on_detached_storage,
-    _get_password,
-    _replication_password,
-    _bootstrap_cluster,
-    _member_started,
-    _,
-    __,
-    _postgresql,
-    _update_relation_endpoints,
-    _oversee_users,
-    _get_postgresql_version,
-    _snap_cache,
-    _enable_disable_extensions,
-):
-    rel_id = harness.add_relation(PEER, harness.charm.app.name)
-    _get_postgresql_version.return_value = "14.0"
+def test_on_start(harness):
+    with (
+        patch("charm.PostgresqlOperatorCharm.enable_disable_extensions") as _enable_disable_extensions,
+        patch("charm.snap.SnapCache") as _snap_cache,
+        patch("charm.Patroni.get_postgresql_version") as _get_postgresql_version, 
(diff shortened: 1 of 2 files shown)


tests/unit/test_charm.py
```diff
@@ -40,47 +38,43 @@
...
-
-    @pytest.fixture
-    def use_caplog(self, caplog):
-        self._caplog = caplog
-
-    @patch_network_get(private_address="1.1.1.1")
-    @patch("charm.subprocess.check_call")
-    @patch("charm.snap.SnapCache")
-    @patch("charm.PostgresqlOperatorCharm._install_snap_packages")
-    @patch("charm.PostgresqlOperatorCharm._reboot_on_detached_storage")
```
dragomirp, line 64: Why move away from patch decorators?

lucasgameiroborges: My main motivation here was to make clear what parameter correspond to what patch (also not needing to have blank params). It was also the approach adopted in Kafka's repo I think. I don't have a strong opinion here, can revert if wanted.

Another option would be to convert the patches into fixtures, but this would considerably increase verbosity in the beginning of the file.

dragomirp: OK, I'm fine with the change.

dragomirp (approved):

taurus-forever (approved):
LGTM, re: decorators => the final decision is up to Marcelo here.

tests/unit/test_charm.py
```diff
@@ -1683,45 +1668,53 @@ def test_push_tls_files_to_workload(self, _get_tls_files, _render_file, _update_
...
+        assert harness.charm.push_tls_files_to_workload()
+        assert _render_file.call_count == 3
 
         # Test when not all TLS files are available.
         for _ in range(3):
             _render_file.reset_mock()
-            self.assertFalse(self.charm.push_tls_files_to_workload())
-            self.assertEqual(_render_file.call_count, 2)
+            assert not (harness.charm.push_tls_files_to_workload())
+            assert _render_file.call_count == 2
```
marceloneppel, line 1678: Why the number of calls changed from 3 to 2 in this case?

tests/unit/test_charm.py
```diff
@@ -1735,479 +1728,470 @@ def test_juju_run_exec_divergence(self, _juju_version: Mock, _topology_observer:
...
 
-            self.harness.charm.remove_secret("unit", "non-existing-secret")
-            assert (
-                "Non-existing field 'non-existing-secret' was attempted to be removed"
-                in self._caplog.text
-            )
+        # as this test checks for a migration from databag to secrets,
+        # there's no need for this test when secrets are not enabled.
+        if not _has_secrets:
+            return
```
marceloneppel, line 1954: This block may be moved to before the above `add_relation` command to exit early. The same may be done on `test_migration_from_single_secret `.

tests/unit/test_charm.py
```diff
@@ -1735,479 +1728,470 @@ def test_juju_run_exec_divergence(self, _juju_version: Mock, _topology_observer:
...
-                in self._caplog.text
-            )
+@pytest.mark.parametrize("scope,is_leader", [("app", True), ("unit", True), ("unit", False)])
+@patch_network_get(private_address="1.1.1.1")
+def test_migration_from_databag(harness, _has_secrets, scope, is_leader):
+    with (
+        patch("charm.PostgresqlOperatorCharm._on_leader_elected"),
+    ):
+        rel_id = harness.add_relation(PEER, harness.charm.app.name)
+        """Check if we're moving on to use secrets when live upgrade from databag to Secrets usage."""
```
marceloneppel, line 1949: The docstring should come before any code. The same applies to `test_migration_from_single_secret`.

tests/unit/test_charm.py
```diff
@@ -40,47 +38,43 @@
...
-    def test_on_install(
-        self,
-        _is_storage_attached,
-        _reboot_on_detached_storage,
-        _install_snap_packages,
-        _snap_cache,
-        _check_call,
-    ):
+    ) as _is_storage_attached:
+        harness.add_relation(PEER, harness.charm.app.name)
```
marceloneppel, line 70: I think it may be possible to move this to the fixture (to avoid forgetting about it when creating new tests and also have it in all tests, as we should have a peer relation present in pretty much every hook execution). For the calls that need the relation ID (`rel_id = harness.add_relation(PEER, harness.charm.app.name)`), you probably may use `harness.model.get_relation(PEER).id`).

tests/unit/test_charm.py
```diff
@@ -1735,479 +1728,470 @@ def test_juju_run_exec_divergence(self, _juju_version: Mock, _topology_observer:
...
-        assert self.harness.charm.get_secret("app", "replication") is None
 
-        self.harness.set_leader(False)
-        self.harness.update_relation_data(
-            self.rel_id, self.charm.unit.name, {"somekey": "somevalue"}
-        )
-        self.harness.charm.remove_secret("unit", "somekey")
-        assert self.harness.charm.get_secret("unit", "somekey") is None
+@patch_network_get(private_address="1.1.1.1")
+def test_delete_existing_password_secrets(harness, _has_secrets, caplog):
```
marceloneppel, line 1896: IMO, we may rename this test `test_delete_password` as it's being executed for both secret-enabled and secrets-not-enabled Juju versions. I'd also do the same for `test_update_certificate_secrets` (which I'd rename `test_update_certificate`).

marceloneppel (commented):
Great work, @lucasgameiroborges! I left some comments in the code. Please take a look and evaluate what makes sense and what doesn't. Thanks!

tests/unit/test_charm.py
lucasgameiroborges: I don't see a change in the number of calls when comparing to the version in the main branch. the assertion before the `for` loop is testing `call_count == 3` and the assertion inside the loop is testing `call_count == 2`.

marceloneppel: My bad. I looked at it wrongly. The change is right. Sorry for the confusion.

marceloneppel (approved):

tests/unit/test_charm.py
marceloneppel: I'm also fine with the change. Let's use the patches as they were implemented in this PR.

PR #317, opened 2024-01-10
Experimental standby cluster

marceloneppel:

PR #452, opened 2024-04-22
[DPE-2953] Cross-region async replication

marceloneppel:
## Issue
It's not possible to replicate data between regions.

## Solution
Implement cross-region async replication. This PR is a rebranded and more stable version of https://github.com/canonical/postgresql-operator/pull/317.

With this PR, it's no longer necessary to remove the relation and relate again when a switchover is needed.

Also, the names of the relations can easily be changed to others, like `cluster-one` and `cluster-two`, for example, to avoid confusing users.

This is mostly a copy and paste from https://github.com/canonical/postgresql-k8s-operator/pull/447.

Important changes:

* `src/relations/async_replication.py` contains the logic to make one cluster the primary and the other the standby. To make the standby cluster follow the primary cluster, the candidate for the standby cluster needs to be restarted.
  
  * The most important part of the logic is handled by the `_on_async_relation_changed`, which takes care of restarting the standby cluster units databases in order to make them replicate data from the primary cluster.
  
  * Additionally to everything implemented on https://github.com/canonical/postgresql-k8s-operator/pull/447, this PR adds the coordination of which RAFT members should be part of the Patroni internal RAFT cluster while starting the units in the standby cluster. It's done through the return value from the `get_partner_addresses` method.

* Passwords update will be implemented in another PR, as this one is already huge.

* If the standby cluster has its relation removed, it goes to a `read-only mode` and can be promoted later to a normal cluster through the `promote-cluster` action.

How to deploy: https://discourse.charmhub.io/t/charmed-postgresql-deploy-async-replication/13991
How to trigger a switchover: https://discourse.charmhub.io/t/charmed-postgresql-async-switchover/13993

Additional instructions:
* https://discourse.charmhub.io/t/charmed-postgresql-clients-for-async-replication/13992
* https://discourse.charmhub.io/t/charmed-postgresql-recovery/13994
* https://discourse.charmhub.io/t/charmed-postgresql-removal-of-async-replication/13997

Integration tests: https://github.com/canonical/postgresql-operator/pull/453


actions.yaml
```diff
@@ -17,6 +17,12 @@ list-backups:
   description: Lists backups in s3 storage.
 pre-upgrade-check:
   description: Run necessary pre-upgrade checks and preparations before executing a charm refresh.
+promote-cluster:
+  description: Promotes the cluster of choice to a primary cluster. Must be ran against the leader unit.
+  params:
+    force-promotion:
```
taurus-forever, line 23: AS a followup commit, please sync with MySQL to use common option (force=Trues there).

actions.yaml
```diff
@@ -17,6 +17,12 @@ list-backups:
...
+      type: boolean
+      description: Force the promotion of a cluster when there is already a primary cluster.
```
taurus-forever, line 25: AS discussed, force promotion should be used for Failover. The safe switchover do not need force.

P.S. the very first join, when two are Primary can be done though Force to be safe.

metadata.yaml
```diff
@@ -26,6 +26,10 @@ peers:
     interface: upgrade
 
 provides:
+  async-primary:
+    interface: async_replication
+    limit: 1
+    optional: true
```
taurus-forever, line 32: as followup commit: let's move to agreed `async-offer`.

metadata.yaml
```diff
@@ -37,6 +41,10 @@ provides:
     limit: 1
 
 requires:
+  async-replica:
+    interface: async_replication
+    limit: 1
+    optional: true
```
taurus-forever, line 47: as followup commit: let's move to agreed `async` here.

src/cluster.py
```diff
@@ -492,8 +556,10 @@ def render_patroni_yml_file(
             version=self.get_postgresql_version().split(".")[0],
             minority_count=self.planned_units // 2,
             pg_parameters=parameters,
+            primary_cluster_endpoint=self.charm.async_replication.get_primary_cluster_endpoint(),
+            extra_replication_endpoints=self.charm.async_replication.get_standby_endpoints(),
         )
-        self.render_file(f"{PATRONI_CONF_PATH}/patroni.yaml", rendered, 0o600)
+        self.render_file(f"{PATRONI_CONF_PATH}/patroni.yaml", rendered, 0o644)
```
taurus-forever, line 562: Q: who else should be able to read patroni.yaml? 
I mean, I can imagine group read requirements, but you are exposing read to 'other' as well.

taurus-forever (approved):
LGTM, but one critical questuon about patroni.yaml perms.

src/cluster.py
```diff
@@ -241,6 +256,38 @@ def get_primary(self, unit_name_pattern=False) -> str:
                             primary = "/".join(primary.rsplit("-", 1))
                         return primary
 
+    def get_standby_leader(
+        self, unit_name_pattern=False, check_whether_is_running: bool = False
+    ) -> str:
```
delgod, line 261: None if standby leader not found
```suggestion
    ) -> Optional[str]:
```

marceloneppel: It was updated on https://github.com/canonical/postgresql-operator/pull/452/commits/e918e4dfb7410cd9d4d48b5c8f1bc93198c0918a.

src/cluster.py
marceloneppel: There is no need for others to read. I reverted this change on https://github.com/canonical/postgresql-operator/pull/452/commits/bbd5976285a1d0072cba4f6854d531c85b3c2594.

actions.yaml
marceloneppel: Sure. I created the following ticket for that: https://warthogs.atlassian.net/browse/DPE-4257

metadata.yaml
marceloneppel: Sure. I created the following ticket for that: https://warthogs.atlassian.net/browse/DPE-4257

metadata.yaml
marceloneppel: Sure. I created the following ticket for that: https://warthogs.atlassian.net/browse/DPE-4257

dragomirp (approved):

PR #453, opened 2024-04-22
[DPE-2955] Cross-region async replication integration tests

marceloneppel:
## Issue
There are no integration tests for cross-region async replication.

## Solution
Add integration tests for the following scenarios:
* Set up async replication between two clusters.
* Trigger a switchover.
* Detach a cluster (turning it read-only).
* Reestablish the async replication.
* Crash the synchronous standby in the primary cluster.
* Crash the Standby Leader in the standby cluster.
* Scale the clusters.

Some of the helper functions were changed to accommodate checks on multiple models, like when checking whether writes are increasing (to now check it in both clusters).


taurus-forever (approved):
Nice. LGTM!

At a followup PR we IMHO need to add test for https://warthogs.atlassian.net/browse/DPE-4238 (A=>B, B=>C) as it is a recovery scenario for the lost cluster A.

dragomirp (approved):

PR #455, opened 2024-04-23
[MISC] Update data-platform-libs: data_interfaces to 34

lucasgameiroborges:
The libpatch 34 version of the lib had secret refactoring/fixes that required adapting the unit tests. Summary:

* `data_interfaces` lib -> regular bump
* `src/charm.py` -> Porting the exact same changes for get_secret and set_secret as done previously in PGbouncer.
* `test_charm.py` -> after review feedback, reverted previous changes to test structure. Kept only small (but necessary) changes: variable naming (to correspond to operator_password as used here in Postgres, for example), using fixtures to skip tests instead of early return, etc. test_set_secret_in_databag being juju-2 specific (otherwise it fails on juju-3).
* `test_db.py` -> Explained in earlier comment https://github.com/canonical/postgresql-operator/pull/455#discussion_r1584554717

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 15
+LIBPATCH = 16
 
 PYDEPS = ["pydantic>=1.10,<2", "poetry-core"]
 
@@ -501,7 +501,7 @@ class DataUpgrade(Object, ABC):
 
     STATES = ["recovery", "failed", "idle", "ready", "upgrading", "completed"]
 
-    on = UpgradeEvents()  # pyright: ignore [reportGeneralTypeIssues]
+    on = UpgradeEvents()  # pyright: ignore [reportAssignmentType]
 
     def __init__(
         self,
@@ -605,6 +605,21 @@ class DataUpgrade(Object, ABC):
 
         self.peer_relation.data[self.charm.app].update({"upgrade-stack": json.dumps(stack)})
         self._upgrade_stack = stack
+
+    @property
+    def other_unit_states(self) -> list:
+        """Current upgrade state for other units.
+
+        Returns:
+            Unsorted list of upgrade states for other units.
+        """
+        if not self.peer_relation:
+            return []
+
+        return [
+            self.peer_relation.data[unit].get("state", "")
+            for unit in list(self.peer_relation.units)
+        ]
 
     @property
     def unit_states(self) -> list:
@@ -926,11 +941,8 @@ class DataUpgrade(Object, ABC):
             return
 
         if self.substrate == "vm" and self.cluster_state == "recovery":
-            # Only defer for vm, that will set unit states to "ready" on upgrade-charm
-            # on k8s only the upgrading unit will receive the upgrade-charm event
-            # and deferring will prevent the upgrade stack from being popped
-            logger.debug("Cluster in recovery, deferring...")
-            event.defer()
+            # skip run while in recovery. The event will be retrigged when the cluster is ready
+            logger.debug("Cluster in recovery, skip...")
             return
 
         # if all units completed, mark as complete
@@ -981,6 +993,7 @@ class DataUpgrade(Object, ABC):
             self.charm.unit == top_unit
             and top_state in ["ready", "upgrading"]
             and self.cluster_state == "ready"
+            and "upgrading" not in self.other_unit_states
         ):
             logger.debug(
                 f"{top_unit.name} is next to upgrade, emitting `upgrade_granted` event and upgrading..."
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 16
+LIBPATCH = 15
 
 PYDEPS = ["pydantic>=1.10,<2", "poetry-core"]
 
@@ -501,7 +501,7 @@ class DataUpgrade(Object, ABC):
 
     STATES = ["recovery", "failed", "idle", "ready", "upgrading", "completed"]
 
-    on = UpgradeEvents()  # pyright: ignore [reportAssignmentType]
+    on = UpgradeEvents()  # pyright: ignore [reportGeneralTypeIssues]
 
     def __init__(
         self,
@@ -605,21 +605,6 @@ class DataUpgrade(Object, ABC):
 
         self.peer_relation.data[self.charm.app].update({"upgrade-stack": json.dumps(stack)})
         self._upgrade_stack = stack
-
-    @property
-    def other_unit_states(self) -> list:
-        """Current upgrade state for other units.
-
-        Returns:
-            Unsorted list of upgrade states for other units.
-        """
-        if not self.peer_relation:
-            return []
-
-        return [
-            self.peer_relation.data[unit].get("state", "")
-            for unit in list(self.peer_relation.units)
-        ]
 
     @property
     def unit_states(self) -> list:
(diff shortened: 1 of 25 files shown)


tests/unit/test_charm.py
```diff
@@ -1875,38 +1875,20 @@ def test_delete_password(harness, _has_secrets, caplog):
         assert harness.charm.get_secret("unit", "operator-password") is None
 
         harness.set_leader(True)
-        with caplog.at_level(logging.ERROR):
```
lucasgameiroborges, line 1878: removed this checking of error logs because it would break every time the lib gets updated and the log message is changed. Instead, we should only check if we can call `remove_secret` on non-existing secrets without error

lucasgameiroborges: inspired by the approach of mysql: https://github.com/canonical/mysql-operator/pull/440

juditnovak: Nononono.

Just change it to 
```
with caplog.at_level(logging.DEBUG):
```

juditnovak (changes requested):
There should be changes on the charm code as well, see https://github.com/canonical/pgbouncer-operator/pull/207

Let me or @dragomirp know if you need any help :-)

juditnovak (changes requested):
Charm code must be adjusted, see https://github.com/canonical/pgbouncer-operator/pull/207

tests/unit/test_charm.py
```diff
@@ -1769,47 +1771,55 @@ def test_on_get_password_secrets(harness):
...
+        return
+
     with patch("charm.PostgresqlOperatorCharm._on_leader_elected"):
         rel_id = harness.model.get_relation(PEER).id
         harness.set_leader()
 
         # Test application scope.
-        assert "password" not in harness.get_relation_data(rel_id, harness.charm.app.name)
-        harness.charm.set_secret("app", "password", "test-password")
+        assert "operator_password" not in harness.get_relation_data(rel_id, harness.charm.app.name)
```
marceloneppel, line 1799: Is there any special reason to test with the text having an underscore instead of a hyphen? I want to know if I missed something about a specific test case (maybe testing a migration from underscore to hyphen).

tests/unit/test_charm.py
```diff
@@ -1769,47 +1771,55 @@ def test_on_get_password_secrets(harness):
...
+        harness.charm.set_secret(scope, field, "test")
+        assert harness.charm.get_secret(scope, field) == "test"
 
 
 @patch_network_get(private_address="1.1.1.1")
-def test_set_secret(harness):
+def test_set_secret(harness, _has_secrets):
+    # after new data_interfaces version, every secret we set using set_secret
+    # should not go to the databag, but to juju secrets instead.
+    if _has_secrets:
```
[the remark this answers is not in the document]
marceloneppel: In the past, before secrets were available, this test was intended to check relation data because the secrets were exactly there. IMHO, we can now update the test by using `get_secret` to check whether the operations worked correctly.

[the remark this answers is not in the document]
lucasgameiroborges: If we change the `get_relation_data` in this test for `get_secret`, then it would become pretty much a copy of `test_set_reset_new_secret`, right? (setting with `set_secret` and asserting with `get_secret`). Do we want to do this?

tests/unit/test_charm.py
lucasgameiroborges: AFAIK this underscore use is needed when we want `get_secret` to return what's stored in the databag instead of the secret, as per [here](https://github.com/canonical/postgresql-operator/blob/f0fc78509468c0d14a3e7329a1c3d74f76e53b2c/src/charm.py#L257-L261). But you are right, in this specific test it was not necessary to have it. Thanks!

tests/unit/test_charm.py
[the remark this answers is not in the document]
lucasgameiroborges: Actually, I think I got what you mean, we could use get_secret on the specific asserts using the relation data, now I see. Adapted the code, thanks!

marceloneppel (approved):

taurus-forever (approved):

lib/charms/data_platform_libs/v0/upgrade.py
dragomirp, line 1: I think the upgrade rollback tests break due to the upgrade lib update.

lucasgameiroborges: though this can be true, I'm seeing failures on [test_db](https://github.com/canonical/postgresql-operator/actions/runs/8850106249/job/24370477014) and [test_relations](https://github.com/canonical/postgresql-operator/actions/runs/8850106249/job/24370475545) which seem to be related to the data interfaces lib update.

tests/unit/test_db.py
```diff
@@ -477,23 +476,8 @@ def test_update_endpoints_with_relation(harness):
...
-            )
-            harness.update_relation_data(
-                peer_rel_id,
-                harness.charm.app.name,
-                {
-                    user: password,
-                    f"{user}-database": DATABASE,
-                },
-            )
+            harness.charm.set_secret("app", user, password)
```
lucasgameiroborges, line 479: i re-wrote this test so it betters reflect the actual way this variables are set inside the charm (using `set_secret` instead of `update_relation_data`) [here](https://github.com/canonical/postgresql-operator/blob/b4dcc6af9379b98f40cd604a9928d765ee81175f/src/relations/db.py#L198-L201)

lib/charms/data_platform_libs/v0/upgrade.py
lucasgameiroborges: those seem to have been fixed! reverting the upgrade lib to its original version did the trick as well

tests/unit/test_charm.py
```diff
@@ -1769,21 +1771,25 @@ def test_on_get_password_secrets(harness):
...
+        assert harness.charm.get_secret(scope, field) == "test"
 
 
 @patch_network_get(private_address="1.1.1.1")
-def test_set_secret(harness):
+def test_set_secret_in_databag(harness, _has_secrets):
+    # As this test set secrets using set_secret and checks if they end up
+    # inside the relation databag, this is meant to run on juju2 only
+    if _has_secrets:
+        return
```
juditnovak, line 1792: Pls don't do this.

Tests must be deterministic (except when clearly indcated by approriate tools).

See my comment on the same thing on https://github.com/canonical/postgresql-k8s-operator/pull/454#discussion_r1584573182

juditnovak (changes requested):

lucasgameiroborges:
After newest commits, I'll try to summarize the changes in this PR:
- The `data_interfaces` lib -> regular bump
- `src/charm.py` -> Porting the exact same changes for `get_secret` and `set_secret` as done previously in PGbouncer.
- `test_charm.py` -> after review feedback, reverted previous changes to test structure. Kept only small (but necessary) changes: variable naming (to correspond to `operator_password` as used here in PG, for example), using fixtures to skip tests instead of early return, etc. `test_set_secret_in_databag` being juju-2 specific.
- `test_db.py` -> Explained in earlier comment [here](https://github.com/canonical/postgresql-operator/pull/455#discussion_r1584554717)

I tried to make it in such a way so we can merge the library bump with only the necessary test changes.

tests/unit/test_charm.py
```diff
@@ -1719,7 +1719,7 @@ def test_scope_obj(harness):
 
 
 @patch_network_get(private_address="1.1.1.1")
-def test_get_secret(harness):
+def test_get_secret_from_databag(harness):
     with patch("charm.PostgresqlOperatorCharm._on_leader_elected"):
```
juditnovak, line 1723: Renaming is great! I'd add a docstring/comment highlighting that this test is for shared functionality between Juju2 and Juju3 as we have backwards compatibility support on reading from the databag.

tests/unit/test_charm.py
```diff
@@ -1785,11 +1785,7 @@ def test_get_secret_secrets(harness, scope, field):
 
 
 @patch_network_get(private_address="1.1.1.1")
-def test_set_secret_in_databag(harness, _has_secrets):
-    # As this test set secrets using set_secret and checks if they end up
-    # inside the relation databag, this is meant to run on juju2 only
-    if _has_secrets:
-        return
+def test_set_secret_in_databag(harness, only_without_juju_secrets):
```
juditnovak, line 1788: We should keep the comment it was very good :-)

juditnovak: Those comments that you had 1-2 commits ago were great. I think it's a very good idea to add a small explanation (docstring?) to these tests. I left them pretty confusing :-(

You don't have to add them but if you choose to do they will be helpful :-)

juditnovak (approved):

dragomirp (approved):

PR #456, opened 2024-04-25
[MISC] Rollback Patroni update

dragomirp:
Rollback to Patroni 3.1.2.

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -182,9 +182,11 @@ async def test_postgresql_parameters_change(ops_test: OpsTest) -> None:
     for unit_id in UNIT_IDS:
         host = get_unit_address(ops_test, f"{DATABASE_APP_NAME}/{unit_id}")
         logger.info("connecting to the database host: %s", host)
-        with db_connect(host, password) as connection:
-            settings_names = ["max_prepared_transactions", "shared_buffers", "lc_monetary"]
-            with connection.cursor() as cursor:
+        try:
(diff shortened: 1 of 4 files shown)


dragomirp:
I'm still seeing the /config endpoint being out of sync, but that seems to be the case with 14/stable as well. We'll have to take a closer look after the sprint.

tests/integration/test_charm.py
```diff
@@ -193,12 +195,13 @@ async def test_postgresql_parameters_change(ops_test: OpsTest) -> None:
...
-        connection.close()
 
-        # Validate each configuration set by Patroni on PostgreSQL.
-        assert settings["max_prepared_transactions"] == "100"
-        assert settings["shared_buffers"] == "128"
-        assert settings["lc_monetary"] == "en_GB.utf8"
+                # Validate each configuration set by Patroni on PostgreSQL.
+                assert settings["max_prepared_transactions"] == "100"
+                assert settings["shared_buffers"] == "128"
+                assert settings["lc_monetary"] == "en_GB.utf8"
```
dragomirp, line 202: Check each node's config values.

taurus-forever (approved):
I suspected to see here only snap revision update, but all other changes looks OK. LGTM.

tests/integration/test_charm.py
marceloneppel: Thanks for catching that and improving it!

marceloneppel (approved):
LGTM!

PR #464, opened 2024-05-08
Run integration tests on arm64

carlcsaposs-canonical:


taurus-forever (approved):

tests/integration/architecture.py
```diff
@@ -0,0 +1,7 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+import subprocess
+
+architecture = subprocess.run(
+    ["dpkg", "--print-architecture"], capture_output=True, check=True, encoding="utf-8"
+).stdout.strip()
```
dragomirp, line 7: I think we should rather use `platform.machine()` and a mapping to the correct architecture. Otherwise the tests will only run on Debian derivatives without tweaking.

tests/integration/test_tls.py
```diff
@@ -30,13 +31,18 @@
 
 APP_NAME = METADATA["name"]
 if juju_major_version < 3:
-    TLS_CERTIFICATES_APP_NAME = "tls-certificates-operator"
-    TLS_CHANNEL = "legacy/stable"
-    TLS_CONFIG = {"generate-self-signed-certificates": "true", "ca-common-name": "Test CA"}
+    if architecture == "arm64":
+        pytest.skip(allow_module_level=True)
```
dragomirp, line 35: There's an arm reversion in `legacy/edge` for the legacy tls charm.

dragomirp (approved):

tests/integration/architecture.py
dragomirp: Yes and it is what we use for installing the correct snap revision.

PR #472, opened 2024-05-20
[MISC] Add experimental max_connections config

dragomirp:
Adds experimental support for force setting max_connections.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -402,3 +402,7 @@ options:
       Allowed values are: from 0 to 2000000000.
     type: int
     default: 150000000
+  experimental_max_connections:
+    type: int
+    description: |
+      [EXPERIMENTAL] Force set max_connections.
--- a/src/charm.py
+++ b/src/charm.py
@@ -1547,8 +1547,14 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             logger.warning("Early exit update_config: Cannot connect to Postgresql")
             return False
 
+        # Use config value if set, calculate otherwise
+        if self.config.experimental_max_connections:
+            max_connections = self.config.experimental_max_connections
+        else:
+            max_connections = max(4 * os.cpu_count(), 100)
+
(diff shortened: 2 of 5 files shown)


src/charm.py
```diff
@@ -1468,7 +1468,7 @@ def _restart(self, event: RunWithLock) -> None:
             return
 
         try:
-            for attempt in Retrying(wait=wait_fixed(3), stop_after_delay=stop_after_delay(300)):
+            for attempt in Retrying(wait=wait_fixed(3), stop=stop_after_delay(300)):
```
dragomirp, line 1471: Typo™

src/charm.py
```diff
@@ -1547,8 +1547,14 @@ def update_config(self, is_creating_backup: bool = False) -> bool:
             logger.warning("Early exit update_config: Cannot connect to Postgresql")
             return False
 
+        # Use config value if set, calculate otherwise
+        if self.config.experimental_max_connections:
+            max_connections = self.config.experimental_max_connections
+        else:
+            max_connections = max(4 * os.cpu_count(), 100)
+
```
dragomirp, line 1555: Doing this only here would trigger a restart on deploy with `experimental_max_connections`.  I can also inject it in the config file, by changing the internal library, but I would rather keep it KISS.

config.yaml
```diff
@@ -402,3 +402,7 @@ options:
       Allowed values are: from 0 to 2000000000.
     type: int
     default: 150000000
+  experimental_max_connections:
+    type: int
+    description: |
+      [EXPERIMENTAL] Force set max_connections.
```
dragomirp, line 408: Should we add some kind of validation here?

lucasgameiroborges (approved):
LGTM!

config.yaml
lucasgameiroborges: should we maybe enforce an upper limit?

dragomirp: It's a footgun config.

PR #475, opened 2024-05-27
[DPE-4416] Fetch charm libs to the latest LIBPATCH (update dp-libs to v36 to fix secrets issue)

delgod:
Issue:
It is impossible to upgrade charm from the current 14/stable to 14/candidate due to the issue https://warthogs.atlassian.net/browse/DPE-4416

Solution:
Fix dp-library in revision 36 https://github.com/canonical/data-platform-libs/pull/170
Update all libraries to the latest fixes/LIBPATCH.
Original issue is no longer reproducible and the new charm passed the internal testing.

PR #478, opened 2024-05-27
[DPE-4416] Update rolling-ops lib to version 0.7

lucasgameiroborges:
## Issue
`charm-rolling-ops` update to version LIBPATCH 7 causes unit tests to break (see #475). 

Root cause: New lib version will trigger lock processing hook with every leader election event. This hook was failing in every test that triggers a leader election due to a missing relation on unit test setup.

## Solution
Harness setup was missing required `restart` relation for `RollingOpsManager`. 

## Additional Changes
Two fixes for instabilities in `test_async_replication` integration test, specific description in PR comments.

### DIFF
--- a/src/relations/async_replication.py
+++ b/src/relations/async_replication.py
@@ -487,7 +487,7 @@ class PostgreSQLAsyncReplication(Object):
     def _on_async_relation_departed(self, event: RelationDepartedEvent) -> None:
         """Set a flag to avoid setting a wrong status message on relation broken event handler."""
         # This is needed because of https://bugs.launchpad.net/juju/+bug/1979811.
-        if event.departing_unit == self.charm.unit:
+        if event.departing_unit == self.charm.unit or self.charm._peers is None:
             self.charm._peers.data[self.charm.unit].update({"departing": "True"})
 
     def _on_async_relation_joined(self, _) -> None:
--- a/src/relations/async_replication.py
+++ b/src/relations/async_replication.py
@@ -487,7 +487,7 @@ class PostgreSQLAsyncReplication(Object):
     def _on_async_relation_departed(self, event: RelationDepartedEvent) -> None:
         """Set a flag to avoid setting a wrong status message on relation broken event handler."""
         # This is needed because of https://bugs.launchpad.net/juju/+bug/1979811.
-        if event.departing_unit == self.charm.unit or self.charm._peers is None:
+        if event.departing_unit == self.charm.unit and self.charm._peers is not None:
             self.charm._peers.data[self.charm.unit].update({"departing": "True"})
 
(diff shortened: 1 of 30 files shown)


tests/integration/ha_tests/test_async_replication.py
```diff
@@ -434,7 +434,6 @@ async def test_async_replication_failover_in_main_cluster(
     # Check that the sync-standby unit is not the same as before.
     new_sync_standby = await get_sync_standby(ops_test, first_model, DATABASE_APP_NAME)
     logger.info(f"New sync-standby: {new_sync_standby}")
-    assert new_sync_standby != sync_standby, "Sync-standby is the same as before"
```
marceloneppel, line 437: I requested this change as Patroni is recovering quickly and becoming the `sync-standby` again in the former unit. We shouldn't always expect the behaviour tested on this check.

src/relations/async_replication.py
```diff
@@ -487,7 +487,7 @@ def _on_async_relation_changed(self, event: RelationChangedEvent) -> None:
     def _on_async_relation_departed(self, event: RelationDepartedEvent) -> None:
         """Set a flag to avoid setting a wrong status message on relation broken event handler."""
         # This is needed because of https://bugs.launchpad.net/juju/+bug/1979811.
-        if event.departing_unit == self.charm.unit:
+        if event.departing_unit == self.charm.unit or self.charm._peers is None:
```
marceloneppel, line 490: This additional check was previously done in the `postgresql_client` interface relations. It's needed because sometimes the peer relation is not available anymore and we need to prevent destructive actions in the `relation-broken` hook when scaling down the cluster.

marceloneppel (approved):
LGTM!

src/relations/async_replication.py
dragomirp: Shouldn't this be `self.charm._peers is not None`? If `self.charm._peers` is None, it won't have `data` attr to update.

marceloneppel: You're right, Dragomir!

taurus-forever (approved):
LGTM to rolling-ops changes, but please agree with Mykola sync_stanby simplifications.

tests/integration/ha_tests/test_async_replication.py
lucasgameiroborges: The way I see it, this specific change may be a little out of scope for this rolling-ops update PR, I propose to revert this change and create a follow-up ticket to investigate the underlying issue at play here. What do you think?

lucasgameiroborges: change reverted and follow-up ticket created https://warthogs.atlassian.net/browse/DPE-4550

marceloneppel (approved):

PR #479, opened 2024-05-28
[DPE-4462] Add Incremental+Differential backup support

lucasgameiroborges:
VM version of https://github.com/canonical/postgresql-k8s-operator/pull/476 and https://github.com/canonical/postgresql-k8s-operator/pull/487

## Follow-up

https://warthogs.atlassian.net/browse/DPE-4475 for including incremental backup check in integration test.

### DIFF
--- a/templates/pgbackrest.conf.j2
+++ b/templates/pgbackrest.conf.j2
@@ -4,6 +4,7 @@ lock-path=/tmp
 log-path={{ log_path }}
 repo1-retention-full-type=time
 repo1-retention-full={{ retention_full }}
+repo1-retention-history=365
 repo1-type=s3
 repo1-path={{ path }}
 repo1-s3-region={{ region }}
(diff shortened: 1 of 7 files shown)


templates/pgbackrest.conf.j2
```diff
@@ -4,6 +4,7 @@ lock-path=/tmp
 log-path={{ log_path }}
 repo1-retention-full-type=time
 repo1-retention-full={{ retention_full }}
+repo1-retention-history=365
```
taurus-forever, line 7: Are we wiping backups older then 1 year by default? It can be unacceptable for list of companies we know.

taurus-forever (commented):
Nice! One question about retention policy:

templates/pgbackrest.conf.j2
lucasgameiroborges: What happens here is: Whenever pgbackrest creates a backup, it will also create an associated manifest file that contains some information about the backup itself, WAL size, etc. When a backup is expired, its manifest file does not expire automatically, but is kept as a "history" of backups/WAL size growth.

This configuration specifies that, whenever a backup is expired, its associated manifest file will only expire a year later. 

Valid non-expired backups will not have their manifest files deleted, so if the user sets 10 years retention, the manifest files will stay for 11 years.

365 is a rather typical value for this config (from as far as I could see online and on the docs).

marceloneppel (approved):
LGTM!

taurus-forever (approved):

PR #483, opened 2024-06-03
[DPE-4544] Fix model switch

marceloneppel:


dragomirp (approved):

taurus-forever (approved):
I am not sure how the test passed in AMD before enabling ARM... but LGTM.

lucasgameiroborges (approved):
LGTM!

marceloneppel:
> I am not sure how the test passed in AMD before enabling ARM... but LGTM.

Before enabling ARM, we didn't have the first `juju switch` call (used to set the model constraints when in ARM), which caused the issue.

PR #487, opened 2024-06-04
[POC] Data Interfaces v38 candidate

juditnovak:


dragomirp:
Closing until @juditnovak is back.

juditnovak:
Sure.

I In fact, the candidate v38 used here was not yet final, so updates were needed here anyway :-)

PR #489, opened 2024-06-07
[MISC] Revert upgrade lib

dragomirp:
Reverting the upgrade charm lib

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 16
+LIBPATCH = 15
 
 PYDEPS = ["pydantic>=1.10,<2", "poetry-core"]
 
@@ -501,7 +501,7 @@ class DataUpgrade(Object, ABC):
 
     STATES = ["recovery", "failed", "idle", "ready", "upgrading", "completed"]
 
-    on = UpgradeEvents()  # pyright: ignore [reportAssignmentType]
+    on = UpgradeEvents()  # pyright: ignore [reportGeneralTypeIssues]
 
     def __init__(
         self,
@@ -605,21 +605,6 @@ class DataUpgrade(Object, ABC):
 
         self.peer_relation.data[self.charm.app].update({"upgrade-stack": json.dumps(stack)})
         self._upgrade_stack = stack
-
-    @property
-    def other_unit_states(self) -> list:
-        """Current upgrade state for other units.
-
-        Returns:
-            Unsorted list of upgrade states for other units.
-        """
-        if not self.peer_relation:
-            return []
-
-        return [
-            self.peer_relation.data[unit].get("state", "")
-            for unit in list(self.peer_relation.units)
-        ]
 
     @property
     def unit_states(self) -> list:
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -941,8 +926,11 @@ def on_upgrade_changed(self, event: EventBase) -> None:
             return
 
         if self.substrate == "vm" and self.cluster_state == "recovery":
-            # skip run while in recovery. The event will be retrigged when the cluster is ready
-            logger.debug("Cluster in recovery, skip...")
+            # Only defer for vm, that will set unit states to "ready" on upgrade-charm
+            # on k8s only the upgrading unit will receive the upgrade-charm event
+            # and deferring will prevent the upgrade stack from being popped
+            logger.debug("Cluster in recovery, deferring...")
+            event.defer()
```
dragomirp, line 933: I think this is the cause why the rollback gets stuck.

taurus-forever: CC: @paulomach  re: https://github.com/canonical/data-platform-libs/pull/146

taurus-forever (approved):
LGTM. sad.

Should we test PostgreSQL test on MySQL changes? :-D

lib/charms/data_platform_libs/v0/upgrade.py
paulomach: Hey @dragomirp , this was causing concurrent upgrades on units, so I don't think we should revert it on the lib.
Does manual upgrades run also stuck? 
Is the test timming out? I've also observed juju takes humongous amount of time to trigger `upgrade-charm` after a refresh being run, eating a lot of the wait period, forcing us to increase timeout times for upgrade tests.

marceloneppel: The real issue is the following:
* In the `test_upgrade_from_edge` test:
    * TimescaleDB in the snap is leading to a longer snap refresh (only some more seconds).
    * With that, Patroni is performing a failover, so a new unit, different from the Juju leader unit (the former primary), becomes the primary.
    * Also with that, the former primary unit, the Juju leader unit, becomes a replica (not a synchronous standby).
* In the `test_fail_and_rollback` test:
    * After the failed upgrade, the `pre-upgrade-check` is run, which sets the `recovery` state for the Juju leader unit (which is a replica).
    * When the refresh is started, the upgrade charm event is fired in each of the three units, setting the state to ready in all of them.
    * The upgrade changed event is fired only in the non-primary units because only the relation data from the Juju leader unit changed (the state changed from `recovery` to `ready`, while in the other units, it was already set to `ready`, so no change, no events firing for the Juju leader unit).
    * As the Juju leader unit is the replica, it’s the top of the stack unit (the one that should be upgraded first). However, as the upgrade relation changed event hasn’t fired for it, it cannot start the upgrade (in this case is a rollback).

The new version of the upgrade library (which is being reverted through this PR) is also causing this issue to start happening because of the missing defer call that Dragomir commented about.

However, the root cause is the failover happening, which has shown a specific path of the execution that wasn’t covered by the lib. Before adding TimescaleDB, there was no failover, so the Juju leader unit was the primary during the entire test, and what happened in the `test_fail_and_rollback` test was the following:
    * After the failed upgrade, the `pre-upgrade-check` is run, which sets the `recovery` state for the Juju leader unit (which is the primary).
    * When the refresh is started, the upgrade charm event is fired in each of the three units, setting the state to ready in all of them.
    * As the Juju leader unit was not the top of the stack unit, it was ok for it not to receive the upgrade relation event in the first moment.

For Kafka and MySQL operators on VM, it hasn’t happened because both always set the upgrade stack with the Juju leader unit as the last unit to be upgraded. For the following codes, remember that the `build_upgrade_stack` is always executed in the Juju leader unit.
* [Kafka code](https://github.com/canonical/kafka-operator/blob/af5fd600dd6bb6767044868df1e8d8f78a3d4b68/src/events/upgrade.py#L77)
* [MySQL code](https://github.com/canonical/mysql-operator/blob/14c06ff88c4e564cd6d098aa213bd03e78e84b52/src/upgrade.py#L97)

It may be fixed by https://github.com/canonical/data-platform-libs/pull/176.

dragomirp:
Superseded by https://github.com/canonical/postgresql-operator/pull/492

PR #507, opened 2024-06-24
[MISC] Suppress oversee users in standby clusters

dragomirp:
Stop deleting managed users without a relation. This is necessary to keep all credentials working when there is a switchover to standby cluster.

PR #509, opened 2024-06-25
[MISC] Suppress oversee user action

dragomirp:


taurus-forever:
@dragomirp abandon due to https://github.com/canonical/postgresql-operator/pull/507 ?

dragomirp:
> @dragomirp abandon due to #507 ?

It's a continuation if we decide to add the action at some point. I will rebase once #507 gets merged.

actions.yaml
```diff
@@ -62,3 +62,5 @@ set-tls-private-key:
     private-key:
       type: string
       description: The content of private key for communications with clients. Content will be auto-generated if this option is not specified.
+reenable-oversee-users:
+  description: Reenable purging of managed credentials after a standby cluster is promoted.
```
delgod, line 66: any reason why it is not run as part of `promote-to-primary` action?

dragomirp: The original issue was the promoting the replica cluster deleted users for relations to the old primary cluster, since there were no relations for them in the new primary cluster. We can't know if the client charms got rerelated to the new primary cluster, so we can't know when it's safe to reenable that functionality without user input.

dragomirp:
Drifted away. Should be redone if needed.

PR #510, opened 2024-06-25
[DPE-4759] Bump version

marceloneppel:
## Issue
`_on_upgrade_charm_check_legacy()` logic is leading to a slow bootstrap of replicas.

This method is not needed anymore (only when upgrading from revision 288).

## Solution
Bump the version in `src/dependency.json` so that after the next stable, we can bump the restrictions (`upgrade_supported` field) to allow the upgrade only from the latest stable (which will allow us to remove `_on_upgrade_charm_check_legacy()`).

### DIFF
--- a/src/dependency.json
+++ b/src/dependency.json
@@ -3,12 +3,12 @@
     "dependencies": {"pgbouncer": ">0"},
     "name": "postgresql",
     "upgrade_supported": ">0",
-    "version": "1"
+    "version": "2"
   },
   "snap": {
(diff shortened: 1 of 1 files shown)


taurus-forever (approved):

taurus-forever (commented):
2 @carlcsaposs-canonical should we use charm version `2` here or better `1.0.1` or alike , re: versioning on upgrade.

lucasgameiroborges (approved):

PR #511, opened 2024-06-25
[WIP][DPE-4751] Slice `test_backups.py` to speedup testing

lucasgameiroborges:
## Issue

Backups integration test was taking too long to finish + unstable, which means really annoying retries.

## Solution

Test got split into 2 groups:
 - group 1: `test_backup_aws` tests basic operations on AWS + switchover
 - group 2: `test_backup_gcp` tests basic operations on GCP + smaller tests `test_restore_on_new_cluster` and `test_invalid_config_and_recovery_after_fixing_it`
 
 With this, the 1h+ long `test_backup.py` got divided into two roughly 35min duration groups. The common operations used in both AWS and GCP clouds got moved to the `helpers` file
 
 Plus some bonus stabilization fixes being ported from K8s

### DIFF
--- a/tests/integration/test_backups.py
+++ b/tests/integration/test_backups.py
@@ -202,7 +202,7 @@ async def test_backup_gcp(ops_test: OpsTest, cloud_configs: Tuple[Dict, Dict], c
     await ops_test.model.remove_application(tls_certificates_app_name, block_until_done=True)
 
 
-@pytest.mark.group(3)
+@pytest.mark.group(2)
 async def test_restore_on_new_cluster(ops_test: OpsTest, github_secrets, charm) -> None:
     """Test that is possible to restore a backup to another PostgreSQL cluster."""
     previous_database_app_name = f"{DATABASE_APP_NAME}-gcp"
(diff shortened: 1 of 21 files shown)


marceloneppel (approved):
LGTM!

taurus-forever (approved):

marceloneppel (approved):

tests/integration/test_backups.py
```diff
@@ -99,277 +100,110 @@ async def cloud_configs(ops_test: OpsTest, github_secrets) -> None:
...
+    # Remove the database app.
+    await ops_test.model.remove_application(database_app_name, block_until_done=True)
 
-        # Remove the database app.
-        await ops_test.model.remove_application(database_app_name, block_until_done=True)
+    # Remove the TLS operator.
+    await ops_test.model.remove_application(tls_certificates_app_name, block_until_done=True)
+
+
+@pytest.mark.group(2)
```
dragomirp, line 179: nitpick: IIRC there's support for string groups, so we could use the cloud names  instead of 1-2

dragomirp (approved):

tests/integration/test_backups.py
lucasgameiroborges: really? Never seen it before. Will Try it out in another follow-up PR (for self_healing test). Thanks for the tip.

Issue #519, opened 2024-07-02
charm install fails if python3-jinja2 is not installed


baycarbone:
## Steps to reproduce
1. Deploy a fresh ubuntu desktop image
2. add a manual cloud
3. bootstrap juju provider on manual cloud
4. create model
5. deploy postgres charm from channel 14/stable

## Expected behavior
Installation succeeds and status in juju is all green. If a dependency is missing I would expect the charm to install it.

## Actual behavior
Installation fails and juju shows the postgres unit in error state.
From the juju logs it seems the charms expects the jinja module to be available but on the desktop image the python3-jinja2 package is not installed by default.
After installing the package manually, the installation proceeds successfuly.

## Versions

Operating system: Ubuntu 22.04.4 LTS

Juju CLI: 3.4.4-genericlinux-amd64

Juju agent: 3.4.4 
```
Model       Controller        Cloud/Region  Version  SLA          Timestamp
controller  anvil-controller  maas/default  3.4.4    unsupported  23:52:57+02:00

App          Version  Status  Scale  Charm            Channel        Rev  Exposed  Message
controller            active      1  juju-controller  3.4/stable     101  no
haproxy               active      1  haproxy          latest/stable   79  no       Unit is ready
maas-agent   35359    active      1  maas-agent       latest/edge      9  no
maas-region  35359    active      1  maas-region      latest/edge      9  no
postgresql   14.11    active      1  postgresql       14/stable      429  no

Unit            Workload  Agent  Machine  Public address  Ports                                                                               Message
controller/0*   active    idle   0        172.27.40.1
haproxy/0*      active    idle   0        172.27.40.1     80/tcp                                                                              Unit is ready
maas-agent/0*   active    idle   0        172.27.40.1
maas-region/0*  active    idle   0        172.27.40.1     53,3128,5239-5247,5250-5274,5280-5284,5443,8000/tcp 53,67,69,123,323,5241-5247/udp
postgresql/0*   active    idle   0        172.27.40.1     5432/tcp                                                                            Primary

Machine  State    Address      Inst id  Base          AZ  Message
0        started  172.27.40.1  manual:  ubuntu@22.04      Manually provisioned machine
```

Charm revision: 429

## Log output

Juju debug log: 
```
2024-07-02 21:41:31 INFO juju.worker.uniter uniter.go:363 unit "postgresql/0" started
2024-07-02 21:41:31 INFO juju.worker.uniter uniter.go:689 resuming charm install
[...]

PR #520, opened 2024-07-03
[DPE-4816] Add jinja2 as a dependency

dragomirp:
Make sure the charm has jinja2 installed

Related to https://github.com/canonical/postgresql-operator/issues/519


lucasgameiroborges (approved):
LGTM!
Who knows how many other packages we just expect to exist in the host system, like this one? Is there a way of reliably check this kind of thing? maybe on CI?

dragomirp:
> Who knows how many other packages we just expect to exist in the host system, like this one? Is there a way of reliably check this kind of thing? maybe on CI?

I think jinja was a known and special case, but I can't remember why we kept on using the system installed one.

marceloneppel (approved):

PR #533, opened 2024-07-17
[MISC] Always check databag

dragomirp:
Fix a missing case for peer to secrets translation.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -294,9 +294,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             return None
         secret_key = self._translate_field_to_secret_key(key)
         # Old translation in databag is to be taken
-        if key != secret_key and (
-            result := self.peer_relation_data(scope).fetch_my_relation_field(peers.id, key)
-        ):
+        if result := self.peer_relation_data(scope).fetch_my_relation_field(peers.id, key):
             return result
 
         return self.peer_relation_data(scope).get_secret(peers.id, secret_key)
@@ -312,10 +310,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         peers = self.model.get_relation(PEER)
         secret_key = self._translate_field_to_secret_key(key)
         # Old translation in databag is to be deleted
-        if key != secret_key and self.peer_relation_data(scope).fetch_my_relation_field(
-            peers.id, key
-        ):
-            self.peer_relation_data(scope).delete_relation_data(peers.id, [key])
+        self.peer_relation_data(scope).delete_relation_data(peers.id, [key])
         self.peer_relation_data(scope).set_secret(peers.id, secret_key, value)
 
     def remove_secret(self, scope: Scopes, key: str) -> None:


src/charm.py
```diff
@@ -294,9 +294,7 @@ def get_secret(self, scope: Scopes, key: str) -> Optional[str]:
             return None
         secret_key = self._translate_field_to_secret_key(key)
         # Old translation in databag is to be taken
-        if key != secret_key and (
-            result := self.peer_relation_data(scope).fetch_my_relation_field(peers.id, key)
-        ):
```
dragomirp, line 299: If key == secret_key we ignored that peer data value

delgod (approved):

src/charm.py
```diff
@@ -294,9 +294,7 @@ def get_secret(self, scope: Scopes, key: str) -> Optional[str]:
...
+        if result := self.peer_relation_data(scope).fetch_my_relation_field(peers.id, key):
             return result
 
         return self.peer_relation_data(scope).get_secret(peers.id, secret_key)
```
juditnovak, line 300: Hmmm..... This line WAS supposed to be checking the databag.

juditnovak: This IS the issue.

https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py#L1873-L1877

juditnovak: Well... Having said all that...

It becomes a conceptual question.

Should the lib's `get_secret()` return secret fields ONLY?
Or should it return the key requested _also_ in case it's in the databag?

(I consider the current situation a bug on my side, as my original intent was along the lines of the latter. However once we are at this point, I think the question is valid regarding policy.)

juditnovak: @marceloneppel :point_up:

src/charm.py
lucasgameiroborges: I'm a bit confused: If key == secret_key, then this means we should check inside the secret, not the databag (wasn't this the original intention of this code?). Then how come we can have key == secret_key and still have to check the databag? Doesn´t this mean something went wrong when setting the secret in the first place? Sorry I think I never understood this workflow properly :grimacing:

lucasgameiroborges: I suppose the bug itself is not clear to me.

lucasgameiroborges (approved):
LGTM, just left some comments for my own understanding.

src/charm.py
dragomirp: Migrating from Juju 2 to Juju 3 would switch the charm from trying to use secrets, but all the values are still in the peer databag. The issue here is that we only check if the `key` and `secret_key` are different, but the data would still be in the peer databag even if the key is the same (no translation needed).

src/charm.py
marceloneppel: In my mind, we'd expect that lib's `get_secret()` would return secret fields only if Juju Secrets are enabled (Juju 3+). It would return the databag field value only on Juju 2 (it's still sensitive data, even if not protected by any means). So, the code in the lib is right; no bug there.

IMO, the issue here may be fixed with Dragomir changes.

marceloneppel (approved):

src/charm.py
juditnovak: If the lib `get_secret()` function _should_ return requested fields on Juju2 (we have no way to tell if a field in Juju2 is sensitive or not, so we could just return the value if found)... Then the lib is NOT doing the right thing.

Currently lib's `get_secret()` is NOT doing that. Due to the snippet (condition) I've highlighted. We don't return anything, if not Juju3 at all.

Also, if this is the preferred direction, @dragomirp 's code (that's I'm afraid is also mine :sweat_smile: ) is NOT wrong here.

marceloneppel: Isn't `self.secrets_enabled` from the condition preventing Juju from returning nothing?

src/charm.py
lucasgameiroborges: Thanks for the explanation. Looks like a specific edge case for which we didn´t have any tests.

src/charm.py
marceloneppel: I think it can be kept here.

juditnovak: So.

We should deal with the translarion.
But no more.

Just like in MySQL: https://github.com/canonical/mysql-k8s-operator/blob/main/lib/charms/mysql/v0/mysql.py#L742-L745

We should check for the new key (using lib `get_secret()` but thechnically `fetch_relation_data()` can work as well).
Then we should do the same for the old name.

The current "fix" is covering up for the bug on the libs side. IT's not the way the code should be wrtiten ideally.

Huh I don't know if I made my point clear :sweat_smile:

marceloneppel: I'm okay with going the way you commented if it avoids hiding the bug.

PR #551, opened 2024-07-26
[MISC] Show start/end time in UTC time in list-backups output

lucasgameiroborges:
## Issue
Output of command was converting timezones to local time, which can generate confusion + unit test was failing if local timezone not UTC.

Thanks to @dragomirp for raising this issue.


dragomirp (approved):

marceloneppel (approved):
Thanks!

taurus-forever (approved):

PR #554, opened 2024-07-27
[MISC] Remove Sentry test

dragomirp:
* Remove sentry test
* Mark livepatch test as unstable


tests/integration/test_db.py
```diff
@@ -196,85 +194,6 @@ async def test_relation_data_is_updated_correctly_when_scaling(ops_test: OpsTest
                 psycopg2.connect(primary_connection_string)
 
 
-@markers.juju2
-@pytest.mark.group(1)
-@markers.amd64_only  # sentry snap not available for arm64
-async def test_sentry_db_blocked(ops_test: OpsTest, charm: str) -> None:
```
dragomirp, line 202: Test fails consistently with:
```
juju.errors.JujuAPIError: cannot add relation "haproxy:reverseproxy sentry1:http": unknown series for version: "24.04"
```
Since the sentry charm doesn't seem to be maintained I propose removing the test.

tests/integration/test_db.py
```diff
@@ -333,7 +252,6 @@ async def test_roles_blocking(ops_test: OpsTest, charm: str) -> None:
     )
 
 
-@markers.juju3  # As the Sentry test  already checks Juju 2.
```
dragomirp, line 336: Enable the test-app-extensions blocking on both juju2 and juju3.

tests/integration/test_db.py
```diff
@@ -377,6 +295,7 @@ async def test_extensions_blocking(ops_test: OpsTest, charm: str) -> None:
 
 @markers.juju2
 @pytest.mark.group(1)
+@pytest.mark.unstable
```
dragomirp, line 298: Test fails consistently with:
```
AttributeError: 'NoneType' object has no attribute 'workload_status_message'
```
There doesn't seem to be a livepatch unit deployed. We should investigate further and reenable.

lucasgameiroborges (approved):
Thanks!

marceloneppel (approved):

tests/integration/test_db.py
dragomirp: Created [DPE-4989](https://warthogs.atlassian.net/browse/DPE-4989) to investigate and fix.

marceloneppel: Thanks!

PR #559, opened 2024-07-30
[DPE-4198] Switch to constant locales

dragomirp:
Hardcode the locales, so that there's no need to get them from the snap all the time.


src/locales.py
dragomirp, line 1: Essentially the sorted output of `ls /snap/charmed-postgresql/current/usr/lib/locale` + `C`

src/config.py
```diff
@@ -90,9 +91,9 @@ class CharmConfig(BaseConfigModel):
     request_standard_conforming_strings: Optional[bool]
     request_time_zone: Optional[str]
     response_bytea_output: Optional[str]
-    response_lc_monetary: Optional[str]
-    response_lc_numeric: Optional[str]
-    response_lc_time: Optional[str]
+    response_lc_monetary: Optional[Literal[tuple(SNAP_LOCALES)]]
+    response_lc_numeric: Optional[Literal[tuple(SNAP_LOCALES)]]
+    response_lc_time: Optional[Literal[tuple(SNAP_LOCALES)]]
```
dragomirp, line 96: I don't think we need the validator.

tests/integration/test_charm.py
```diff
@@ -166,6 +169,23 @@ async def test_settings_are_correct(ops_test: OpsTest, unit_id: int):
...
+    locales = raw_locales.splitlines()
+    locales.append("C")
+    locales.sort()
+
+    # Juju 2 has an extra empty element
+    if "" in locales:
+        locales.remove("")
+    assert locales == SNAP_LOCALES
+
+
```
dragomirp, line 188: This should fail if the snap locales change after an update, so we can regenerate the SNAP_LOCALES

tests/unit/test_charm.py
```diff
@@ -1429,22 +1429,35 @@ def test_validate_config_options(harness):
 
         with pytest.raises(ValueError) as e:
             harness.charm._validate_config_options()
-            assert e.msg == "request_date_style config option has an invalid value"
```
dragomirp, line 1432: The context exists when the exception is raised

tests/unit/test_charm.py
```diff
@@ -1848,11 +1861,8 @@ def test_add_cluster_member(harness):
 
         # Not ready error if not all members are ready
         _are_all_members_ready.return_value = False
-        try:
+        with pytest.raises(NotReadyError):
             harness.charm.add_cluster_member("postgresql/0")
-            assert False
```
dragomirp, line 1853: Should we keep the assert here?

src/locales.py
dragomirp, line 1: Essentially the sorted output of `ls /snap/charmed-postgresql/current/usr/lib/locale` + `C`

taurus-forever (commented):
I am not happy of hardcoding the highly changeable locales file locally. 
It is an extra efforts we are going to handle forever. And we will forget to update it.

Why don't we simply defer leader-elected if locale file is missing... OR use default `C.UTF-8` locale if snap failed to give us /snap/charmed-postgresql/current/usr/lib/locale


IMHO, it is foreign responsibility we are workarounding for no real reason.

dragomirp:
> I am not happy of hardcoding the highly changeable locales file locally. It is an extra efforts we are going to handle forever. And we will forget to update it.

I don't think the locale names will change all that often and they should only change when we rebuild the snap. If we forget to update the list, the integration test check will fail, since the hardcoded list and what's in the snap will differ.

> Why don't we simply defer leader-elected if locale file is missing... OR use default `C.UTF-8` locale if snap failed to give us /snap/charmed-postgresql/current/usr/lib/locale

The original issue can potentially happen not only when we actually set a locale, but when we construct the validator for the locales. If `ls` can't execute at that time, we have to catch the exception and defer the given event. I don't think that leader-elected is the only place this can happen.

tests/integration/test_charm.py
lucasgameiroborges: If the `ls` command was failing/unstable before (when running inside the leader elected hook), wouldn't this test potentially be unstable for the same underlying reason? Shouldn't we maybe retry this command in case of failure? No hard opinion here, just worrying for instabilities :sweat_smile:

lucasgameiroborges (approved):
I personally don't think the hard-coded list will be that much trouble, the test should warn us in case of issues, so LGTM. Only semi-worried about a potentially unstable test, does the original issue came up very often? I hope not, otherwise it could be +1 unstable test :pray:

tests/integration/test_charm.py
dragomirp: > If the `ls` command was failing/unstable before (when running inside the leader elected hook), wouldn't this test potentially be unstable for the same underlying reason? Shouldn't we maybe retry this command in case of failure? No hard opinion here, just worrying for instabilities 😅

The original issue happens if we try to validate the config model while the snap is not yet available/failed to install, so we can't list the locales available within the snap. If we are reaching this far into the test suite, the snap has already been installed, since the cluster has already bootstrapped.

lucasgameiroborges: Now I get the timing! Sounds good then

marceloneppel (approved):
LGTM!

taurus-forever (approved):
Discussed a lot on sync call. Necessary evil. Sold for now.

PR #565, opened 2024-08-02
[DPE-5029] Wait for exact number of units after scale down

marceloneppel:
## Issue
The `test_async_replication_failover_in_main_cluster` test often fails with the following error:
```sh
File "/home/runner/actions-runner/_work/postgresql-operator/postgresql-operator/tests/integration/ha_tests/test_async_replication.py", line 489, in test_async_replication_failover_in_main_cluster
    assert new_sync_standby != sync_standby, "Sync-standby is the same as before"
AssertionError: Sync-standby is the same as before
assert 'postgresql-1' != 'postgresql-1'
```

## Solution
Wait for the scale in operation to finish before checking whether the sync standby changed.


dragomirp (approved):

taurus-forever (approved):
Nice catch!

Issue #566, opened 2024-08-02
`Failed to retrieve the PostgreSQL version to initialise/update db-admin relation` during failover


nobuto-m:
## Steps to reproduce

1. deploy landscape stable bundle
   ```
   $ juju deploy landscape-scalable
   Located bundle "landscape-scalable" in charm-hub, revision 33
   ```
2. scale postgresql to 2 units (primary + replica)
   `$ juju add-unit postgresql -n 1`

3. take down the primary unit and trigger the failover

## Expected behavior

The failover succeeds, so the replica node will be promoted to primary. Then, the consumer of postgresql will be notified to write a new configuration file with the new primary node.

## Actual behavior

The unit gets stuck at blocked `Failed to retrieve the PostgreSQL version to initialise/update db-admin relation`

```
$ juju status
Model      Controller            Cloud/Region       Version  SLA          Timestamp
landscape  mysunbeam-controller  mysunbeam/default  3.5.3    unsupported  15:37:11Z

App               Version  Status  Scale  Charm             Channel        Rev  Exposed  Message
haproxy                    active      1  haproxy           latest/stable   75  yes      Unit is ready
landscape-server           active      1  landscape-server  latest/stable  111  no       Unit is ready
postgresql        14.11    active    1/2  postgresql        14/stable      429  no       
rabbitmq-server   3.9.13   active      1  rabbitmq-server   3.9/stable     188  no       Unit is ready

Unit                 Workload  Agent  Machine  Public address   Ports           Message
haproxy/0*           active    idle   0        192.168.151.106  80,443/tcp      Unit is ready
landscape-server/0*  active    idle   1        192.168.151.107                  Unit is ready
postgresql/0         unknown   lost   2        192.168.151.108  5432/tcp        agent lost, see 'juju show-status-log postgresql/0'
postgresql/1*        blocked   idle   4        192.168.151.110  5432/tcp        Failed to retrieve the PostgreSQL version to initialise/update db-admin relation
rabbitmq-server/0*   active    idle   3        192.168.151.109  5672,15672/tcp  Unit is ready

Machine  State    Address          Inst id    Base          AZ       Message
0        started  192.168.151.106  machine-7  ubuntu@22.04  default  Deployed
1        started  192.168.151.107  machine-8  ubuntu@22.04  default  Deployed
2        down     192.168.151.108  machine-1  ubuntu@22.04  default  Deployed
3        started  192.168.151.109  machine-2  ubuntu@22.04  default  Deployed
[...]

Issue #352, opened 2024-02-13
Charm blocks legacy relation if it cannot connect to database


dragomirp:
Updating the endpoints will [block](https://github.com/canonical/postgresql-operator/blob/main/src/relations/db.py#L294) the relation if it cannot connect and get the version of the database. I don't think we can clear the status. We should most probably retry the hook until connectivity recovers.

marceloneppel:
Yeah, that blocked status is not being cleared. We should work on something like you commented on.

taurus-forever:
@dragomirp the initial links points nowhere. Did we remove those blocking status?

dragomirp:
> @dragomirp the initial links points nowhere. Did we remove those blocking status?

No, it's still here: https://github.com/canonical/postgresql-operator/blob/8c47bae588ba2c77c95a33f613a6d2ebe82197a9/src/relations/db.py#L322-L329

dragomirp:
> IMHO, we should replace it to Error for now... and `juju resolve` will be a way to re-try when PG is up... WDYT?

I don't think we should be calling the DB here at all, we should get the version from the snap.

taurus-forever:
Good point. Is it safe during the charm refresh? Snap is new on the first unit / but Primary DB is still old, etc.

dragomirp:
Data is set (I'm not sure it's updated) per unit, so each unit should report its own PG version.

PR #578, opened 2024-08-09
[DPE-3562] Don't block on failure to get the db version

dragomirp:
Conditionally update the db version field in legacy relations if database is not accessible.

Related to #352 and #566

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -425,15 +425,19 @@ WHERE lomowner = (SELECT oid FROM pg_roles WHERE rolname = '{}');""".format(user
             timezones = cursor.fetchall()
             return {timezone[0] for timezone in timezones}
 
-    def get_postgresql_version(self) -> str:
+    def get_postgresql_version(self, current_host=True) -> str:
         """Returns the PostgreSQL version.
 
         Returns:
             PostgreSQL version number.
         """
+        if current_host:
+            host = self.current_host
+        else:
+            host = None
         try:
             with self._connect_to_database(
-                database_host=self.current_host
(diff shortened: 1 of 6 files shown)


src/relations/db.py
```diff
@@ -392,6 +391,8 @@ def update_endpoints(self, relation: Relation = None) -> None:
                 "state": self._get_state(),
                 "extensions": ",".join(required_extensions),
             }
+            if postgresql_version:
+                data["version"] = postgresql_version
```
dragomirp, line 395: Only add the version if we were able to get it.

src/relations/db.py
```diff
@@ -319,14 +319,14 @@ def update_endpoints(self, relation: Relation = None) -> None:
         if len(relations) == 0:
             return
 
+        postgresql_version = None
         try:
             postgresql_version = self.charm.postgresql.get_postgresql_version()
```
dragomirp, line 324: Alternatively we can use `cluster.get_postgresql_version` instead and get the version from the snap.

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -425,14 +425,20 @@ def get_postgresql_timezones(self) -> Set[str]:
...
+    def get_postgresql_version(self, current_host=True) -> str:
         """Returns the PostgreSQL version.
 
         Returns:
             PostgreSQL version number.
         """
+        if current_host:
+            host = self.current_host
+        else:
+            host = None
```
dragomirp, line 437: Use current_host by default, I think we still need the primary for pgbouncer.

taurus-forever (approved):
LGTM! Tnx!

marceloneppel (approved):
Thanks!

PR #584, opened 2024-08-14
[MISC] Move uri generation to update endpoints

dragomirp:
Move uri generation to `update_endpoints`

### DIFF
--- a/src/relations/postgresql_provider.py
+++ b/src/relations/postgresql_provider.py
@@ -104,9 +104,6 @@ class PostgreSQLProvider(Object):
             # Share the credentials with the application.
             self.database_provides.set_credentials(event.relation.id, user, password)
 
-            # Update the read/write and read-only endpoints.
-            self.update_endpoints(event)
-
             # Set the database version.
             self.database_provides.set_version(
                 event.relation.id, self.charm.postgresql.get_postgresql_version()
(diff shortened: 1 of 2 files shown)


src/relations/postgresql_provider.py
```diff
@@ -115,11 +112,8 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
             # Set the database name
             self.database_provides.set_database(event.relation.id, database)
 
-            # Set connection string URI.
-            self.database_provides.set_uris(
-                event.relation.id,
-                f"postgresql://{user}:{password}@{self.charm.primary_endpoint}:{DATABASE_PORT}/{database}",
-            )
```
dragomirp, line 122: `_on_database_requested` is ran only when the db is requested, so the uri won't be updated if there's a switchover.

src/relations/postgresql_provider.py
```diff
@@ -196,19 +194,29 @@ def update_endpoints(self, event: DatabaseRequestedEvent = None) -> None:
             else ""
         )
 
-        for relation in relations:
+        for relation_id in rel_data.keys():
+            user = f"relation-{relation_id}"
+            database = rel_data[relation_id].get("database")
+            password = secret_data.get(relation_id, {}).get("password")
```
dragomirp, line 200: This will refetch all the secrets.

taurus-forever (approved):
Nice fix! CC: @lucasgameiroborges

marceloneppel (approved):

Issue #27, opened 2022-08-29
Secure Patroni REST API from unauthorized access


marceloneppel:
Secure the Patroni REST API to prevent misuse, like the commented example [here](https://github.com/canonical/postgresql-k8s-operator/pull/26#discussion_r955963163).

taurus-forever:
@marceloneppel https://www.percona.com/blog/securing-patroni-rest-api-end-points-part-1/

taurus-forever:
This issue will be addressed in https://warthogs.atlassian.net/browse/DPE-5269

PR #596, opened 2024-08-27
[DPE-5269] Add passwords to Syncobj and Patroni

dragomirp:
* Add passwords to Patroni and Syncobj
* Update libs

Closes https://github.com/canonical/postgresql-operator/issues/27

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -27,7 +27,7 @@ get-password:
     username:
       type: string
       description: The username, the default value 'operator'.
-        Possible values - operator, replication, rewind.
+        Possible values - operator, replication, rewind, patroni.
 list-backups:
   description: Lists backups in s3 storage.
 pre-upgrade-check:
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -93,6 +93,8 @@ class Patroni:
         replication_password: str,
         rewind_password: str,
         tls_enabled: bool,
+        raft_password: str,
+        patroni_password: str,
     ):
         """Initialize the Patroni class.
 
@@ -107,6 +109,8 @@ class Patroni:
             replication_password: password for the user used in the replication
             rewind_password: password for the user used on rewinds
             tls_enabled: whether TLS is enabled
+            raft_password: password for raft
+            patroni_password: password for the user used on patroni
         """
         self.charm = charm
         self.unit_ip = unit_ip
@@ -117,11 +121,17 @@ class Patroni:
         self.superuser_password = superuser_password
         self.replication_password = replication_password
         self.rewind_password = rewind_password
+        self.raft_password = raft_password
+        self.patroni_password = patroni_password
         self.tls_enabled = tls_enabled
         # Variable mapping to requests library verify parameter.
         # The CA bundle file is used to validate the server certificate when
         # TLS is enabled, otherwise True is set because it's the default value.
         self.verify = f"{PATRONI_CONF_PATH}/{TLS_CA_FILE}" if tls_enabled else True
+
+    @property
+    def _patroni_auth(self) -> requests.auth.HTTPBasicAuth:
+        return requests.auth.HTTPBasicAuth("operator", self.patroni_password)
 
     @property
     def _patroni_url(self) -> str:
@@ -164,6 +174,7 @@ class Patroni:
             f"{self._patroni_url}/{PATRONI_CLUSTER_STATUS_ENDPOINT}",
             verify=self.verify,
             timeout=API_REQUEST_TIMEOUT,
+            auth=self._patroni_auth,
         )
         return {member["name"] for member in cluster_status.json()["members"]}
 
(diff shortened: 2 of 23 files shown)


templates/patroni.yml.j2
```diff
@@ -21,6 +21,9 @@ log:
 restapi:
   listen: '{{ self_ip }}:8008'
   connect_address: '{{ self_ip }}:8008'
+  authentication:
+    username: operator
+    password: {{ patroni_password }}
```
delgod, line 26: let's do it consistently to avoid confusion e.g. why operator_password doesn't work for the operator user in patroni
```suggestion
    username: patroni
    password: {{ patroni_password }}
```

delgod (changes requested):

actions.yaml
```diff
@@ -27,7 +27,7 @@ get-password:
     username:
       type: string
       description: The username, the default value 'operator'.
-        Possible values - operator, replication, rewind.
+        Possible values - operator, replication, rewind, patroni.
```
dragomirp, line 30: Users and tests need to be able to get at least the patroni password.

src/cluster.py
```diff
@@ -164,6 +174,7 @@ def cluster_members(self) -> set:
             f"{self._patroni_url}/{PATRONI_CLUSTER_STATUS_ENDPOINT}",
             verify=self.verify,
             timeout=API_REQUEST_TIMEOUT,
+            auth=self._patroni_auth,
```
dragomirp, line 177: GETs are usually open, but adding for consistency.

src/cluster_topology_observer.py
```diff
@@ -129,14 +130,16 @@ def main():
 
     Watch the Patroni API cluster info. When changes are detected, dispatch the change event.
     """
-    patroni_url, verify, run_cmd, unit, charm_dir = sys.argv[1:]
+    patroni_url, verify, run_cmd, unit, charm_dir, patroni_password = sys.argv[1:]
```
dragomirp, line 133: Passing the pass as an argument, which is not very safe. In principle, the old behaviour should keep on working as long as patroni doesn't protect the get calls. Should we keep the auth here?

src/relations/async_replication.py
```diff
@@ -302,7 +302,8 @@ def _get_secret(self) -> Secret:
             logger.debug("Secret not found, creating a new one")
             pass
 
-        return self.charm.model.app.add_secret(content=shared_content, label=SECRET_LABEL)
+        if self.charm.unit.is_leader():
+            return self.charm.model.app.add_secret(content=shared_content, label=SECRET_LABEL)
```
dragomirp, line 306: Started hitting this on follower units

delgod (approved):

src/upgrade.py
```diff
@@ -86,6 +93,14 @@ def _on_upgrade_charm_check_legacy(self) -> None:
...
             return
 
+        if self.charm.unit.is_leader() and self.charm._peers:
+            for key in (
+                RAFT_PASSWORD_KEY,
+                PATRONI_PASSWORD_KEY,
+            ):
+                if self.charm.get_secret(APP_SCOPE, key) is None:
+                    self.charm.set_secret(APP_SCOPE, key, new_password())
+
```
dragomirp, line 103: Is this the correct place for generating the passes?

taurus-forever: ^ @marceloneppel

taurus-forever (approved):
LGTM. Thanl you for the quick fix.
Let's wait @marceloneppel before merging.
P.S. we need to backport Patroni part to K8s charm :-(

src/upgrade.py
marceloneppel: Yeah, it's ok to put it here.

marceloneppel (approved):
LGTM! Thanks @dragomirp!

lucasgameiroborges (approved):

src/cluster.py
```diff
@@ -406,8 +424,21 @@ def is_replication_healthy(self) -> bool:
...
+                        member_status = requests.get(
+                            f"{url}/{endpoint}", verify=self.verify, auth=self._patroni_auth
+                        )
+                        # If raft is getting encrypted some of the calls will fail
+                        if member_status.status_code == 503 and raft_encryption:
+                            logger.warning(
+                                "Failed replication check for %s during raft encryption"
+                                % members_ip
+                            )
+                            continue
```
dragomirp, line 436: Still executing to have warnings in the log

src/upgrade.py
```diff
@@ -143,6 +158,14 @@ def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
...
         self.charm._setup_exporter()
         self.charm.backup.start_stop_pgbackrest_service()
+        raft_encryption = (
+            int(
+                json.loads(self.peer_relation.data[self.charm.app].get("dependencies", "{}"))
+                .get("charm", {})
+                .get("version", 0)
+            )
+            < 3
+        )
```
dragomirp, line 168: Check for versions that don't encrypt RAFT. Seems to be only an issue for 1 (current `14/stable`) but adding both previous version just in case.

taurus-forever (approved):
LGTM, Thank you!

PR #600, opened 2024-08-28
Try to ignore upgrade replication error

dragomirp:


src/cluster.py
```diff
@@ -428,7 +428,12 @@ def is_replication_healthy(self) -> bool:
                         member_status = requests.get(
                             f"{url}/{endpoint}", verify=self.verify, auth=self._patroni_auth
                         )
-                        if member_status.status_code != 200:
+                        if member_status.status_code == 503:
+                            logger.warning(
+                                "Failed replication check for %s with code %d"
+                                % (members_ip, member_status.status_code)
+                            )
```
marceloneppel, line 435: I'd call `request.get` again here (without `auth=self._patroni_auth`) to ensure that the problem is related to a member not enabling password authentication yet instead of letting the method return `True`.

PR #604, opened 2024-08-30
[MISC] Conditional password

dragomirp:
Conditionally add REST password to patroni config

### DIFF
--- a/src/upgrade.py
+++ b/src/upgrade.py
@@ -158,14 +158,6 @@ class PostgreSQLUpgrade(DataUpgrade):
 
         self.charm._setup_exporter()
         self.charm.backup.start_stop_pgbackrest_service()
-        raft_encryption = (
-            int(
-                json.loads(self.peer_relation.data[self.charm.app].get("dependencies", "{}"))
-                .get("charm", {})
-                .get("version", 0)
-            )
-            < 3
-        )
 
         try:
             self.charm.unit.set_workload_version(
@@ -185,7 +177,7 @@ class PostgreSQLUpgrade(DataUpgrade):
                         not self.charm._patroni.member_started
                         or self.charm.unit.name.replace("/", "-")
                         not in self.charm._patroni.cluster_members
-                        or not self.charm._patroni.is_replication_healthy(raft_encryption)
+                        or not self.charm._patroni.is_replication_healthy()
                     ):
                         logger.debug(
                             "Instance not yet back in the cluster."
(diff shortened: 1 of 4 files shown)


templates/patroni.yml.j2
```diff
@@ -21,9 +21,11 @@ log:
 restapi:
   listen: '{{ self_ip }}:8008'
   connect_address: '{{ self_ip }}:8008'
+  {%- if patroni_password %}
   authentication:
     username: patroni
     password: {{ patroni_password }}
+  {%- endif %}
```
dragomirp, line 28: If the leader elected executes after the template is rendered we'll be stuck with `None` pass and will be unable to reload via the REST API

src/upgrade.py
```diff
@@ -158,14 +158,6 @@ def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
...
         self.charm._setup_exporter()
         self.charm.backup.start_stop_pgbackrest_service()
-        raft_encryption = (
-            int(
-                json.loads(self.peer_relation.data[self.charm.app].get("dependencies", "{}"))
-                .get("charm", {})
-                .get("version", 0)
-            )
-            < 3
-        )
```
dragomirp, line 168: Looks like the 503 errors were API related not RAFT related.

templates/patroni.yml.j2
taurus-forever: Can this happen in wild? What is our answer there?

dragomirp: I don't think so, but when we add rotation ([DPE-5270](https://warthogs.atlassian.net/browse/DPE-5270)) we should switch to SIGHUPing the service instead of reloading via REST.

lucasgameiroborges (approved):

templates/patroni.yml.j2
lucasgameiroborges: Basic question: At which point during charm execution the patroni template is rendered?

taurus-forever: 2 @lucasgameiroborges good to have this answer in https://charmhub.io/postgresql-k8s/docs/e-flowchart-charm for VM charm :-D

dragomirp: > Basic question: At which point during charm execution the patroni template is rendered?

I don't think there's a set order, especially during charm's bootstrap.

lucasgameiroborges: I might ask @marceloneppel once hes back, might be good to have this clear and documented somewhere (maybe on the diagrams!)

marceloneppel: I believe that the issue might be happening at https://github.com/canonical/postgresql-operator/blob/5f546f1251b21b8d7a6b04f03a8bfb71a98d383b/src/upgrade.py#L149, when a non-leader unit is the first one to be updated.

The changes from this PR LGTM.

PR #607, opened 2024-08-31
[MISC] Add back stable update workaround

dragomirp:


dragomirp:
Doesn't look like we'll need that.

PR #615, opened 2024-09-08
[DPE-5306] Data Interafces v40

juditnovak:
Data Interfaces v40 introduces a blocking mechanism, to avoid write operations on Relation Data before the relation is inialized.

This is important so that `Provider` would never try to add data before `Requirer` clearly indicating if Juju Secrest are to be used or not. (The `Provider` is adding this information within its `relation-created` handler).

NOTE: Any calls on `update_relation_data()` before this stage may appear as plain text. Thus the blocking behavior introduced.

Without fixes on the code, tests publishing `uri` premature may fail.
https://github.com/canonical/postgresql-operator/actions/runs/11002875807/job/30551092035
Note that in real life this is not a valid scenario, since `uris` contain credentials. Now, before credentials are distributed (which happens correctly in the charm, after secrets initialization) the `uris` can't be publshed. This is the proof why there has been no risk in real life usage.

### DIFF
--- a/src/relations/postgresql_provider.py
+++ b/src/relations/postgresql_provider.py
@@ -8,6 +8,7 @@ import logging
 from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseProvides,
     DatabaseRequestedEvent,
+    PrematureDataAccessError,
 )
 from charms.postgresql_k8s.v0.postgresql import (
     INVALID_EXTRA_USER_ROLE_BLOCKING_MESSAGE,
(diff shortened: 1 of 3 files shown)


src/relations/postgresql_provider.py
```diff
@@ -199,11 +200,16 @@ def update_endpoints(self, event: DatabaseRequestedEvent = None) -> None:
...
-            )
+            try:
+                # Set the read/write endpoint.
+                self.database_provides.set_endpoints(
+                    relation_id,
+                    f"{self.charm.primary_endpoint}:{DATABASE_PORT}",
+                )
+            except PrematureDataAccessError:
+                # The relation is not yet fullin initialized
+                # Updates should be performed on a later 'update-status' cycle
```
dragomirp, line 211: Better to leave a debug message in the logs.

juditnovak: Done.

src/relations/postgresql_provider.py
```diff
@@ -207,11 +208,23 @@ def update_endpoints(self, event: DatabaseRequestedEvent = None) -> None:
                 read_only_endpoints,
             )
 
-            # Set connection string URI.
-            self.database_provides.set_uris(
-                relation_id,
-                f"postgresql://{user}:{password}@{self.charm.primary_endpoint}:{DATABASE_PORT}/{database}",
-            )
```
dragomirp, line 214: Thinking about it more we can most probably just `continue` on line 198 if there's no database or password

lucasgameiroborges (approved):

dragomirp (approved):

PR #627, opened 2024-09-26
[MISC] Disable tox requirements generation for check libraries flow

dragomirp:
Check libs workflow also relied on the tox wrappers


taurus-forever (approved):

taurus-forever:
@carlcsaposs-canonical 
What about other repos? should the same PR be in https://github.com/canonical/postgresql-k8s-operator/blob/main/.github/workflows/lib-check.yaml#L39 ?

PR #645, opened 2024-10-14
[DPE-5601] Add pgBackRest logrotate configuration

marceloneppel:
## Issue
pgBackRest logs are not rotated.

## Solution
Rotate the logs by using logrotate and following the rule we have for other logs: rotate every minute.

A background process was created for that job.


dragomirp (approved):

taurus-forever (approved):
Tnx!

PR #652, opened 2024-10-17
build: use the Charmcraft poetry plugin

lengau:
This modifies charmcraft.yaml to use Charmcraft 3.3's `poetry` plugin for building the charm.

The charm is currently bigger due to duplicated files (see: https://github.com/canonical/charmcraft/issues/1966), but will be ~5 MiB smaller than the original once that's fixed, mostly due to not vendoring `pip`.


lengau:
@marceloneppel this will only be usable once Charmcraft 3.3 is out - I'll take it out of draft then.

PR #662, opened 2024-10-26
reinit PostgreSQL only if there is a healthy majority

delgod:


taurus-forever:
@delgod I thinlk this is already available inside Patroni logic: https://github.com/patroni/patroni/blob/master/patroni/ha.py#L1900-L1901

Should we copy it in the charm?

PR #679, opened 2024-11-19
[DPE-5714] Filter out degraded read only endpoints

dragomirp:
Currently, the charm sets all replica units as readonly endpoints to clients, no matter if the unit is functioning or not.

The PR adds filtering for the read only endpoints based on Patroni's cluster endpoint.

### DIFF
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -524,6 +524,23 @@ class Patroni:
 
         return len(cluster_status.json()["members"]) == 0
 
+    def are_replicas_up(self) -> dict[str, bool] | None:
+        """Check if cluster members are running or streaming."""
+        try:
+            response = requests.get(
+                f"{self._patroni_url}/cluster",
+                verify=self.verify,
+                auth=self._patroni_auth,
+                timeout=PATRONI_TIMEOUT,
+            )
+            return {
+                member["host"]: member["state"] in ["running", "streaming"]
+                for member in response.json()["members"]
+            }
+        except Exception:
+            logger.exception("Unable to get the state of the cluster")
+            return
+
     def promote_standby_cluster(self) -> None:
         """Promote a standby cluster to be a regular cluster."""
         config_response = requests.get(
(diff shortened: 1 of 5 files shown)


src/cluster.py
```diff
@@ -524,6 +524,23 @@ def is_member_isolated(self) -> bool:
...
 
+    def are_replicas_up(self) -> dict[str, bool] | None:
+        """Check if cluster members are running or streaming."""
+        try:
+            response = requests.get(
+                f"{self._patroni_url}/cluster",
+                verify=self.verify,
+                auth=self._patroni_auth,
+                timeout=PATRONI_TIMEOUT,
+            )
```
dragomirp, line 535: I think that if the majority of units goes down, the cluster endpoint will stop being updated. Best to address this after the RAFT reinit PRs land.

tests/integration/new_relations/test_new_relations.py
```diff
@@ -184,6 +194,35 @@ async def test_database_relation_with_charm_libraries(ops_test: OpsTest):
...
+
+    for attempt in Retrying(stop=stop_after_attempt(3), wait=wait_fixed(3), reraise=True):
+        with attempt:
+            data = await get_application_relation_data(
+                ops_test,
+                APPLICATION_APP_NAME,
+                FIRST_DATABASE_RELATION_NAME,
+                "read-only-endpoints",
+            )
+            assert data is None
```
dragomirp, line 220: There are two units in the cluster, so there will be no readonly endpoints if the down unit is filtered out.

marceloneppel (approved):
LGTM!

taurus-forever (approved):
LGTM, but two points:
*  Q: will we trigger DPlib events, so client app will react (restart) every time we add RO unit?
* BTW, did we agree on consensus with @delgod and @7annaba3l about adding RW as RO endpoint if the system has 1 unit only?

dragomirp:
> * Q: will we trigger DPlib events, so client app will react (restart) every time we add RO unit?

DPL should fire a separate event (`read_only_endpoints_changed`) for read only changes so it would be up to the client charm to decide.

>  * BTW, did we agree on consensus with @delgod and @7annaba3l about adding RW as RO endpoint if the system has 1 unit only?

I don't think so, we can add it in a follow up PR.

Issue #656, opened 2024-10-24
COS Integration / Some illegible sections in the dashboard


gustavosr98:
## Steps to reproduce
Follow tutorial to integrate COS and Charmed PostgreSQL

## Expected behavior
Be able to read / understand all sections in the dashboard

## Actual behavior
Some illegible sections in the dashboard
![image](https://github.com/user-attachments/assets/4a320521-d7bb-4a4c-8570-aaa4efd887e0)

See the following sections:
- Version
- Start Time
- Max connections
- Shared buffers
- Effective cache
- Maintanance Work Mem
- Work Mem
[..]

## Versions
Operating system: Ubuntu 22.04
Juju: 3.4.6-genericlinux-amd64
Charm revision: 14/edge rev 502

gustavosr98:
It seems that if I select a Juju unit I can read normally those sections
![image](https://github.com/user-attachments/assets/7418f6f0-d7ae-48d8-99e1-eb452e1033f2)

Now, this is not intuitive
I spend quite some time thinking the dashboard was broken
I just happened to realice I needed to do that becase of some screenshots I was pointed to https://github.com/canonical/postgresql-k8s-operator/pull/733 

It should either be readable from the beginning or have a doc / tutorial on how to use the dashboard (not intuitive)

lucasgameiroborges:
Hi @gustavosr98 , thank you for the report ! The issue should've been addressed with the merge of #702 , let me know if you see it again with this patch. Closing this for now, feel free to reopen if needed.

PR #702, opened 2024-12-17
[DPE-5706] remove clutter from singlestat panels in cos

lucasgameiroborges:
## Issue

This PR focuses on addressing #656 in particular. 

## Solution

This comes down to select the appropriate value (amongst the possible `juju-unit` labels) and display it on the singlestat panel. There are 3 main cases:

- *Workload versions* (PG or Patroni): Since we don't support cross-version upgrades, they should be the same across all units, so we could choose any value. Since they are not integers, using aggregation functions like `avg` or `min/max` doesn't work, hence the usage of `topk(1, <version_metric>)` to just pick a value.
- *Timestamps*: Must decide if we want to keep the most recent `max` or oldest `min` timestamp among the possible unit's values. Depends on the metric, will put comments on what was chosen and where.
- *Config values*: Should also be consistent across units 99% of the time, but can also be aggregated using `min/max`, just in case they end up being different. Here, I chose to pick the largest value by default.

Please, let me know if all this makes sense!


src/grafana_dashboards/postgresql-metrics.json
```diff
@@ -212,7 +212,7 @@
       "tableColumn": "",
       "targets": [
         {
-          "expr": "process_start_time_seconds{release=\"$release\", instance=\"$instance\"} * 1000",
+          "expr": "min(process_start_time_seconds{release=\"$release\", instance=\"$instance\"} * 1000) by (label)",
```
lucasgameiroborges, line 215: Picks the oldest start time across the selected units.

src/grafana_dashboards/postgresql-patroni-metrics.json
```diff
@@ -246,7 +246,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "patroni_dcs_last_seen{service_name=~\"$service_name\",scope=~\"$scope_name\"}*1000",
+          "expr": "max(patroni_dcs_last_seen{service_name=~\"$service_name\",scope=~\"$scope_name\"}*1000) by (label)",
```
lucasgameiroborges, line 249: Picks the most recent `Last Seen` time across selected units

src/grafana_dashboards/postgresql-patroni-metrics.json
```diff
@@ -826,7 +826,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "patroni_postmaster_start_time{service_name=~\"$service_name\",scope=~\"$scope_name\"}*1000",
+          "expr": "min(patroni_postmaster_start_time{service_name=~\"$service_name\",scope=~\"$scope_name\"}*1000) by (label)",
```
lucasgameiroborges, line 829: picks the oldest start time across selected units

marceloneppel (approved):
Those changes make sense to me.

dragomirp (approved):

PR #722, opened 2025-01-12
[MISC] Reenable db-admin test

dragomirp:


src/relations/db.py
```diff
@@ -189,13 +192,14 @@ def set_up_relation(self, relation: Relation) -> bool:
...
             password = unit_relation_databag.get("password", new_password())
 
             # Store the user, password and database name in the secret store to be accessible by
             # non-leader units when the cluster topology changes.
-            self.charm.set_secret(APP_SCOPE, user, password)
-            self.charm.set_secret(APP_SCOPE, f"{user}-database", database)
+            if password != self.charm.get_secret(APP_SCOPE, user):
+                self.charm.set_secret(APP_SCOPE, user, password)
+            if database and database != self.charm.get_secret(APP_SCOPE, f"{user}-database"):
+                self.charm.set_secret(APP_SCOPE, f"{user}-database", database)
```
dragomirp, line 202: Seems we have some kind of issue when setting multiple secret fields at the same time or when  resetting the same value.

tests/integration/helpers.py
```diff
@@ -393,6 +393,13 @@ async def deploy_and_relate_bundle_with_postgresql(
             ]
 
             # Remove PostgreSQL and relations with it from the bundle.yaml file.
+            config = data["applications"]["postgresql"]["options"]
+            if config.get("experimental_max_connections", 0) > 200:
+                config["experimental_max_connections"] = 200
+            for key, val in config.items():
+                config[key] = str(val)
+            logger.info(f"Bundle {bundle_name} needs configuration {config}")
+            await ops_test.model.applications[DATABASE_APP_NAME].set_config(config)
```
dragomirp, line 402: Get the postgresql config from the bundle

tests/integration/helpers.py
```diff
@@ -393,6 +393,13 @@ async def deploy_and_relate_bundle_with_postgresql(
             ]
 
             # Remove PostgreSQL and relations with it from the bundle.yaml file.
+            config = data["applications"]["postgresql"]["options"]
+            if config.get("experimental_max_connections", 0) > 200:
+                config["experimental_max_connections"] = 200
```
dragomirp, line 398: Landscape sets it to 500, limiting to 200

tests/integration/test_db_admin.py
```diff
@@ -103,11 +105,6 @@ async def test_landscape_scalable_bundle_db(ops_test: OpsTest, charm: str) -> No
 
     await ensure_correct_relation_data(ops_test, DATABASE_UNITS, LANDSCAPE_APP_NAME, RELATION_NAME)
 
-    # Enable automatically-retry-hooks due to https://bugs.launchpad.net/juju/+bug/1999758
-    # (the implemented workaround restarts the unit in the middle of the start hook,
-    # so the hook fails, and it's not retried on CI).
-    await ops_test.model.set_config({"automatically-retry-hooks": "true"})
-
```
dragomirp, line 110: Already set by default.

tests/integration/ha_tests/test_self_healing.py
```diff
@@ -502,12 +502,14 @@ async def test_network_cut_without_ip_change(
...
     all_units_names = [unit.name for unit in ops_test.model.applications[app].units]
-    for unit_name in set(all_units_names) - {primary_name}:
-        logger.info(f"checking for no connectivity between {primary_name} and {unit_name}")
-        hostname = await get_machine_from_unit(ops_test, unit_name)
-        assert not is_machine_reachable_from(hostname, primary_hostname), (
-            "unit is reachable from peer"
-        )
+    for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
+        with attempt:
+            for unit_name in set(all_units_names) - {primary_name}:
```
dragomirp, line 507: Seems to be timing out frequently.

tests/integration/helpers.py
taurus-forever: OK, as it is a test only in limited HW env (CI runner).

tests/integration/test_db_admin.py
```diff
@@ -45,7 +45,7 @@ async def test_landscape_scalable_bundle_db(ops_test: OpsTest, charm: str) -> No
         application_name=DATABASE_APP_NAME,
         num_units=DATABASE_UNITS,
         base=CHARM_BASE,
-        config={"profile": "testing", "plugin_plpython3u_enable": "True"},
```
taurus-forever, line 48: No longer necessary?

taurus-forever (approved):

tests/integration/test_db_admin.py
dragomirp: The exact plugins will be added and logged when we patch the bundle, so we only need the profile here.

marceloneppel (approved):

PR #726, opened 2025-01-14
[DPE-6320][DPE-6325] General secrets resetting fix

dragomirp:
A more general fix to secrets resetting

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,7 +47,8 @@ jobs:
     name: Build charm
     uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v24.0.6
     with:
-      cache: true
+      cache: false
+      artifact-prefix: test-build
 
   integration-test:
     strategy:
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -318,7 +325,7 @@ def set_secret(self, scope: Scopes, key: str, value: str | None) -> str | None:
             return None
         secret_key = self._translate_field_to_secret_key(key)
         # Old translation in databag is to be deleted
-        self.peer_relation_data(scope).delete_relation_data(peers.id, [key])
+        self.scoped_peer_data(scope).pop(key, None)
         self.peer_relation_data(scope).set_secret(peers.id, secret_key, value)
```
dragomirp, line 329: Secrets caching breaks when we try to reset the secret to the same value. `set_secret()` will not go through if it is the same value, but `delete_relation_data()` will go through and set the secret to None. Just clearing the peer data for potential legacy value should be sufficient here.

src/relations/db.py
dragomirp, line 1: Reverted, since `set_secret` should now handle resetting the secret generally.

.github/workflows/ci.yaml
dragomirp, line 1: To reenable builds

lib/charms/postgresql_k8s/v0/postgresql_tls.py
dragomirp, line 1: Reverted, since `set_secret` should now handle resetting the secret generally. Same change in https://github.com/canonical/postgresql-k8s-operator/pull/822 to publish the internal lib.

lucasgameiroborges (approved):

PR #729, opened 2025-01-16
[DPE-6377] Split topology script

dragomirp:
* Split scripts to not rely on charm venv
* Redo cluster topology observer to use standard lib


tox.ini
```diff
@@ -7,8 +7,9 @@ env_list = lint, unit
 
 [vars]
 src_path = "{tox_root}/src"
+scripts_path = "{tox_root}/scripts"
 tests_path = "{tox_root}/tests"
-all_path = {[vars]src_path} {[vars]tests_path}
+all_path = {[vars]src_path} {[vars]scripts_path} {[vars]tests_path}
```
dragomirp, line 12: For linting.

charmcraft.yaml
```diff
@@ -29,6 +29,13 @@ parts:
       craftctl default
     charm-strict-dependencies: true
     charm-requirements: [requirements.txt]
+  files:
+    plugin: dump
+    source: .
+    stage:
+      - LICENSE
+      - scripts
+      - templates
```
dragomirp, line 38: Priming conflicts on ccraft 2 over `requirements.txt`

tests/integration/new_relations/test_new_relations.py
```diff
@@ -220,7 +220,9 @@ async def test_filter_out_degraded_replicas(ops_test: OpsTest):
             assert data is None
 
     await start_machine(ops_test, machine)
-    await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=200)
+    await ops_test.model.wait_for_idle(
+        apps=[DATABASE_APP_NAME], status="active", timeout=200, raise_on_error=False
+    )
```
dragomirp, line 225: Unrelated, but the charm seems to frequently error out when the container is started: https://github.com/canonical/postgresql-operator/actions/runs/12847342758/job/35866184943

scripts/cluster_topology_observer.py
```diff
@@ -0,0 +1,61 @@
...
+from time import sleep
+from urllib.request import urlopen
+
+API_REQUEST_TIMEOUT = 5
+PATRONI_CLUSTER_STATUS_ENDPOINT = "cluster"
+
+# File path for the spawned cluster topology observer process to write logs.
+LOG_FILE_PATH = "/var/log/cluster_topology_observer.log"
+
+ssl._create_default_https_context = ssl._create_unverified_context
```
dragomirp, line 19: Disable TLS validation.

scripts/cluster_topology_observer.py
```diff
@@ -0,0 +1,61 @@
...
+# See LICENSE file for licensing details.
+
+"""Cluster topology changes observer."""
+
+import json
+import ssl
+import subprocess
+import sys
+from time import sleep
+from urllib.request import urlopen
```
dragomirp, line 11: Using standard lib instead of `requests`.

scripts/cluster_topology_observer.py
```diff
@@ -0,0 +1,61 @@
...
+    # Input is generated by the charm
+    subprocess.run([run_cmd, "-u", unit, dispatch_sub_cmd.format(charm_dir)])  # noqa: S603
+
+
+def main():
+    """Main watch and dispatch loop.
+
+    Watch the Patroni API cluster info. When changes are detected, dispatch the change event.
+    """
+    patroni_url, run_cmd, unit, charm_dir = sys.argv[1:]
```
dragomirp, line 34: Validate is removed, since cert validation is generally disabled. I don't think we need to maintain args compatibility.

scripts/__init__.py
dragomirp, line 1: To be able to import for unit tests.

dragomirp:
[Test runs](https://github.com/canonical/postgresql-operator/actions/runs/12866215514/job/35873319343?pr=732) with charmcraft 3 migration branch included

scripts/cluster_topology_observer.py
dragomirp: When TLS is enabled, Patroni's REST API will also be served over TLS with the provided certificate. As AFAIK we don't install the CA chain in the system store, the verification of the certificate will fail and should be disabled. This should override the default TLS context to not verify for the interpreter instance and should be equivalent to disabling verification in the `requests` calls in the previous script.

marceloneppel (approved):

taurus-forever (approved):
I would use a different names for cluster_top_observer.py to avoid confusion for newbies (the same tool named in two places). Anyway up-2-you here. LGTM

PR #741, opened 2025-01-24
[DPE-6056] Add preload libs to regular startup params

dragomirp:


tests/integration/ha_tests/test_upgrade_from_stable.py
```diff
@@ -22,7 +22,7 @@
 
 logger = logging.getLogger(__name__)
 
-TIMEOUT = 600
+TIMEOUT = 900
```
dragomirp, line 25: Timing out on CI e.g. https://github.com/canonical/postgresql-operator/actions/runs/12948714150/job/36124952260

marceloneppel (approved):

templates/patroni.yml.j2
```diff
@@ -140,6 +140,7 @@ postgresql:
   bin_dir: /snap/charmed-postgresql/current/usr/lib/postgresql/{{ version }}/bin
   data_dir: {{ data_path }}
   parameters:
+    shared_preload_libraries: 'timescaledb,pgaudit'
```
taurus-forever, line 143: How does CI tests passed with this line? LGTM.

taurus-forever (approved):

templates/patroni.yml.j2
dragomirp: I think the issue affects only the followers, which is why the tests pass. I think going forward we should try to move most (all?) bootstrap params into the regular params, so we have a single source of truth.

taurus-forever: ACK. the patch should work AS IS, AFAIK. Do you want to merge it or proceed with refactoring now?

dragomirp: > ACK. the patch should work AS IS, AFAIK. Do you want to merge it or proceed with refactoring now?

Let's merge this as is and leave the refactoring for better days.

PR #749, opened 2025-01-30
[MISC] Define charm utility properties

sinclert-canonical:
This PR defines a set of charm properties to reduce the chance of introducing typos when checking for specific keys in the app / unit data-bags, as well as reduce the cognitive load of logical statements.

The **new** properties are:
- `is_cluster_restoring_backup`.
- `is_cluster_restoring_to_time`.
- `is_unit_departing`
- `is_unit_stopped`.

The **existing** properties, now used throughout the codebase are:
- `is_cluster_initialised`.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -345,6 +345,16 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         return "cluster_initialised" in self.app_peer_data
 
     @property
+    def is_cluster_restoring_backup(self) -> bool:
+        """Returns whether the cluster is restoring a backup."""
+        return "restoring-backup" in self.app_peer_data
+
+    @property
+    def is_cluster_restoring_to_time(self) -> bool:
+        """Returns whether the cluster is restoring a backup to a specific time."""
+        return "restore-to-time" in self.app_peer_data
+
+    @property
     def postgresql(self) -> PostgreSQL:
         """Returns an instance of the object used to interact with the database."""
         return PostgreSQL(
@@ -561,7 +571,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             return
 
         if (
-            "restoring-backup" in self.app_peer_data or "restore-to-time" in self.app_peer_data
+            self.is_cluster_restoring_backup or self.is_cluster_restoring_to_time
         ) and not self._was_restore_successful():
             logger.debug("on_peer_relation_changed early exit: Backup restore check failed")
             return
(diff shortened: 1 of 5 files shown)


src/relations/async_replication.py
```diff
@@ -524,11 +524,11 @@ def _on_async_relation_changed(self, event: RelationChangedEvent) -> None:
         if not self._stop_database(event):
             return
 
-        if not all(
+        if not (self.charm.is_unit_stopped or self._is_following_promoted_cluster()) or not all(
             "stopped" in self.charm._peers.data[unit]
             or self.charm._peers.data[unit].get("unit-promoted-cluster-counter")
             == self._get_highest_promoted_cluster_counter_value()
-            for unit in {*self.charm._peers.units, self.charm.unit}
+            for unit in self.charm._peers.units
```
sinclert-canonical, line 531: Review these lines carefully. I think I preserved the original logic

src/charm.py
```diff
@@ -344,6 +344,26 @@ def is_cluster_initialised(self) -> bool:
...
+    @property
+    def is_unit_departing(self) -> bool:
+        """Returns whether the unit is departing."""
+        return "departing" in self.unit_peer_data
+
+    @property
+    def is_unit_stopped(self) -> bool:
+        """Returns whether the unit is stopped."""
+        return "stopped" in self.unit_peer_data
+
```
dragomirp, line 366: If we are factoring those out, we might as well check if the peer data is not None.

src/relations/async_replication.py
```diff
@@ -524,11 +524,11 @@ def _on_async_relation_changed(self, event: RelationChangedEvent) -> None:
         if not self._stop_database(event):
             return
 
-        if not all(
+        if not (self.charm.is_unit_stopped or self._is_following_promoted_cluster()) or not all(
```
dragomirp, line 527:
```suggestion
        if not (self.charm.is_unit_stopped or self._is_following_promoted_cluster()) and not all(
```
I think all units have to fulfil the criteria here

sinclert-canonical: This part is hard to reason about 😕 

- I read the [original code](https://github.com/canonical/postgresql-operator/blob/cd6cc0cfc9aa88d99f044ec51cc42e9033bb8e71/src/relations/async_replication.py#L527-L532) as: enter block if any of the units is not (stopped or following).
- I read the [proposed code](https://github.com/canonical/postgresql-operator/blob/becfdf5715facfa2e06d4bb71abd4498105b66a3/src/relations/async_replication.py#L527-L532) as: enter block if current unit is not (stopped or following) `OR` any of the rest are not (stopped or following).

The uppercase `OR` is the one you are proposing to change to an `AND`, making it impossible for the code to enter the block unless current unit AND at least one of the rest are not stopped / following.

lucasgameiroborges: I had the same understanding as @sinclert-canonical here.

lucasgameiroborges (approved):

src/relations/async_replication.py
sinclert-canonical: Hey @dragomirp, did you have time to revisit this topic?

dragomirp (approved):

PR #774, opened 2025-02-17
[MISC] Define charm constants

sinclert-canonical:
This PR moves certain string into constant values, so that they could be referenced throughout the codebase.

These two in particular (i.e. the `admin` permission group and the `postgres` default database name) will be used in the upcoming LDAP support. These changes are proposed an isolation, as they have nothing to do with the LDAP support itself.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -6,6 +6,7 @@
 BACKUP_ID_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 PGBACKREST_BACKUP_ID_FORMAT = "%Y%m%d-%H%M%S"
 DATABASE = "database"
+DATABASE_DEFAULT_NAME = "postgres"
 DATABASE_PORT = "5432"
 LEGACY_DB = "db"
 LEGACY_DB_ADMIN = "db-admin"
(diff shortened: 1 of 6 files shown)


src/constants.py
```diff
@@ -6,6 +6,7 @@
 BACKUP_ID_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 PGBACKREST_BACKUP_ID_FORMAT = "%Y%m%d-%H%M%S"
 DATABASE = "database"
+DATABASE_NAME = "postgres"
```
sinclert-canonical, line 9: Please, let me know if this name is too generic.

marceloneppel (approved):

src/constants.py
dragomirp, line 9: Considering we will have multiple databases, a more specific constant name would be better.

dragomirp (approved):

src/constants.py
dragomirp: `DATABASE_DEFAULT_NAME` should be good.

sinclert-canonical: Done in commit https://github.com/canonical/postgresql-operator/pull/774/commits/db7d4c275218cbd3c5064064768f59047e99f136.

marceloneppel (approved):

PR #781, opened 2025-02-26
Fix boto3 upload to gcs

marceloneppel:


marceloneppel:
The right solution is the one from https://github.com/canonical/postgresql-k8s-operator/pull/901, as the one from this PR is deprecated.

PR #789, opened 2025-03-07
[MISC] Skip backup tests without creds

dragomirp:


tests/integration/conftest.py
dragomirp, line 1: Maybe we should also move the microceph fixtures here?

tests/integration/test_subordinates.py
```diff
@@ -21,8 +21,18 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture(scope="module")
+async def check_subordinate_env_vars(ops_test: OpsTest) -> None:
+    if (
+        not os.environ.get("UBUNTU_PRO_TOKEN", "").strip()
+        or not os.environ.get("LANDSCAPE_ACCOUNT_NAME", "").strip()
+        or not os.environ.get("LANDSCAPE_REGISTRATION_KEY", "").strip()
```
dragomirp, line 29: Secrets not accessible by forks.

taurus-forever (approved):

PR #833, opened 2025-04-15
[DPE-6189] Manage passwords with user secrets

reneradoi:
backported from `postgresql-k8s-operator`

This PR introduces juju user secrets for managing passwords. It contains the following changes:

**Functionality:**
- no longer support `get-password` and `set-password` actions
- new config option `system-users` to configure a secret that includes the system users' password(s)
- it is no longer possible to leave out the password parameter to trigger a password rotation as secrets cannot have empty values

**Implementation:**
- add handler for `secret_changed` event to `charm.py`
- add method `get_secret_from_id()` to `charm.py`
- trigger updating the system user passwords on `config_changed`
- consider pre-configured system user passwords on `leader_elected`
- remove `get_password` handler
- replace `set_password` handler with `_update_admin_passwords()` method
- `_update_admin_passwords()` is responsible for the actual business logic:
  - retrieving the passwords from the configured secret
  - checking if password updates need to be performed
  - calling the `postgresql.update_user_password()` method

**Testing:**
- remove integration tests that are no longer required, such as testing for empty password or testing if the output of get-password is the same as the output of juju show-secret
- adjust integration test helpers for get_password() and set_password to use secrets
- remove unit tests for get-/set-password actions


dragomirp:
Looks like the downgrade of LXD to LTS version didn't get applied in the merge of main and it is likely causing the instability of the arm self healing test. Started a new merge in https://github.com/canonical/postgresql-operator/pull/843 including a manual addition of that change. I think it's OK the merge the current PR as is or alternatively I can sync it up after #843 gets merged. @reneradoi, @marceloneppel what do you think?

reneradoi:
> I think it's OK the merge the current PR as is or alternatively I can sync it up after #843 gets merged. @reneradoi, @marceloneppel what do you think?

Thank you for investigating the failure @dragomirp , I'm fine with merging it as it is (after Marcelo confirmed the testing) if there's no concerns from your side.

marceloneppel (approved):

dragomirp (approved):

PR #837, opened 2025-04-16
[DPE-6752] Implement predefined roles

shayancanonical:
## Checklist

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -164,63 +164,57 @@ class PostgreSQL:
         connection.autocommit = True
         return connection
 
-    def create_database(
-        self,
-        database: str,
-        user: str,
-        plugins: Optional[List[str]] = None,
-        client_relations: Optional[List[Relation]] = None,
-    ) -> None:
+    def create_database(self, database: str,) -> bool:
         """Creates a new database and grant privileges to a user on it.
 
         Args:
-            database: database to be created.
-            user: user that will have access to the database.
-            plugins: extensions to enable in the new database.
-            client_relations: current established client relations.
+            database: database to be created
+
+        Returns:
+            boolean indicating whether a database was created
         """
-        plugins = plugins if plugins else []
-        client_relations = client_relations if client_relations else []
         try:
             connection = self._connect_to_database()
             cursor = connection.cursor()
+
             cursor.execute(
                 SQL("SELECT datname FROM pg_database WHERE datname={};").format(Literal(database))
             )
-            if cursor.fetchone() is None:
-                cursor.execute(SQL("CREATE DATABASE {};").format(Identifier(database)))
-            cursor.execute(
-                SQL("REVOKE ALL PRIVILEGES ON DATABASE {} FROM PUBLIC;").format(
-                    Identifier(database)
-                )
-            )
-            for user_to_grant_access in [user, *self.system_users]:
+
+            if cursor.fetchone() is not None:
+                return False
+
+            cursor.execute(SQL("CREATE DATABASE {};").format(Identifier(database)))
+            cursor.execute(SQL("ALTER DATABASE {} OWNER TO {}").format(Identifier(database), Identifier(f"{database}_owner")))
+            cursor.execute(SQL("CREATE ROLE {} NOSUPERUSER NOCREATEDB NOCREATEROLE NOLOGIN NOREPLICATION;").format(Identifier(f"{database}_admin")))
+            cursor.execute(SQL("REVOKE ALL PRIVILEGES ON DATABASE {} FROM PUBLIC;").format(Identifier(database)))
+
+            for user_to_grant_access in self.system_users:
                 cursor.execute(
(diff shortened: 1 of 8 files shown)


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -716,11 +868,6 @@ def set_up_database(self) -> None:
                             Identifier(user)
                         )
                     )
-                self.create_user(
-                    PERMISSIONS_GROUP_ADMIN,
-                    extra_user_roles=["pg_read_all_data", "pg_write_all_data"],
-                )
```
marceloneppel, line 722: As this seems to be a breaking change, I think we need to bump the `LIBAPI` value to 1, so we can avoid needing to backport this change to PostgreSQL 14 or adapt this library to work on both PG versions.

shayancanonical: bumped postgresql lib to v1 in [539b82f](https://github.com/canonical/postgresql-operator/pull/837/commits/539b82f31c2f61c3a824a831f3fc224ea3c4ae53)

will open a separate PR shortly to add the file in the postgresql-k8s repo. this way, the lib can be cleanly published on charmhub

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -202,129 +220,243 @@ def create_access_groups(self) -> None:
...
+    is_user_admin BOOLEAN;
+BEGIN
+    IF NOT login_hook.is_executing_login_hook()
+    THEN
+        RAISE EXCEPTION 'The login_hook.login() function should only be invoked by the login_hook code';
+    END IF;
+
+    cur_user := (SELECT current_user);
+
+    EXECUTE 'SELECT current_database()' INTO db_name;
```
marceloneppel, line 293: Out of curiosity: does this work in replicas? I had some errors in the past where I tried to use `INTO variable` in the replicas. I'm not sure if it was really related to the unit being a replica or another reason.

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -202,129 +220,243 @@ def create_access_groups(self) -> None:
...
+            createdb_enabled = True
+            roles.remove("createdb")
+
+        if "createrole" in roles:
+            createrole_enabled = True
+            roles.remove("createrole")
+
+        if "superuser" in roles:
+            logger.warning("superuser privileges not allowed via extra-user-roles")
+            roles.remove("superuser")
```
marceloneppel, line 366: Won't this break PgBouncer? I believe PgBouncer needs the `SUPERUSER` role to grant the `SUPERUSER` role to Landscape or another application when it connects to the PgBouncer legacy relation called `db-admin`. I know that it might not apply to this charm version (PG 16), but we may need to handle this on PgBouncer side, to avoid it breaking.

marceloneppel (commented):
Nice implementation, Shayan!

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: Thanks!

lib/charms/postgresql_k8s/v1/postgresql.py
dragomirp: The PGB auth user also can't be created, since it needs access to user password hashes.

lib/charms/postgresql_k8s/v1/postgresql.py
dragomirp: The PGB auth user also can't be created, since it needs access to user password hashes.

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -202,129 +220,243 @@ def create_access_groups(self) -> None:
...
+                        cursor.execute(SQL(query))
+        except psycopg2.Error as e:
+            logger.error(f"Failed to create predefined roles: {e}")
+            raise PostgreSQLCreatePredefinedRolesError() from e
+
+    def grant_database_privileges_to_user(self, user: str, database: str, privileges: list[str]) -> None:
+        """Grant the specified priviliges on the provided database for the user."""
+        try:
+            with self._connect_to_database() as connection, connection.cursor() as cursor:
+                cursor.execute(SQL("GRANT {} ON DATABASE {} TO {};").format(Identifier(", ".join(privileges)), Identifier(database), Identifier(user)))
```
dragomirp, line 456: PGB and create role users will need to be able to regrant at least connect privileges to created users.

lib/charms/postgresql_k8s/v1/postgresql.py
shayancanonical: I was wondering if in PG 16 only, we can translate `SUPERUSER` extra-user-role to `CREATEDB,CREATEROLE` in postgresql charm. if pgbouncer does not need superuser, it can continue to request superuser, and we can just provide it with what it needs in PG 16. we will also add support for `pgbouncer` extra-user-role which pgbouncer can request once we build in support for sharing PG versions through data interfaces

PR #841, opened 2025-04-16
feat: [DPE-6904] Spaces support

paulomach:
## Issue

Currently juju spaces is not supported.

## Solution

Add support for juju spaces per spec [DA-162](https://docs.google.com/document/d/1rNbiO9dF_kGD79PVzp-02PyBa2cKwcVNt-nShfJVUdU/edit?tab=t.0)

## Checklist


src/charm.py
```diff
@@ -1015,6 +1016,35 @@ def _unit_ip(self) -> str:
...
+    @property
+    def listen_ips(self) -> list[str]:
+        """Return the IPs to listen on.
+
+        This is used to configure the PostgreSQL server.
+        Peer relation IP must be first in list.
+        ref.: https://patroni.readthedocs.io/en/latest/yaml_configuration.html#postgresql
+        """
+        if self._unit_ip == self._database_ip:
+            return [self._unit_ip]
```
paulomach, line 1033: Original idea was to use a set and skip this test, but order is important on patroni file, from [docs](https://patroni.readthedocs.io/en/latest/yaml_configuration.html#postgresql):

> Patroni will use the first address from this list to establish local connections to the PostgreSQL node.

src/relations/postgresql_provider.py
```diff
@@ -73,6 +81,15 @@ def _sanitize_extra_roles(extra_roles: str | None) -> list[str]:
 
         return [role.lower() for role in extra_roles.split(",")]
 
+    def register_address_for_endpoint(self, _) -> None:
+        """Register the address for the endpoint.
+
+        If spaces are defined for endpoint, address are registered for the unit in peer relation.
+        """
+        address = str(self.model.get_binding(self.relation_name).network.bind_address)
+        self.charm.unit_peer_data[f"{self.relation_name}-address"] = address
```
paulomach, line 90: this will introduce a new field for all unit peer databag with current IP for this endpoint, e.g.:

```yaml
local-unit:
  in-scope: true
  data:
    database-address: 10.239.34.146 <--- this
    egress-subnets: 10.163.154.207/32
    exporter-started: "True"
    ingress-address: 10.163.154.207
    ...
```

The value is used to update endpoint for related apps, since patroni will not report the same addresses when using separated spaces for `peer` and `database`

src/relations/postgresql_provider.py
```diff
@@ -57,6 +62,9 @@ def __init__(self, charm: CharmBase, relation_name: str = "database") -> None:
             charm.on[self.relation_name].relation_changed,
             self._on_relation_changed_event,
         )
+        self.framework.observe(
+            charm.on[self.relation_name].relation_created, self.register_address_for_endpoint
+        )
```
marceloneppel, line 67: Probably we also need to run that method in the config changed event to update the IP address when it changes as per https://documentation.ubuntu.com/juju/3.6/reference/hook/#config-changed:

> config-changed
It also runs whenever:
when any IP address of the machine hosting the unit changes.

paulomach:
Deprecated by same PR for 16/edge branch

PR #845, opened 2025-04-20
[MISC] Disable network cut tests on arm

dragomirp:
Disable network cut tests for arm64

## Checklist


taurus-forever (approved):
Can you please add reasoning (and reference to backlog) for this disabling.
I suspect they are unstable, but what is the plan? :-D

marceloneppel (approved):

PR #853, opened 2025-04-24
Add conditional expose directive

swetha1654:
## Issue
The current terraform configuration does not include the option to expose the postgresql application. Since this module is used by [hockeypuck](https://github.com/canonical/hockeypuck-k8s-operator/blob/73052786638e36ea011e0dcead39ea583ce8558d/terraform/product/main.tf#L27) charm in the staginge environment, this is a requirement for the hockeypuck terraform files to work. The expose directive needs to be enabled to allow cross model relation between postgresql charm and the hockeypuck charm.

## Solution
Add a variable `enable_expose` that adds `expose {}` only if it is set to `True`.

## Checklist

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -16,4 +16,9 @@ resource "juju_application" "machine_postgresql" {
   units       = var.units
   constraints = var.constraints
   config      = var.config
+
+  dynamic "expose" {
+    for_each = var.enable_expose ? [1] : []
+    content {}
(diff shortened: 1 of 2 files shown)


marceloneppel (approved):
Thanks for the contribution, @swetha1654! I tested it locally and it's working fine.

taurus-forever (approved):

PR #866, opened 2025-05-01
Add in-place upgrades & rollbacks with charm-refresh

carlcsaposs-canonical:
Uses `charm-refresh` Python package: https://github.com/canonical/charm-refresh

Only supports refreshing to/from postgresql charm code with refresh v3

### DIFF
--- a/tests/integration/ha_tests/test_smoke.py
+++ b/tests/integration/ha_tests/test_smoke.py
@@ -143,6 +143,7 @@ async def test_charm_garbage_ignorance(ops_test: OpsTest, charm: str):
 
 
 @pytest.mark.abort_on_fail
+@pytest.mark.skip(reason="Unstable")
 async def test_app_resources_conflicts_v3(ops_test: OpsTest, charm: str):
     """Test application deploy in dirty environment with garbage storage from another application."""
     async with ops_test.fast_forward():


src/relations/async_replication.py
```diff
@@ -691,8 +691,14 @@ def _relation(self) -> typing.Optional[Relation]:
...
                 return relation
 
-    def _set_app_status(self) -> None:
+    def set_app_status(self) -> None:
         """Set the app status."""
+        if self.charm.refresh is not None and self.charm.refresh.app_status_higher_priority:
+            self.charm.app.status = self.charm.refresh.app_status_higher_priority
+            return
+        if self.charm._peers is None:
+            # TODO set active status?
```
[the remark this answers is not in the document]
marceloneppel: I remember that Mobhamed wants an app status set during async replication (as we're currently doing). However, I remember that the Juju team commented that once we set an app status, the app status won't be set anymore based on units statuses. Do you know if this is reset after the charm is refreshed? (I mean, does the app status start to be calculated again based on units statuses after a refresh, where before refreshing we manually set the app status).

refresh_versions.toml
```diff
@@ -0,0 +1,13 @@
...
+charm = "16/1.3.0" # TODO remove
+
+[snap]
+name = "charmed-postgresql"
+
+[snap.revisions]
+# amd64
+x86_64 = "170"
+# arm64
+aarch64 = "169"
```
dragomirp, line 13: Moving the snap revs will conflict with https://github.com/canonical/postgresql-operator/pull/881. Merging will need to be synchronised.

shayancanonical: ack, i will update once rebasing after 881 is merged

dragomirp (approved):

taurus-forever:
...
> @taurus-forever please confirm this looks good to you

Confirmed.

PR #881, opened 2025-05-06
[DPE-6752] Implement instance level predefined roles

shayancanonical:
## Issue
Contains part of the implementation in https://github.com/canonical/postgresql-operator/pull/837

We would like to break out the implementation of predefined roles, to be able to merge instance level predefined roles. We will implement the catalog level predefined roles once we complete some polish.

## Solution
Port over instance level predefined roles from PR 837

## Checklist


lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -257,74 +272,126 @@ def create_database(
...
+            roles.remove("createdb")
+
+        if "createrole" in roles:
+            createrole_enabled = True
+            roles.remove("createrole")
+
+        if "superuser" in roles:
+            superuser_enabled = True
+            roles.remove("superuser")
+
```
taurus-forever, line 308: Are createdb/createrole necessary to be changed in this PR?

shayancanonical: i believe that the implementation is different but functionality is actually identical. one can specify `extra-user-roles=createdb,createrole,admin,superuser` as they could before. we just process the requested roles using conditions like this instead of creating a whitelist of accepted attributes/privileges

lib/charms/postgresql_k8s/v1/postgresql.py
dragomirp, line 1: Pending changes will need to be merged, if we are moving the library, everyone else is currently working against v0.

dragomirp:
Check libs is likely failing because v1 doesn't exist in the remote charm (PG k8s).

dragomirp (approved):

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -31,17 +31,24 @@
 LIBID = "24ee217a54e840a598ff21a079c3e678"
 
 # Increment this major API version when introducing breaking changes
-LIBAPI = 0
+LIBAPI = 1
```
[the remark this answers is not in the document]
marceloneppel: Initially, I believed it was the best way to separate this new feature for PG 16. But we can revisit it to avoid conditional library version use on PgBouncer later (if we find issues in that charm due to this change). I don't see a problem with doing this now.

src/constants.py
```diff
@@ -24,15 +24,15 @@
...
+SYSTEM_USERS = [REPLICATION_USER, REWIND_USER, USER, MONITORING_USER]
 
 # Snap constants.
 PGBACKREST_EXECUTABLE = "charmed-postgresql.pgbackrest"
 POSTGRESQL_SNAP_NAME = "charmed-postgresql"
 SNAP_PACKAGES = [
     (
         POSTGRESQL_SNAP_NAME,
-        {"revision": {"aarch64": "169", "x86_64": "170"}},
+        {"revision": {"aarch64": "181", "x86_64": "182"}},
```
dragomirp, line 35: https://github.com/canonical/postgresql-operator/pull/866 is moving the snap revs definitions.

tests/integration/test_predefined_roles.py
```diff
@@ -0,0 +1,220 @@
...
+    get_juju_secret,
+)
+
+logger = logging.getLogger(__name__)
+
+TIMEOUT = 15 * 60
+DATA_INTEGRATOR_APP_NAME = "data-integrator"
+
+# NOTE: We are unable to test set_user_u('operator') as dba user because psycopg2
+# runs every query in a transaction and running set_user() is not supported in transactions.
```
marceloneppel, line 33: I faced a similar error when trying to create a database (it cannot be done inside a transaction block). For that case, I solved by connecting to the database without a context manager (I mean, without the `with` keyword in front of either `psycopg2.connect()` or `db_connect()`, like it's done at: https://github.com/canonical/postgresql-k8s-operator/blob/fcfa631e80c4880f1e05c0a09e4f513caa1ed3e5/tests/integration/new_relations/test_new_relations_1.py#L171-L180).

marceloneppel (approved):

taurus-forever (approved):
Tests for charmhub branch `16/edge/predefined-roles` had no issues and code is OK as a phase 1. Thankl you @shayancanonical !

PR #900, opened 2025-05-09
[DPE-7129] Fix network cut without ip change test

marceloneppel:
## Issue
The `network cut without IP change` test is failing consistently and preventing the publishing of the charm.

## Solution
Move the `_update_member_ip` call to correctly remove the Raft cluster member when the network is cut.

The' config-changed' event is the correct event that tells the charm if a unit IP has changed.

## Checklist


dragomirp (approved):

taurus-forever (approved):
This is a nice catch! Thank you!

PR #909, opened 2025-05-20
[DPE-7309][DPE-7326] Switch to TLSv4

dragomirp:
* Add support for TLSv4 charm lib
* Split the certificate relation into peer and client
* Add internal CA to be used if no peer rel is available
* Updated charm libs

## Checklist


src/relations/tls.py
```diff
@@ -0,0 +1,175 @@
...
+        for rel in TLS_RELS:
+            self.framework.observe(self.charm.on[rel].relation_created, self._on_relation_created)
+            self.framework.observe(
+                self.charm.on[rel].relation_broken, self._on_certificates_broken
+            )
+
+    def _on_relation_created(self, event: RelationCreatedEvent) -> None:
+        if not self.model.get_relation(TLS_CLIENT_RELATION) or self.model.get_relation(
+            TLS_PEER_RELATION
+        ):
```
marceloneppel, line 119: Should a `not` be part of the second condition?

templates/patroni.yml.j2
```diff
@@ -188,25 +182,25 @@ postgresql:
...
     superuser:
       username: {{ superuser }}
       password: {{ superuser_password }}
       {%- if enable_tls %}
       sslrootcert: {{ conf_path }}/ca.pem
-      sslcert: {{ conf_path }}/cert.pem
-      sslkey: {{ conf_path }}/key.pem
       {%- endif %}
+      sslcert: {{ conf_path }}/patroni_cert.pem
+      sslkey: {{ conf_path }}/patroni_key.pem
```
dragomirp, line 203: Early during initialisation Patroni fails to connect to the database with:
```
2025-05-30 00:27:48 UTC [21649]: ERROR: get_postgresql_status 
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 1401, in query
    heartbeat_connection.get()  # try to open psycopg connection to postgres
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/patroni/postgresql/connection.py", line 54, in get
    self._connection = psycopg.connect(**self._conn_kwargs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/patroni/psycopg.py", line 103, in connect
    ret = _connect(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 122, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.OperationalError: connection to server at "10.108.175.139", port 5432 failed: could not open certificate file "/root/snap/charmed-postgresql/170/.postgresql/postgresql.crt": Permission denied
```
and pg_basebackup fails on replica likely for similar reasons. Pointing the client certs to nonexistent location to be ignored during connection.

src/cluster.py
```diff
@@ -164,11 +162,10 @@ def __init__(
         self.rewind_password = rewind_password
         self.raft_password = raft_password
         self.patroni_password = patroni_password
-        self.tls_enabled = tls_enabled
         # Variable mapping to requests library verify parameter.
         # The CA bundle file is used to validate the server certificate when
         # TLS is enabled, otherwise True is set because it's the default value.
-        self.verify = f"{PATRONI_CONF_PATH}/{TLS_CA_FILE}" if tls_enabled else True
+        self.verify = f"{PATRONI_CONF_PATH}/{TLS_CA_BUNDLE_FILE}"
```
dragomirp, line 168: When switching between real peer certificate from operator and the internally generated one, the charm can get stuck validating against the wrong CA, so both are bundled inside this file to validate against in requests.

src/backups.py
dragomirp, line 1: We always have peer tls now, so we should just check based on number of units.

src/charm.py
```diff
@@ -1571,6 +1576,13 @@ def _on_start(self, event: StartEvent) -> None:
             event.defer()
             return
 
+        if not self.get_secret(APP_SCOPE, "internal-ca"):
+            logger.info("leader not elected and/or internal CA not yet generated")
+            event.defer()
+            return
+        if not self.get_secret(UNIT_SCOPE, "internal-cert"):
+            self.tls.generate_internal_peer_cert()
+
```
dragomirp, line 1585: If the leader has generated the internal CA, the followers should generate their own certs.

src/charm.py
```diff
@@ -2059,8 +2071,10 @@ def _update_certificate(self) -> None:
         # Request the certificate only if there is already one. If there isn't,
         # the certificate will be generated in the relation joined event when
         # relating to the TLS Certificates Operator.
-        if all(self.tls.get_tls_files()):
-            self.tls._request_certificate(self.get_secret("unit", "private-key"))
+        if all(self.tls.get_client_tls_files()) or all(self.tls.get_peer_tls_files()):
+            self.tls.refresh_tls_certificates_event.emit()
+        if self.get_secret(UNIT_SCOPE, "internal-cert"):
+            self.tls.generate_internal_peer_cert()
```
dragomirp, line 2077: Regenerate the internal cert in case of IP change.

src/cluster.py
```diff
@@ -256,6 +253,8 @@ def cluster_status(
     ) -> Optional[list[ClusterMember]]:
         """Query the cluster status."""
         # Request info from cluster endpoint (which returns all members of the cluster).
+        # TODO we don't know the other cluster's ca
+        verify = self.verify if not alternative_endpoints else False
```
dragomirp, line 257: With async replication (when we add the alternative endpoints of the other cluster), we don't know the other side's CA, so can't really verify the cert at the moment. We should add the CAs to the async relations so that we can verify the certs as needed.

src/relations/tls_transfer.py
dragomirp, line 1: Moved out of the internal tls lib.

dragomirp, line 1: Moved out of the internal tls lib.

tests/integration/helpers.py
```diff
@@ -858,8 +855,8 @@ async def check_tls_patroni_api(ops_test: OpsTest, unit_name: str, enabled: bool
                 # TLS is enabled, otherwise True is set because it's the default value
                 # for the verify parameter.
```
delgod, line 856: most probably comment should be adjusted

src/charm.py
delgod: TBD

src/cluster.py
taurus-forever: LGTM, but can't we use "internal_CA if self.tls_enabled else relation_CA"?

taurus-forever (approved):
LGTM in general, please check it with Marcelo and address Mykola comments. Tnx!

src/cluster.py
dragomirp: I don't think so. As is the charm property will be true if all the relation components are available (key, cert and ca), not if those are currently used by Patroni. To be able to use only one CA, we have to track which cert is being used by Patroni ATM on each unit.

src/charm.py
```diff
@@ -837,7 +841,6 @@ def _peer_relation_changed_checks(self, event: HookEvent) -> bool:
         # Prevents the cluster to be reconfigured before it's bootstrapped in the leader.
         if not self.is_cluster_initialised:
             logger.debug("Deferring on_peer_relation_changed: cluster not initialized")
-            event.defer()
```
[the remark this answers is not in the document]
dragomirp: Updated the logging message in https://github.com/canonical/postgresql-operator/pull/909/commits/e1a82de6937ed98ec404515df5cc62f047f76296

tests/integration/helpers.py
dragomirp: Updated to reflect current behaviour in https://github.com/canonical/postgresql-operator/pull/909/commits/0c73c024046826930a714810dce1c60de74c7ad1

delgod (approved):

PR #910, opened 2025-05-21
Fix refresh tests to not rely on existing charm revisions

shayancanonical:
## Issue
Refresh tests currently rely on a specific charm revision to exist - tests fail on 16/edge as the charm version and workload version do not change when upgrading from 16/edge to locally built charm, leading to no-op in refresh.

## Solution
Add helper to update charm version and workload version (bump minor version) to ensure that 16/edge -> local charm or 16/stable -> local charm will always pass the refresh condition checks.

## Checklist

### DIFF
--- a/tests/integration/ha_tests/test_upgrade.py
+++ b/tests/integration/ha_tests/test_upgrade.py
@@ -2,7 +2,6 @@
 # See LICENSE file for licensing details.
 
 import logging
-import pathlib
 import platform
 import shutil
 import zipfile
@@ -40,7 +39,7 @@ async def test_deploy_latest(ops_test: OpsTest) -> None:
         "-n",
         3,
         "--channel",
-        "16/edge/test-refresh-v3-workload2",  # TODO remove branch
+        "16/edge",
         "--config",
         "profile=testing",
         "--base",
@@ -86,8 +85,14 @@ async def test_upgrade_from_edge(ops_test: OpsTest, continuous_writes, charm) ->
 
     application = ops_test.model.applications[DATABASE_APP_NAME]
 
+    logger.info("Update workload minor version in the charm")
+    filename = Path(charm).name
+    temporary_charm = Path("/tmp", f"{filename}.temporary.charm")
+    shutil.copy(charm, temporary_charm)
+    create_valid_upgrade_charm(temporary_charm)
+
     logger.info("Refresh the charm")
-    await application.refresh(path=charm)
+    await application.refresh(path=temporary_charm)
 
     logger.info("Wait for upgrade to start")
     await ops_test.model.block_until(lambda: application.status == "blocked", timeout=TIMEOUT)
@@ -127,6 +132,8 @@ async def test_upgrade_from_edge(ops_test: OpsTest, continuous_writes, charm) ->
         "Number of switchovers is greater than 2"
     )
 
+    temporary_charm.unlink()
+
 
 @pytest.mark.abort_on_fail
 async def test_fail_and_rollback(ops_test, charm, continuous_writes) -> None:
@@ -147,7 +154,7 @@ async def test_fail_and_rollback(ops_test, charm, continuous_writes) -> None:
     await action.wait()
 
     filename = Path(charm).name
-    fault_charm = Path("/tmp/", filename)
+    fault_charm = Path("/tmp", f"{filename}.fault.charm")
     shutil.copy(charm, fault_charm)
 
     logger.info("Inject dependency fault")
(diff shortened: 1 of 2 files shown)


tests/integration/ha_tests/test_upgrade.py
```diff
@@ -203,3 +210,29 @@ async def inject_dependency_fault(charm_file: str | Path) -> None:
...
+    with (
+        zipfile.ZipFile(charm_file, mode="r") as charm_zip,
+        charm_zip.open("refresh_versions.toml") as file,
+    ):
+        versions = tomli.load(file)
+
+    packed_charm_version = versions["charm"]
+    charm_version_components = packed_charm_version.split(".post")
+    major, minor, patch = charm_version_components[0].split(".")
+    versions["charm"] = f"{major}.{int(minor) + 1}.{patch}"
```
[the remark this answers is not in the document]
shayancanonical: agreed, but this charm is only being used in tests - it is not even getting uploaded or packed (but rather getting modified in CI to test the upgrade). are there any other safer alternatives?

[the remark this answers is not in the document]
shayancanonical: which unit must one run `force-refresh-start check-compatibility=false` and when must this action be run?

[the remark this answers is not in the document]
shayancanonical: Addressed in [e566504](https://github.com/canonical/postgresql-operator/pull/910/commits/e5665043e99b55dfc318c160f6e771315219c6b0)

tests/integration/ha_tests/test_upgrade_from_stable.py
```diff
@@ -84,17 +84,32 @@ async def test_upgrade_from_stable(ops_test: OpsTest, charm):
...
     async with ops_test.fast_forward("60s"):
         await ops_test.model.wait_for_idle(
             apps=[DATABASE_APP_NAME], idle_period=30, timeout=TIMEOUT
         )
 
-    logger.info("Run resume-refresh action")
+    assert "Refresh incompatible" in application.status_message, (
+        "Application refresh not blocked due to incompatibility"
+    )
+
```
marceloneppel, line 96: Will the refresh be compatible at some point (with, for example, changes in the workload) and fail at this assertion?

shayancanonical: I believe the refresh will always be incompatible because we are using a locally packed (or CI packed) charm as the refresh target. This charm will have a charm version that contains `.post.dev...` in addition to the semantic versioning of the charm (e.g. `16/1.9.0.post.dev+dirty-somehash`)

marceloneppel (approved):
LGTM as soon as the upgrade tests pass.

taurus-forever (approved):

tests/integration/ha_tests/test_upgrade_from_stable.py
marceloneppel: I was double-checking the still failing upgrade tests, and it seems that if we deploy the old Charmhub branch (`16/edge/test-refresh-v3-workload2`) your PR code works (I suspect it's because there is a line containing `charm = "16/1.8.1" # TODO remove` inside that revision's `refresh_versions.toml` file).

I'm wondering if we might need to use that Charmhub branch in this PR and later, Dragomir or I (in a next PR), rename those test files from `test_upgrade.py` and `test_upgrade_from_stable.py` to `test_incompatible_upgrade.py` and `test_incompatible_upgrade_from_stable.py` (keeping the Charmhub branch or maybe using the first approach Shayan did in this PR: changing the charm to be incompatible), while also creating two new test files `test_upgrade.py` and `test_upgrade_from_stable.py` where we use the correct branches (16/edge and 16/stable), with some modification in the test code to not run `force-refresh-start check-compatibility=false`, because in that case there won't be an incompatibility.

What do you think, @carlcsaposs-canonical and @shayancanonical? Does it make sense what I suspect is the problem here?

marceloneppel: Thank you for the details and for conducting the investigation, @carlcsaposs-canonical. Is there a way for us to have a different snap revision for testing here (or if we need to take a different approach in those tests)? If anyone has any idea, feel free to share. I'm also thinking about that.

marceloneppel: Maybe we can handle both situations in the tests, as, in the real world, there will be situations where the snap changes and others where it won't change. So we will do the relevant checks in the tests based on whether the refresh happened or not (I mean, the snap was refreshed or not). IMHO, that will be better to finish this fix.

marceloneppel: I synced with Alex, and I'll create a separate PR to make the test pass now (with the same snap revision) and create follow-up tickets for the other situations.

PR #912, opened 2025-05-23
[MISC] Delete runner password

dragomirp:
Currently spread conf expects to be able to ssh into the runner user without a password. New images now have a password set, so it's removed to reenable CI.

## Checklist

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -74,6 +74,11 @@ backends:
     # each spread job & select a single job when running spread)
     # Derived from https://github.com/jnsgruk/zinc-k8s-operator/blob/a21eae8399eb3b9df4ddb934b837af25ef831976/spread.yaml#L47
     allocate: |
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -79,6 +79,8 @@ backends:
       PermitEmptyPasswords yes
       EOF
 
+      sudo passwd -d runner
+
```
dragomirp, line 83: Will try to investigate separately if we can use a real password and not rely on passwordless ssh.

marceloneppel (approved):

spread.yaml
taurus-forever: ACK to unblock CI while further investigation is following...

taurus-forever (approved):

dragomirp:
Merging without waiting for the tests, since this is needed by other stabilisation PRs.

PR #946, opened 2025-06-09
[DPE-7521] Fix HBA rules for Landscape related through PgBouncer

marceloneppel:
## Issue
Currently, the `landscape` database user created by the Landscape charm cannot access the database when it’s related through PgBouncer due to a missing HBA rule for that user.

Example error: https://github.com/canonical/pgbouncer-operator/actions/runs/15533582087/job/43728069621#step:7:1612

## Solution
There is already a workaround to put the `landscape` database user in the HBA file when the Landscape charm is related directly to the database. Now, we added another to check when PgBouncer is related to the database.

Additionally, the list of users and databases was sorted to avoid rendering them in a different order each time the `pg_hba.conf` file is rendered and to make testing it easier.

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -2178,7 +2178,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
                 or (
                     relation.name == "database"
                     and relation.data[relation.app].get("extra-user-roles") == "SUPERUSER"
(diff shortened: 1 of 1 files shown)


dragomirp (approved):

dragomirp (approved):

taurus-forever (approved):
:+1:

PR #958, opened 2025-06-10
[DPE-7558] Fixes for predefined roles (backporting fixes from PgBouncer to PostgreSQL)

marceloneppel:
## Issue
While working on making PgBouncer compatible with PG 16 through https://github.com/canonical/pgbouncer-operator/pull/549, it was noted that some logic from the PostgreSQL charm library doesn't work as expected when used by PgBouncer (and also when the client app requests a database with the name containing a hyphen).

## Solution
* Reset PgBouncer role when needed (to have superuser privileges and be able to perform some restricted operations).
* Handle database names containing a hyphen (like `test-database`).

## Checklist

### DIFF
--- a/lib/charms/postgresql_k8s/v1/postgresql.py
+++ b/lib/charms/postgresql_k8s/v1/postgresql.py
@@ -307,7 +307,7 @@ class PostgreSQL:
                     f"WITH LOGIN{' SUPERUSER' if admin else ''} ENCRYPTED PASSWORD '{password}'"
                 )
                 if in_role:
-                    user_definition += f" IN ROLE {in_role}"
+                    user_definition += f" IN ROLE \"{in_role}\""
                 if can_create_database:
                     user_definition += " CREATEDB"
                 if privileges:


lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -1068,6 +1071,7 @@ def update_user_password(
             with self._connect_to_database(
                 database_host=database_host
             ) as connection, connection.cursor() as cursor:
+                cursor.execute(SQL("RESET ROLE;"))
```
marceloneppel, line 1074: PgBouncer needs to reset its user from `pgbouncer_owner` to `relation-X` (which is a superuser) to call `SET LOCAL log_statement = 'none';`.

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -311,6 +312,7 @@ def create_user(
                     user_definition += " CREATEDB"
                 if privileges:
                     user_definition += f" {' '.join(privileges)}"
+                cursor.execute(SQL("RESET ROLE;"))
```
marceloneppel, line 315: PgBouncer needs to reset its user from `pgbouncer_owner` to `relation-X` (which is a superuser) to call `SET LOCAL log_statement = 'none';`.

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -164,6 +164,7 @@ def _configure_pgaudit(self, enable: bool) -> None:
             connection = self._connect_to_database()
             connection.autocommit = True
             with connection.cursor() as cursor:
+                cursor.execute("RESET ROLE;")
```
marceloneppel, line 167: PgBouncer needs to reset its user from `pgbouncer_owner` to `relation-X` (which is a superuser) to call `ALTER SYSTEM...`.

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -976,8 +978,9 @@ def set_up_predefined_catalog_roles_function(self) -> None:
 BEGIN
 	database := (SELECT current_database());
 	current_session_user := (SELECT session_user);
-    owner_user := database || '_owner';
-    admin_user := database || '_admin';
+    owner_user := quote_ident(database || '_owner');
+    admin_user := quote_ident(database || '_admin');
+    database := quote_ident(database);
```
marceloneppel, line 983: When a database name contains a hyphen (like in `test-database`), it's needed to quote those identifiers, otherwise the SQL statements using them will fail.

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -242,11 +243,11 @@ def create_database(
...
                 cursor.execute(SQL("SET ROLE charmed_databases_owner;"))
                 cursor.execute(SQL("CREATE DATABASE {};").format(Identifier(database)))
-            cursor.execute(
-                SQL("REVOKE ALL PRIVILEGES ON DATABASE {} FROM PUBLIC;").format(
-                    Identifier(database)
+                cursor.execute(
+                    SQL("REVOKE ALL PRIVILEGES ON DATABASE {} FROM PUBLIC;").format(
+                        Identifier(database)
+                    )
                 )
```
marceloneppel, line 250: Execute this `REVOKE` operation only when the database is created, because it's not needed later, as it's already done and would cause an error when a relation requesting the same database name is established.

dragomirp (approved):

PR #960, opened 2025-06-11
Predefined roles compatibility

marceloneppel:
## Issue
A `GRANT` SQL command allows a user to access a database. However, the `pg_hba.conf` file rules are not updated on time to reflect that.

We encounter failures related to this when attempting to sync the latest changes from the `main` branch (which contains the `pg_hba.conf` hardening implementation) to the `16/edge` branch (which includes the predefined roles implementation) through https://github.com/canonical/postgresql-operator/pull/920.

## Solution
Add a new check for ACL changes in the cluster topology observer to update the `pg_hba.conf` file (in the worst case) after 30 seconds.

## Checklist


dragomirp (approved):

dragomirp:
We can most probably change the base branch and merge to `16/edge` directly, since CI is green and it should have all the changes necessary.

taurus-forever (approved):

PR #973, opened 2025-06-16
[MISC] Check is_user_in_hba output

dragomirp:
Bugfix for the pg_hba rules checker

## Checklist

### DIFF
--- a/lib/charms/postgresql_k8s/v0/postgresql.py
+++ b/lib/charms/postgresql_k8s/v0/postgresql.py
@@ -35,7 +35,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 56
+LIBPATCH = 53
 
 # Groups to distinguish HBA access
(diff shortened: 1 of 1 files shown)


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -777,9 +777,10 @@ def list_valid_privileges_and_roles(self) -> Tuple[Set[str], Set[str]]:
                 "superuser",
             }, {role[0] for role in cursor.fetchall() if role[0]}
 
-    def set_up_database(self) -> None:
+    def set_up_database(self, temp_location: Optional[str] = None) -> None:
```
dragomirp, line 780: Divergence from k8s 16/edge, but it doesn't seem necessary for 14/edge charms and we should be using the v1 lib for k8s 16/edge eventually. Should I keep the new param?

marceloneppel: We can drop this new parameter.

marceloneppel (approved):

marceloneppel (approved):

taurus-forever (approved):

PR #988, opened 2025-06-18
[MISC] Move pg wal

dragomirp:
Move pg_wal instead of removing

## Checklist

### DIFF
--- a/lib/charms/postgresql_k8s/v1/postgresql.py
+++ b/lib/charms/postgresql_k8s/v1/postgresql.py
@@ -386,7 +386,7 @@ class PostgreSQL:
                 connection.close()
 
     def grant_database_privileges_to_user(
-        self, user: str, database: str, privileges: list[str]
+        self, user: str, database: str, privileges: List[str]
     ) -> None:
         """Grant the specified privileges on the provided database for the user."""
         try:


src/charm.py
```diff
@@ -2035,8 +2035,13 @@ def _handle_processes_failures(self) -> bool:
         # process is running by the PostgreSQL process not).
         if self._unit_ip in self.members_ips and self._patroni.member_inactive:
             data_directory_contents = os.listdir(POSTGRESQL_DATA_PATH)
-            if len(data_directory_contents) == 1 and data_directory_contents[0] == "pg_wal":
-                os.remove(os.path.join(POSTGRESQL_DATA_PATH, "pg_wal"))
+            if len(data_directory_contents) > 0 and all(
+                content.startswith("pg_wal") for content in data_directory_contents
+            ):
```
dragomirp, line 2040: This will retry the move, if all we have in the directory is pg_wal and previous backups. Should we keep trying (`len(data_directory_contents) > 0`) or try just once`len(data_directory_contents) == 1`?

src/charm.py
```diff
@@ -2035,8 +2035,13 @@ def _handle_processes_failures(self) -> bool:
...
+                os.rename(
+                    os.path.join(POSTGRESQL_DATA_PATH, "pg_wal"),
+                    os.path.join(POSTGRESQL_DATA_PATH, f"pg_wal-{datetime.now(UTC).isoformat()}"),
```
dragomirp, line 2043: Add timestamp in case we have multiple moves.

src/charm.py
marceloneppel: I think only once is enough, because the backup might go away in a re-relation.

marceloneppel (approved):

src/charm.py
taurus-forever: Let's try with once.
Also I would like to see backlog ticket trying to trace this and remove this code/workaround completely.

taurus-forever (approved):

src/charm.py
dragomirp: Restored the original condition to try only once.

dragomirp: > Also I would like to see backlog ticket trying to trace this and remove this code/workaround completely.

https://github.com/canonical/postgresql-operator/issues/989

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -386,7 +386,7 @@ def create_predefined_instance_roles(self) -> None:
                 connection.close()
 
     def grant_database_privileges_to_user(
-        self, user: str, database: str, privileges: list[str]
+        self, user: str, database: str, privileges: List[str]
```
dragomirp, line 389: Copied from PGB compatibility PR.

PR #991, opened 2025-06-20
Create users and roles on restoration

delgod:
## Issue
Issues arise during restoration if the backup doesn't contain users and roles that are not present on the target system.

## Solution
Ensure any missing roles are created after the restoration process is complete.


src/charm.py
```diff
@@ -1705,7 +1705,46 @@ def _setup_ldap_sync(self, postgres_snap: snap.Snap | None = None) -> None:
...
+
+        access_groups = self.postgresql.list_access_groups()
+        if access_groups != set(ACCESS_GROUPS):
+            self.postgresql.create_access_groups()
+            self.postgresql.grant_internal_access_group_memberships()
+
+        access_groups = self.postgresql.list_access_groups()
+        if access_groups != set(ACCESS_GROUPS):
+            self.postgresql.create_access_groups()
+            self.postgresql.grant_internal_access_group_memberships()
```
marceloneppel, line 1743: This block can be removed (it was already duplicated in the charm code before this PR).

src/charm.py
```diff
@@ -1705,7 +1705,46 @@ def _setup_ldap_sync(self, postgres_snap: snap.Snap | None = None) -> None:
...
+                BACKUP_USER, new_password(), extra_user_roles=[ROLE_BACKUP]
+            )
+            self.postgresql.grant_database_privileges_to_user(BACKUP_USER, "postgres", ["connect"])
+        if MONITORING_USER not in users:
+            # Create the monitoring user.
+            self.postgresql.create_user(
+                MONITORING_USER,
+                self.get_secret(APP_SCOPE, MONITORING_PASSWORD_KEY),
+                extra_user_roles=[ROLE_STATS],
+            )
```
marceloneppel, line 1729: Those two user creations can be removed as they are created inside the call to `self.postgresql.create_predefined_instance_roles()` (they were already forgotten in the charm code before this PR).

marceloneppel (approved):
LGTM!

I only added comments related to leftovers from our previous PRs that I identified as a result of this PR.

src/charm.py
marceloneppel: Apologies, I was wrong. The roles ("groups") were created, not the users. Ignore this comment.

dragomirp (approved):

taurus-forever (approved):
Thank you @delgod !

PR #994, opened 2025-06-23
[DPE-7614] restart during starting

dragomirp:
## Checklist


src/cluster.py
```diff
@@ -54,7 +54,8 @@
 
 PG_BASE_CONF_PATH = f"{POSTGRESQL_CONF_PATH}/postgresql.conf"
 
-RUNNING_STATES = ["running", "streaming"]
+STARTED_STATES = ["running", "streaming"]
```
dragomirp, line 57: Everything except the check if the unit is running should be using the old states in the new list.

marceloneppel (approved):

PR #999, opened 2025-06-24
[DPE-7614] Remove replica reinitialisations

dragomirp:
Remove replica reinitialisations to allow them to sync up in case of a large database

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -913,23 +913,6 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             event.defer()
             return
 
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -2054,40 +2038,6 @@ def _handle_processes_failures(self) -> bool:
 
         return False
 
-    def _handle_workload_failures(self) -> bool:
```
dragomirp, line 2057: Only one if check left, so moved in the other method.

taurus-forever (approved):

PR #1014, opened 2025-07-01
[MISC] Converge and cache cluster call logic

dragomirp:
## Checklist

### DIFF
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -224,17 +224,10 @@ class Patroni:
         os.chown(path, uid=user_database.pw_uid, gid=user_database.pw_gid)
 
     @property
-    @retry(stop=stop_after_attempt(10), wait=wait_exponential(multiplier=1, min=2, max=10))
     def cluster_members(self) -> set:
         """Get the current cluster members."""
         # Request info from cluster endpoint (which returns all members of the cluster).
-        cluster_status = requests.get(
-            f"{self._patroni_url}/{PATRONI_CLUSTER_STATUS_ENDPOINT}",
-            verify=self.verify,
-            timeout=API_REQUEST_TIMEOUT,
-            auth=self._patroni_auth,
-        )
-        return {member["name"] for member in cluster_status.json()["members"]}
+        return {member["name"] for member in self.cached_cluster_status}
 
     def _create_directory(self, path: str, mode: int) -> None:
         """Creates a directory.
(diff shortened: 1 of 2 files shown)


src/cluster.py
```diff
@@ -535,27 +520,6 @@ def member_inactive(self) -> bool:
             "restarting",
         ]
 
-    @property
-    def member_replication_lag(self) -> str:
```
dragomirp, line 539: No longer used.

src/cluster_topology_observer.py
```diff
@@ -121,9 +121,3 @@ def stop_observer(self):
             self._charm._peers.data[self._charm.unit].update({"observer-pid": ""})
         except OSError:
             pass
-
-    @property
-    def unit_tag(self):
-        """Juju-style tag identifying the unit being run by this charm."""
-        unit_num = self._charm.unit.name.split("/")[-1]
-        return f"unit-{self._charm.app.name}-{unit_num}"
```
dragomirp, line 129: Not called anywhere.

src/cluster.py
```diff
@@ -509,7 +494,7 @@ def member_started(self) -> bool:
         if not self.is_patroni_running():
             return False
         try:
-            response = self.get_patroni_health()
+            response = self.cached_patroni_health
```
dragomirp, line 497: Property, so using cached call.

src/cluster.py
```diff
@@ -535,27 +520,6 @@ def member_inactive(self) -> bool:
...
-            return "unknown"
-
-        for member in cluster_status.json()["members"]:
-            if member["name"] == self.member_name:
-                return member.get("lag", "unknown")
-
-        return "unknown"
-
     @property
     def is_member_isolated(self) -> bool:
```
dragomirp, line 524: There's another call to cluster status here, but need to verify how the response will look on an isolated unit.

src/cluster.py
dragomirp, line 1: Tried to use the cached cluster call in `get_primary()` but it seems to be causing an instability in the process freeze tests, so leaving be for now.

taurus-forever (approved):
LGTM!

src/cluster.py
```diff
@@ -411,27 +409,22 @@ def are_all_members_ready(self) -> bool:
...
         # Check if all members are running and one of them is a leader (primary) or
         # a standby leader, because sometimes there may exist (for some period of time)
         # only replicas after a failed switchover.
-        return all(
-            member["state"] in STARTED_STATES for member in cluster_status.json()["members"]
-        ) and any(
-            member["role"] in ["leader", "standby_leader"]
-            for member in cluster_status.json()["members"]
+        return all(member["state"] in STARTED_STATES for member in members) and any(
+            member["role"] in ["leader", "standby_leader"] for member in members
```
marceloneppel, line 420: Great addition!

marceloneppel (approved):
LGTM!

PR #1044, opened 2025-07-15
[MISC] Reduce the amount of peer data reloads

dragomirp:
## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -853,6 +853,9 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
     def _peer_relation_changed_checks(self, event: HookEvent) -> bool:
         """Split of to reduce complexity."""
         # Prevents the cluster to be reconfigured before it's bootstrapped in the leader.
+        if event.unit and event.unit != self.unit:
+            return False
+
         if not self.is_cluster_initialised:
             logger.debug("Early exit on_peer_relation_changed: cluster not initialized")
             return False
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -853,7 +853,7 @@ def has_raft_keys(self):
     def _peer_relation_changed_checks(self, event: HookEvent) -> bool:
         """Split of to reduce complexity."""
         # Prevents the cluster to be reconfigured before it's bootstrapped in the leader.
-        if event.unit and event.unit != self.unit:
+        if (hasattr(event, "unit") and event.unit) and event.unit != self.unit:
```
marceloneppel, line 856: I'm not 100% sure, but it seems that the unit is empty when it's the one that triggered the app relation change event (at https://github.com/juju/juju/blob/a48434d3928a2246efaa85536eb1225d4a7d8e22/internal/worker/uniter/relation/resolver.go#L303).

taurus-forever: Marcelo you are correct:
```
unit-postgresql-2: 18:24:41 DEBUG unit.postgresql/2.juju-log database-peers:6: __main__:Calling on_peer_relation_changed, unit: 'None'
unit-postgresql-2: 18:24:41 DEBUG unit.postgresql/2.juju-log database-peers:6: __main__:Calling on_peer_relation_changed, event: '<RelationChangedEvent app='postgresql' unit=None on <ops.model.Relation database-peers:6> via PostgresqlOperatorCharm/on/database_peers_relation_changed[66]>'
```

Hovewer, there are also events with unit=self.... we need to explore this direction further.

dragomirp:
Doesn't seem we are getting more events here than we should be. Closing the PR.

PR #1058, opened 2025-07-23
Remove mention to admin extra user role

marceloneppel:
## Checklist

### DIFF
--- a/docs/explanation/users.md
+++ b/docs/explanation/users.md
@@ -109,13 +109,7 @@ If password rotation is required for users used in relations, the relation must 
 
 ### Extra user roles
 
-When an application charm requests a new user through the relation/integration it can specify that the user should have the `admin` role in the `extra-user-roles` field. The `admin` role enables the new user to read and write to all databases (for the `postgres` system database it can only read data) and also to create and delete non-system databases.
-
-```{note}
-`extra-user-roles` is only supported by the modern interface `postgresql_client`. It is not supported for the legacy `pgsql` interface. R
-
-Read more about the supported charm interfaces in [](/explanation/interfaces-and-endpoints).
-```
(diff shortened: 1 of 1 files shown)


docs/explanation/users.md
```diff
@@ -109,13 +109,7 @@ If password rotation is required for users used in relations, the relation must
...
 ### Extra user roles
 
-When an application charm requests a new user through the relation/integration it can specify that the user should have the `admin` role in the `extra-user-roles` field. The `admin` role enables the new user to read and write to all databases (for the `postgres` system database it can only read data) and also to create and delete non-system databases.
-
-```{note}
-`extra-user-roles` is only supported by the modern interface `postgresql_client`. It is not supported for the legacy `pgsql` interface. R
-
-Read more about the supported charm interfaces in [](/explanation/interfaces-and-endpoints).
-```
+When an application charm requests a new user through the relation/integration, it can specify that the user should be part of a predefined role to give them additional permissions. Please check "[Charm Roles explanations](/explanation/roles)" for the list of available roles.
```
a-velasco, line 112:
```suggestion
When an application charm requests a new user through the relation/integration, it can specify that the user should be part of a predefined role to give them additional permissions. Please check [](/explanation/roles) for the list of available roles.
```

Style note: 
* In general, links to other pages (internal and external) don't need to be wrapped in quotes
* I recommend using this format with empty square brackets for links to internal pages: `[](<path>)` - Sphinx will automatically fill it with the title of the linked page. Easier, faster, and we get the bonus side effect of consistency in the way we refer to pages.

a-velasco (commented):
Thank you @marceloneppel! Just left a small comment about link formatting.

taurus-forever (approved):
LGTM, but please address Andreia sugestions!

docs/explanation/users.md
marceloneppel: Thanks for the tips, @a-velasco!

a-velasco (approved):

PR #1062, opened 2025-07-25
Fix broken MicroStack links

a-velasco:
## Issue
MicroStack docs migration led to some broken links in the Sunbeam guide.

## Solution
Replaced MicroStack links with their equivalents in the [new OpenStack docs](https://canonical-openstack.readthedocs-hosted.com/en/latest/).

Would be great if someone could double-check that they are correct in the context of that guide. There were no automatic redirects, so I searched for them manually.

## Checklist

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -28,6 +28,7 @@ cjk
 codebase
 config
 configs
+CREATEDB
 cryptographically
 CSR
 CSRs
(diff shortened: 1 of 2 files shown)


marceloneppel (approved):
Thanks, @a-velasco!

I checked the links, and those are the ones that make sense.

I also asked about those links at https://github.com/canonical/canonical.com/issues/1794#issuecomment-3114342877.

PR #1072, opened 2025-07-31
[DPE-7594] Add custom users to pg_hba filter

dragomirp:
Don't skip custom users when generating pg_hba configuration

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -2458,9 +2458,16 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             return {USER: "all", REPLICATION_USER: "all", REWIND_USER: "all"}
         user_database_map = {}
         try:
-            for user in self.postgresql.list_users_from_relation(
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -2458,9 +2458,16 @@ def relations_user_databases_map(self) -> dict:
...
+            for user in self.postgresql.list_users(current_host=self.is_connectivity_enabled):
+                if user in (
+                    "backup",
+                    "monitoring",
+                    "operator",
+                    "postgres",
+                    "replication",
+                    "rewind",
+                ):
+                    continue
```
dragomirp, line 2470: Copied from the trigger.

taurus-forever: Please create a dedicated Jira ticket on Mohamed to decide the behavior here, as we are adding here backward compatibility to the current behavior which is not the best way to do in modernly hardened charm.

Thank you!

marceloneppel (approved):

PR #1087, opened 2025-08-11
Refresh v3 documentation

a-velasco:
## Issue
There is no documentation about how to perform minor upgrades and rollbacks on PostgreSQL 16. 

## Solution
* Update "Upgrades" landing page with some context around refresh
* Create new "How to perform a minor upgrade" guide
* Create new "How to perform a minor rollback" guide
* Create a "Glossary" page for all docs, which for now only contains some simplified refresh terminology for users 
* Update releases page

I tried to make it easy to skim through, but still emphasizing critical points. I avoided large blocks of text in favor of referencing other documentation and resources.

It does not cover every single nuance, but my intention is to merge a first iteration of refresh documentation as soon as possible so that users have something to part from. I expect it to grow and evolve in future PRs.

## Checklist

### DIFF
--- a/docs/reference/glossary.md
+++ b/docs/reference/glossary.md
@@ -1,0 +1,31 @@
+## Glossary
+
+Common terminology in the context of Charmed PostgreSQL.
+
+### Refresh
+
+This is a simplified summary of refresh terminology. 
+
+For a more detailed glossary, see the [charm refresh  developer documentation](https://canonical-charm-refresh.readthedocs-hosted.com/latest/glossary/). 
+
+```{glossary}
+refresh
+    `juju refresh` to a different workload and/or charm version.
+
+    Note: *rollback*, *upgrade*, and *downgrade* are specific types of refresh.
+
+upgrade
+    `juju refresh` to a higher workload and/or charm version.
+
+downgrade
+    `juju refresh` to a lower workload and/or charm version.
+
+rollback
+    `juju refresh` to the original workload and charm version while a refresh is in progress.
+
+workload
+    A software component that the charm operates. E.g. PostgreSQL.
+
+original version
+    Workload and/or charm version of all units before initiating a refresh process
+```
--- a/docs/how-to/index.md
+++ b/docs/how-to/index.md
@@ -45,6 +45,8 @@ Other deployment scenarios and configurations:
 
 ## Upgrades
 * [How to upgrade]
+    * [Perform a minor upgrade]
+    * [Perform a minor rollback]
 
 ## Cross-regional (cluster-cluster) async replication
 
@@ -95,6 +97,8 @@ This section is for charm developers looking to support PostgreSQL integrations 
 [Enable profiling]: /how-to/monitoring-cos/enable-profiling
 
 [How to upgrade]: /how-to/upgrade/index
+[Perform a minor upgrade]: /how-to/upgrade/minor-upgrade
+[Perform a minor rollback]: /how-to/upgrade/minor-rollback
 
 [Cross-regional async replication]: /how-to/cross-regional-async-replication/index
 [Set up clusters]: /how-to/cross-regional-async-replication/set-up-clusters
--- a/docs/how-to/index.md
+++ b/docs/how-to/index.md
@@ -114,7 +114,6 @@ This section is for charm developers looking to support PostgreSQL integrations 
 ```{toctree}
 :titlesonly:
 :maxdepth: 2
-:glob:
 :hidden:
 
 Deploy <deploy/index>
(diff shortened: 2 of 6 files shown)


docs/how-to/refresh.md
```diff
@@ -0,0 +1,239 @@
...
+In some of these cases, it may be possible to perform an out-of-place upgrade or downgrade.
+
+* Minor in-place downgrade from PostgreSQL 16.10 to 16.9
+* Major in-place upgrade from PostgreSQL 14 to 16
+* Major in-place downgrade from PostgreSQL 16 to 14
+* Any refresh from or to a non-stable version (e.g. 16/edge)
+
+## Create a backup
+
+Create a [backup](/how-to/back-up-and-restore/create-a-backup).
```
a-velasco, line 62:
```suggestion
See [](/how-to/back-up-and-restore/create-a-backup).
```

docs/how-to/refresh.md
```diff
@@ -0,0 +1,239 @@
...
++==============+============+==========+==============+============+==========+===============================================================================================+
+| 843 (amd64)  | 16.9       | 201, 202 | TODO (amd64) | 16.9       | TODO     | `TODO, TODO <https://github.com/canonical/postgresql-operator/releases/tag/v16%2F1.60.0>`__   |
++--------------+            |          +--------------+            |          | `TODO, TODO <https://github.com/canonical/postgresql-operator/releases/tag/v16%2F1.61.0>`__   |
+| 844 (arm64)  |            |          | TODO (arm64) |            |          |                                                                                               |
++--------------+------------+----------+--------------+------------+----------+-----------------------------------------------------------------------------------------------+
+```
+
+### Supported refreshes
+
+These refreshes should be supported.
```
a-velasco, line 34: I'd prefer to write something less vague, like "These versions are expected to work, but are not officially tested or guaranteed".

docs/how-to/refresh.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Review release notes
+
+For every charm version between the version that you are refreshing from and to—and for the version you are refreshing to, review the release notes to understand what changed and if any action is required from you before, during, or after the refresh.
+
+For [recommended refreshes](#recommended-refreshes), refer to the rightmost column of the table.
+
+If the PostgreSQL versions that you are refreshing from and to are different, refer to the [upstream PostgreSQL release notes](https://www.postgresql.org/docs/release/) to understand what changed and if any action is required from you.
+
+## Test on staging environment
```
a-velasco, line 83:
```suggestion
## Test in a staging environment
```

docs/how-to/refresh.md
```diff
@@ -0,0 +1,239 @@
...
+```shell
+juju run postgresql/leader pre-refresh-check
+```
+
+If the action does not succeed, do not refresh.
+
+If the action succeeds, copy down the rollback command.
+Keep the command available in case you need to [roll back](#roll-back).
+
+## Configure pause-after-unit-refresh
```
a-velasco, line 128:
```suggestion
## Configure `pause-after-unit-refresh`
```

docs/how-to/refresh.md
```diff
@@ -0,0 +1,239 @@
...
+## Start the refresh
+
+Use `juju refresh` and specify the charm revision that you are refreshing to.
+
+```shell
+juju refresh postgresql --revision 0
+```
+
+## Roll back
+
```
a-velasco, line 173: We should generally avoid consecutive headings, and instead add a little bit of context introducing the parent section. 

One suggestion could be to take the first sentence of "Start the rollback" and put it here instead:
```suggestion
## Roll back

If something went wrong, the safest recovery path is often to roll back to the original version.
```

docs/reference/glossary.md
```diff
@@ -0,0 +1,3 @@
+# Glossary
```
a-velasco, line 1: Let's fully remove this file for now; I'll re-add it in a different PR with non-refresh terms that we are ready to define.

a-velasco (commented):
Thank you for your work here, @carlcsaposs-canonical. After reading through the whole process carefully, I think your changes to the structure and sequence makes a lot of sense.

Most of my comments are nitpicks and small things. The "big" thing left to do for this PR imo is just converting the file to .rst so that the tables and links are more manageable.

reneradoi (commented):
Nice and extensive docs! 

A general suggestions I'd like to add is that I'd like to see this whole document divided into two parts: "preparation" and "during execution" (would start at "Pre-refresh check). This would make it easier for users to digest the long list of instructions and for us to group all of them a bit more logically.

Other than that, I only have some questions and a few comments to details.

PR #1097, opened 2025-08-13
Change config options to dashes & update charm-refresh to v3.1.0.0

carlcsaposs-canonical:
- Update charm refresh to v3.1.0.0
- Use pause-after-unit-refresh config option instead of pause_after_unit_refresh
- For all other config options, switch to dashes with backwards-compatible fallback to previous name with underscores


taurus-forever (approved):
Thank you for this PR Carl!
LGTM from my side, curios for CI test results... btw, can you please add a test to make sure `Prefer value of option name with dash (-) and fallback to name with underscore (_)` works as expected.

Also, should we move tests from _ to - be default?

Sure, @marceloneppel and @dragomirp to agree on the direction.

dragomirp (approved):

marceloneppel (approved):

src/charm.py
```diff
@@ -257,6 +257,22 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
...
+        """Return a config instance validated and parsed using the provided pydantic class."""
+        config = {
+            # Prefer value of option name with dash (-) and fallback to name with underscore (_)
+            config_option: self.model.config.get(
+                config_option.replace("_", "-"), self.model.config.get(config_option)
+            )
+            for config_option in self.config_type.keys()  # noqa: SIM118
+        }
+        config = {
+            config_option: value for config_option, value in config.items() if value is not None
```
taurus-forever, line 272: @carlcsaposs-canonical  from:
https://chat.canonical.com/canonical/pl/webycyzcrjr4pkfwaw57ca7diw

For boolean options, it always set to False:
```
$ juju config postgresql plugin-timescaledb-enable     
false
 
$ juju config postgresql plugin_timescaledb_enable
true
```

which can lead to disabled plugin when enabled in one/wrong place only... can we fix this?

carlcsaposs-canonical: > For boolean options, it always set to False:

I believe this is incorrect
```
ubuntu@juju-378fe5-stg-carlcsaposs-dev-env-77:~$ juju deploy postgresql --channel 16/edge
Deployed "postgresql" from charm-hub charm "postgresql", revision 1097 in channel 16/edge on ubuntu@24.04/stable
ubuntu@juju-378fe5-stg-carlcsaposs-dev-env-77:~$ juju config postgresql plugin-timescaledb-enable

ubuntu@juju-378fe5-stg-carlcsaposs-dev-env-77:~$ juju config postgresql plugin_timescaledb_enable
false
```

> can we fix this?

I'm a bit confused, what do you have in mind?

the current behavior is that options with `-` override options with `_`.
To preserve backwards compatibility:
- options with `-` do not have a default value so that it requires user action (using `-` option) before overriding
- options with `_` maintain their previous default values

PR #1135, opened 2025-09-02
[MISC] Don't block if the snap cannot be installed

dragomirp:
## Issue
If the snap fails to install the charm will attempt to block, but subsequent hooks will error out and there should be no way to automatically recover

## Solution
Don't handle the snap installation error so that the install hook can fail and can be retried with `juju resolve`

## Checklist

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -119,22 +119,6 @@ def test_on_install(harness):
         assert isinstance(harness.model.unit.status, WaitingStatus)
 
 
-def test_on_install_snap_failure(harness):
-    with (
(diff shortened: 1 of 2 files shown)


tests/unit/test_charm.py
```diff
@@ -119,22 +119,6 @@ def test_on_install(harness):
         assert isinstance(harness.model.unit.status, WaitingStatus)
 
 
-def test_on_install_snap_failure(harness):
```
taurus-forever, line 122: we should have test(s) for failed snap instannation to make sure `juju resolve -all` fixes it.

taurus-forever (approved):

PR #1136, opened 2025-09-02
[MISC] Don't block if the snap cannot be installed

dragomirp:
Port of https://github.com/canonical/postgresql-operator/pull/1135 to 14/edge

## Issue
If the snap fails to install the charm will attempt to block, but subsequent hooks will error out and there should be no way to automatically recover

## Solution
Don't handle the snap installation error so that the install hook can fail and can be retried with `juju resolve`

## Checklist


taurus-forever (approved):
Same as https://github.com/canonical/postgresql-operator/pull/1135#discussion_r2316226419
LGTM

PR #1137, opened 2025-09-04
[DPE-7584] Fix temp tablespace permissions

marceloneppel:
## Issue
When using `tmpfs` for the `temp` storage, if we reboot the host machine, and later, after the charm starts again, it's not possible to create temporary objects, like temporary tables.

The reason is that after the reboot, the storage is mounted with wrong ownership and permissions (owned by `root` instead of `_daemon_` and with permissions set to `755` instead of `700`).

## Solution
Fix the permissions when the charm starts again.

I'll create a follow-up PR in the single kernel library repo to port those changes.

## Checklist

### DIFF
--- a/lib/charms/postgresql_k8s/v1/postgresql.py
+++ b/lib/charms/postgresql_k8s/v1/postgresql.py
@@ -1060,6 +1060,7 @@ class PostgreSQL:
             cursor = connection.cursor()
 
             if temp_location is not None:
+                # Fix permissions on the temporary tablespace location when a reboot happens and tmpfs is being used.
                 user = pwd.getpwnam("_daemon_")
                 os.chown(temp_location, uid=user.pw_uid, gid=user.pw_gid)
                 os.chmod(temp_location, 0o700)


lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -1058,6 +1060,11 @@ def set_up_database(self, temp_location: Optional[str] = None) -> None:
             cursor = connection.cursor()
 
             if temp_location is not None:
+                # Fix permissions on the temporary tablespace location when a reboot happens and tmpfs is being used.
+                user = pwd.getpwnam("_daemon_")
+                os.chown(temp_location, uid=user.pw_uid, gid=user.pw_gid)
```
taurus-forever, line 1065: Should _change_owner() be moved to lib and used to avoid duplicates?

lib/charms/postgresql_k8s/v1/postgresql.py
```diff
@@ -1058,6 +1060,11 @@ def set_up_database(self, temp_location: Optional[str] = None) -> None:
...
+                os.chmod(temp_location, 0o700)
```
taurus-forever, line 1066: Also, it is 750 [there](https://github.com/canonical/postgresql-operator/blob/cc1a6239cd77e84951438db0e51cbf68bc45f953/src/relations/async_replication.py#L663)

Are we consistent?

lib/charms/postgresql_k8s/v1/postgresql.py
marceloneppel: I believe it's a good idea. I'd change it a bit and create a different Python file (and also move it to the single kernel library repo) to contain utilities related to the filesystem, and keep this library containing only things related to the database connection and interaction.

marceloneppel:
Superseeded by https://github.com/canonical/postgresql-single-kernel-library/pull/10.

PR #1146, opened 2025-09-09
[MISC] Add back raft encryption check on upgrade

dragomirp:
Add back check for RAFT encryption when upgrading to prevent the upgrade getting stuck due to 503 errors from the Patroni API.

## Checklist

### DIFF
--- a/src/upgrade.py
+++ b/src/upgrade.py
@@ -157,6 +157,14 @@ class PostgreSQLUpgrade(DataUpgrade):
             self.set_unit_failed()
             return
 
+        raft_encryption = (
+            int(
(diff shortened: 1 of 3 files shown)


src/upgrade.py
```diff
@@ -157,6 +157,14 @@ def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
             self.set_unit_failed()
             return
 
+        raft_encryption = (
+            int(
+                json.loads(self.peer_relation.data[self.charm.app].get("dependencies", "{}"))
+                .get("charm", {})
+                .get("version", 0)
+            )
+            < 3
```
dragomirp, line 166: We bumped to 3 with the RAFT password.

PR #1156, opened 2025-09-11
[DPE-7584] Update single kernel library to fix tmpfs ownership and permission after reboot

marceloneppel:
## Issue
If temp storage is using tmpfs and the unit is rebooted, that storage is remounted with wrong permissions, leading to failures when a database user tries to create temporary objects in the database.

## Solution
Update the single kernel library dependency to fix the issue.

Add an integration test for checking that the tmpfs storage works after a reboot.

Depends on https://github.com/canonical/postgresql-single-kernel-library/pull/11.

## Checklist


pyproject.toml
```diff
@@ -19,7 +19,7 @@ pysyncobj = "^0.3.14"
 psutil = "^7.0.0"
 charm-refresh = "^3.1.0.2"
 httpx = "^0.28.1"
-postgresql-charms-single-kernel = "^0.0.1"
+postgresql-charms-single-kernel = {url = "https://github.com/canonical/postgresql-single-kernel-library/archive/a8a9cdaedbc11365773a80de7aa3b6bef58941f4.zip"}
```
taurus-forever, line 22: 16.0.2?

taurus-forever (approved):
LGTM, should you use proper lib version?

PR #1162, opened 2025-09-17
[DPE-8320] Fix backups with internal certificates

dragomirp:
* Prevent early generation of the internal certificate
* Use peer address as comon name in peer certificates
* Update charm libs

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -32,6 +32,8 @@ from charms.grafana_agent.v0.cos_agent import COSAgentProvider, charm_tracing_co
 from charms.operator_libs_linux.v2 import snap
 from charms.rolling_ops.v0.rollingops import RollingOpsManager, RunWithLock
 from charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm
+from cryptography.x509 import load_pem_x509_certificate
+from cryptography.x509.oid import NameOID
 from ops import (
     ActionEvent,
     ActiveStatus,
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -1403,7 +1416,6 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:  # noqa: C901
 
         if not self.get_secret(APP_SCOPE, "internal-ca"):
             self.tls.generate_internal_peer_ca()
-            self.tls.generate_internal_peer_cert()
```
dragomirp, line 1406: If the internal cert generates before the IP is set to the peer data, the hostname will be used as common name. It also causes issues on Juju 4.

tests/integration/test_backups_ceph.py
```diff
@@ -167,9 +164,9 @@ async def test_backup_ceph(ops_test: OpsTest, cloud_configs, cloud_credentials,
     await backup_operations(
         ops_test,
         S3_INTEGRATOR_APP_NAME,
-        tls_certificates_app_name,
-        tls_config,
-        tls_channel,
+        None,
+        None,
+        None,
```
dragomirp, line 169: Test on microceph with the internal certs, without using a TLS operator.

src/charm.py
```diff
@@ -378,6 +380,17 @@ def _post_snap_refresh(self, refresh: charm_refresh.Machines):
...
         Called after snap refresh
         """
+        try:
+            if raw_cert := self.get_secret(UNIT_SCOPE, "internal-cert"):
+                cert = load_pem_x509_certificate(raw_cert.encode())
+                if (
+                    cert.subject.get_attributes_for_oid(NameOID.COMMON_NAME)[0].value
+                    != self._unit_ip
+                ):
+                    self.tls.generate_internal_peer_cert()
```
taurus-forever, line 390: Please add INFO-like message to log to see when cert is regenerated (to simplify production troubleshooting). Warning is also fine.

taurus-forever (approved):
LGTM, please add one more message as we are going to regenerate cert.

PR #1169, opened 2025-09-19
[TEST] No reinit config hash

dragomirp:
## Checklist


taurus-forever (approved):
Please remove  the logic to `restart the workload if it's stuck on the starting state after a timeline divergence due to a backup that was restored.`.

PR #1183, opened 2025-09-23
[MISC] Allow deployment on Juju 3.1

dragomirp:
## Issue
A lot of users are still on Juju 3.1 so they cannot deploy newer revisions of the charm

## Solution
Allow deployment on Juju 3.1 or higher

## Checklist


dragomirp:
Doesn't seem to be needed for now. Will redo if 3.1 gets requested.

PR #1194, opened 2025-10-01
TEST Patroni 3.3.2

taurus-forever:
TESTING.


taurus-forever:
Tests passed, new PRs will be created after the snap publishing

Issue #1180, opened 2025-09-23
secret revisions clean up


ycliuhw:
Hi,
I noticed that this charm creates secrets, but it doesn’t implement handlers for the secret_removed event. 
As a result, all obsolete revisions remain indefinitely, even if they’re no longer used or consumed by anyone.

We should consider implementing the cleanup hook by watching the secret_removed event.

Thanks!

ycliuhw:
Yes, then the charm will be able to clean up any obsolete/unused revisions by itself.

jameinel:
@kot0dama Juju 2.9 doesn't have support for *Juju* secrets. However, for versions of Juju which do have secret support, we would, indeed, be creating Kubernetes secrets as a result of this, and not implementing `secret-remove` would mean that we would create lots of K8s secret artifacts that did not get cleaned up.

manadart:
This issue has for some time and continues to, contribute to PS6 being on fire. 

IS and others trying to diagnose and rectify the issues are being hampered by the behaviour of this charm. A swift resolution is desired by all.

manadart:
> Can we manually clean old Juju Secret revisions on PS6 to decrease temperature there? Tnx!

@jameinel worked with the SREs to do this. It made a significant improvement to performance, but we could see that they start to accumulate again quite quickly.

taurus-forever:
> Note that we also should look at why we are still creating new revisions. ...

Quick look on PostgreSQL 14/edge: adding/removing node member from cluster:

```
> diff -u <(juju show-secret d3f7fb07ln9deqvhit1g --reveal --revision 182) <(juju show-secret d3f7fb07ln9deqvhit1g --reveal --revision 98)
...
   content:
     password: secretpassword
-    read-only-uris: postgresql://relation-4:secretpassword@10.198.5.134:5432/postgresql_test_app_database
+    read-only-uris: postgresql://relation-4:secretpassword@10.198.5.121,10.198.5.134:5432/postgresql_test_app_database
     uris: postgresql://relation-4:secretpassword@10.198.5.237:5432/postgresql_test_app_database
     username: relation-4
 ...
```

Another secret change is a primary switchover:
```
> diff -u <(juju show-secret d3f7fb07ln9deqvhit1g --reveal --revision 182) <(juju show-secret d3f7fb07ln9deqvhit1g --reveal --revision 188)
...
   content:
     password: secretpassword
[...]

PR #1195, opened 2025-10-02
Remove old revision of secret

marceloneppel:
## Issue
The charm does not remove the old revisions of a secret after updating it.

## Solution
Implement the handler for the `secret-remove` event. https://github.com/juju/juju/issues/20794 is taken into account.

Implement the removal of the charm secrets' old revision in the upgrade logic. It's possible that not all the old revisions will be removed in an environment due to https://github.com/juju/juju/issues/20782. For example, if the latest revision of a secret is 10 and there is an old one with the revision number equal to 1, that old revision won't be removed because its number is the prefix of the latest one.

Fixes https://github.com/canonical/postgresql-operator/issues/1180.

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -214,6 +214,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             model=get_postgresql_dependencies_model(),
             relation_name="upgrade",
             substrate="vm",
+            run_cmd=run_cmd,
         )
         self.postgresql_client_relation = PostgreSQLProvider(self)
         self.legacy_db_relation = DbProvides(self, admin=False)
--- a/src/charm.py
+++ b/src/charm.py
@@ -214,7 +214,6 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             model=get_postgresql_dependencies_model(),
             relation_name="upgrade",
             substrate="vm",
-            run_cmd=run_cmd,
         )
         self.postgresql_client_relation = PostgreSQLProvider(self)
         self.legacy_db_relation = DbProvides(self, admin=False)
--- a/src/charm.py
+++ b/src/charm.py
@@ -46,6 +46,7 @@ from ops.charm import (
     InstallEvent,
     LeaderElectedEvent,
     RelationDepartedEvent,
+    SecretRemoveEvent,
     StartEvent,
 )
 from ops.framework import EventBase
@@ -201,6 +202,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         self.framework.observe(self.on.set_password_action, self._on_set_password)
         self.framework.observe(self.on.promote_to_primary_action, self._on_promote_to_primary)
         self.framework.observe(self.on.update_status, self._on_update_status)
+        self.framework.observe(self.on.secret_remove, self._on_secret_remove)
         self.cluster_name = self.app.name
         self._member_name = self.unit.name.replace("/", "-")
 
@@ -434,6 +436,9 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         # For now, as there is no DNS hostnames on VMs, and it would also depend on
         # the underlying provider (LXD, MAAS, etc.), the unit IP is returned.
         return self._unit_ip
+
+    def _on_secret_remove(self, event: SecretRemoveEvent) -> None:
+        event.remove_revision()
 
     def _on_get_primary(self, event: ActionEvent) -> None:
         """Get primary instance."""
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -435,6 +438,9 @@ def get_hostname_by_unit(self, _) -> str:
         # the underlying provider (LXD, MAAS, etc.), the unit IP is returned.
         return self._unit_ip
 
+    def _on_secret_remove(self, event: SecretRemoveEvent) -> None:
+        event.remove_revision()
```
jameinel, line 442: When I was testing an implementation like this with my own charms, I ran into an issue with juju 3.6.9. We have a bugfix prepared that should make this implementation ok, but we might want to be defensive in how we write our charm.
https://github.com/juju/juju/pull/20796 is the bug fix and
https://github.com/juju/juju/issues/20794 is the issue.

Specifically, the issue is that older versions of juju, when a secret is fully-removed, will trigger secret-remove with the old revisions of the secret. However, that secret no longer exists, so that will cause the hook to fail, because it tries to delete something that doesn't exist.

I do see that you have some code to try and handle some of that:
```
    def remove(self, label: str) -> None:
        """Remove a secret from the cache."""
        if secret := self.get(label):
            try:
                secret.remove()
                self._secrets.pop(label)
            except (SecretsUnavailableError, KeyError):
                pass
            else:
                return
        logging.debug("Non-existing Juju Secret was attempted to be removed %s", label)
```
However, in my testing, because Juju tries to make the removal of a secret revision transaction consistent, it doesn't actually do anything until the hook completes, at which point, the charm has no way of resolving it (hence the above bug).

The fix that I did in my test charm was:
```
val=`secret-get $JUJU_SECRET_ID`
if [ -n "$val" ] ; then
  secret-remove --revision $JUJU_SECRET_REVISION $JUJU_SECRET_ID;
else
  juju-log -l WARNING secret $JUJU_SECRET_ID already deleted;
fi
```

I'm not a huge fan of the workaround. (Not least of which, it means that your application now becomes an observer of the charm content, so it will also get `secret-changed` events. I suppose you might be able to use `--peek`? I'm not sure if that sets you as an observer.)

src/upgrade.py
```diff
@@ -261,6 +267,29 @@ def _prepare_upgrade_from_legacy(self) -> None:
...
+            for revision in range(1, latest_revision):
+                command = [
+                    self.run_cmd,
+                    self.charm.unit.name,
+                    "--",
+                    "secret-remove",
+                    "--revision",
+                    str(revision),
+                    secret_id,
+                ]
```
jameinel, line 288: This also won't work. 
I tested this with just manually running hooks, and juju "queues up" a single revision to be removed. So if you just do:
```sh
juju exec --unit u/0 -- secret-remove --revision 1 $secret; secret-remove --revision 2 $secret
```
Then juju will only remove revision 2.

jameinel: I filed a bug about this behavior:
https://github.com/juju/juju/issues/20805

marceloneppel: I noticed a different behaviour between the approach you commented about and the one that uses `/usr/bin/juju-exec u/0 -- secret-remove --revision 1 $secret; /usr/bin/juju-exec u/0 -- secret-remove --revision 2 $secret` from inside the unit SSH or ops hook context.

The latter approach "queues up" all the revisions to be removed.

src/upgrade.py
```diff
@@ -261,6 +268,59 @@
...
+            for id_or_label in ids_and_labels:
+                if id_or_label.startswith("secret://"):
+                    secret = self.charm.model.get_secret(id=id_or_label)
+                    label = secret.get_info().label
+                else:
+                    try:
+                        secret = self.charm.model.get_secret(label=id_or_label)
+                        label = id_or_label
+                    except SecretNotFoundError:
+                        logger.debug(f"Skipping secret not found with label {id_or_label}")
```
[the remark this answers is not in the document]
marceloneppel: The label is not sensitive.

dragomirp (approved):

marceloneppel:
@taurus-forever, I removed the upgrade logic on https://github.com/canonical/postgresql-operator/pull/1195/commits/fec93f20658a94d1ec16e6ca8c71ca4bb34121fe and tested the remaining workaround in the `secret-remove` event handler in the following Juju versions.

- 3.6.11 (built from commit 37e3e7fec0dd8f287e55f03b4055bad2a927697e - coming from https://github.com/juju/juju/pull/20796 - mentioned on https://github.com/canonical/postgresql-operator/pull/1195#discussion_r2404168116)
- 3.6.9
- 3.5.7
- 3.4.6
- 3.3.7
- 3.2.4
- 3.1.10

It's working on all of them (the revisions are correctly removed, and no error happens when a secret is removed entirely, and the charm runs the `secret-remove` event handler).

taurus-forever:
Fixed for 14/edge (merged) and porting to PG16 VM and PG K8s 14+16 (and PGB).

P.S. See the last messages:
https://github.com/canonical/postgresql-operator/pull/1195#issuecomment-3378342378
https://github.com/canonical/postgresql-operator/pull/1195#pullrequestreview-3313287670

PR #1196, opened 2025-10-02
[MISC] Set strict sync mode

dragomirp:
Enable strict sync mode in Patroni.

## Checklist

### DIFF
--- a/src/cluster.py
+++ b/src/cluster.py
@@ -5,6 +5,7 @@
 """Helper class used to manage cluster lifecycle."""
 
 import glob
+import json
 import logging
 import os
 import pwd
@@ -1057,13 +1058,14 @@ class Patroni:
     def update_synchronous_node_count(self) -> None:
         """Update synchronous_node_count to the minority of the planned cluster."""
         # Try to update synchronous_node_count.
+        member_units = json.loads(self.charm.app_peer_data.get("members_ips", "[]"))
         for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
             with attempt:
                 r = requests.patch(
                     f"{self._patroni_url}/config",
                     json={
                         "synchronous_node_count": self._synchronous_node_count,
-                        "synchronous_mode_strict": self._synchronous_node_count > 0,
+                        "synchronous_mode_strict": len(member_units) > 1,
                     },
                     verify=self.verify,
                     auth=self._patroni_auth,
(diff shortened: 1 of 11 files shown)


src/cluster.py
```diff
@@ -1057,11 +1058,16 @@ def _synchronous_node_count(self) -> int:
     def update_synchronous_node_count(self) -> None:
         """Update synchronous_node_count to the minority of the planned cluster."""
         # Try to update synchronous_node_count.
+        member_units = json.loads(self.charm.app_peer_data.get("members_ips", "[]"))
```
dragomirp, line 1061: Check the size of the members list in the peer data. This is necessary to make sure that the cluster successfully botstrapped.

templates/patroni.yml.j2
```diff
@@ -59,6 +59,7 @@ bootstrap:
     retry_timeout: 10
     maximum_lag_on_failover: 1048576
     synchronous_mode: true
+    synchronous_mode_strict: false
```
dragomirp, line 62: Start as false to let the cluster bootstrap. Since this will be updated in the DCS as soon as the cluster is running, should we have it in the yaml file?

marceloneppel: I think it's not needed, as you commented. But I don't see any problem with keeping it here either, so you can keep it.

tests/integration/ha_tests/test_scaling.py
```diff
@@ -87,12 +87,12 @@ async def test_removing_stereo_primary(ops_test: OpsTest, continuous_writes) ->
...
     await ops_test.model.wait_for_idle(status="active", timeout=600, idle_period=45)
 
-    await are_writes_increasing(ops_test, primary)
-
     logger.info("Scaling back up")
     await ops_test.model.applications[DATABASE_APP_NAME].add_unit(count=1)
     await ops_test.model.wait_for_idle(status="active", timeout=1500)
 
+    await are_writes_increasing(ops_test, primary)
+
```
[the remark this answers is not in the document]
marceloneppel: Just to be on the same page, is this a safe measure to avoid allowing writes to that unit and losing them if that unit crashes?

marceloneppel (approved):
LGTM!

templates/patroni.yml.j2
taurus-forever: No strict positioning here.

src/cluster.py
taurus-forever: I would go with a clean flag but  your approach looks working.

taurus-forever (approved):
LGTM ,tnx!

PR #1214, opened 2025-10-10
[DPE-8550] Check juju version before removing revision

dragomirp:
Don't remove revision unless on safe Juju version.

## Checklist


jameinel:
Juju 3.6.11 is now fully released (on Oct 20), so you should be able to test this more easily.

Issue #1231, opened 2025-10-15
Decrease amount of Goal State requests in PostgreSQL charm


taurus-forever:
## Steps to reproduce

Deploy postgresql charm and count amount of "goal-state" on controller level.

## Expected behavior

Avoid unnecessary (and expencive) "goal-state" calls.

## Actual behavior

(c) John Meinel

Specifically, we see a lot of calls to "goal-state" (30% load for Juju Controller from all charms).
What I was able to track down is that you have a "self._patroni" property, which instantiates a Patroni object and passes in the self.proposed_units(). However, the _patroni object isn't cached so code like this:
```
            # Get the current primary to check if it has changed later.
            current_primary = self._patroni.get_primary()

            # Trigger the switchover.
            self._patroni.switchover()

            # Wait for the switchover to complete.
            self._patroni.primary_changed(current_primary)
```
Is actually instantiating the _patroni object 3 times and calling goal-state 3 times.

## Versions

Operating system: 22.04
Juju CLI: 3.6.9
Juju agent: 3.6.9
Charm revision: 443
LXD: 5.21

PR #1249, opened 2025-10-21
[DPE-8573] Cache _patroni property

dragomirp:
Related to https://github.com/canonical/postgresql-operator/issues/1231

Cache the `_patroni` property to reduce the amount of controller checks.

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -905,7 +905,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         ]
         return set(hosts)
 
-    @property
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -905,7 +905,7 @@ def _hosts(self) -> set:
         ]
         return set(hosts)
 
-    @property
+    @cached_property
```
paulomach: sounds to good to be true - Is there any case where you need to invalidate the cache while in the hook, e.g. change in IP, or planned units within a hook?

PR #1303, opened 2025-11-20
[DPE-8949] Downgrade runner

dragomirp:
Downgrade arm64 container host to jammy due to https://gitlab.com/apparmor/apparmor/-/issues/571 (affects kernel 6.14)

## Checklist


spread.yaml
```diff
@@ -76,10 +76,9 @@ backends:
     allocate: |
       sudo tee /etc/ssh/sshd_config.d/10-spread-github-ci.conf << 'EOF'
       PasswordAuthentication yes
-      PermitEmptyPasswords yes
       EOF
 
-      sudo passwd -d runner
+      echo "runner:$SPREAD_PASSWORD" | sudo chpasswd
```
dragomirp: Could not get jammy to work without a password, so using the random spread pass to ssh into localhost.

PR #1316, opened 2025-11-26
[MISC] Switch to config hash and pin last known working jammy image

dragomirp:
* Switch to config hash restart like on 14/edge
* Pin jammy lxd image for arm

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -2456,7 +2456,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
                 logger.debug(
                     "Early exit update_config: patroni not responding but TLS is enabled."
                 )
-                self._handle_postgresql_restart_need()
+                self._handle_postgresql_restart_need(True)
                 return True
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -108,6 +108,12 @@ jobs:
         with:
           pattern: ${{ inputs.artifact-prefix }}-*
           merge-multiple: true
+      # TODO Remove pinned image when problem is resolved on jammy
+      - name: Pin arm lxc image for jammy
+        if: ${{ contains(matrix.job.spread_job, 'arm') }}
+        run: |
+          sudo lxc image copy ubuntu:3e7ef486d7a6 local:
+          sudo lxc image alias create "juju/ubuntu@22.04/arm64" 3e7ef486d7a6
```
dragomirp: Pinning instead of switching to the jammy runner, because switching the runner will require making the tests compatible with jammy's python version (pyproj limits, linters etc.).

Issue #1299, opened 2025-11-18
pg_hba for replication peer_ip too open


hloeung:
Hi,

Splitting this out from https://github.com/canonical/postgresql-operator/pull/1272, the generated netmask/cidr for replication peer_ip is too open at `/0`:

```
    {%- for peer_ip in peers_ips %}
    - {{ 'hostssl' if enable_tls else 'host' }}     replication    replication    {{ peer_ip }}/0    md5
    {% endfor %}
```

A `/0` allows all and I think should be `/32` though could probably get away with using `/24`. Thankfully it is authenticated with `md5` and in our Canonical IS deployed environments have additional layers of security with firewalling as well as OpenStack security groups.

PR #1317, opened 2025-11-26
Fix CIDR mask for replication peer_ip - #1299

hloeung:
## Issue

The `replication` users' `peer_ip` CIDR mask is too open. A `/0` usually means `ALL`.

See https://github.com/canonical/postgresql-operator/issues/1299

## Solution

Reduce CIDR mask to a single IP, with `/32` which is safest. If not, we can go with something a little more relaxed such as a `/24`.

## Checklist

### DIFF
--- a/templates/patroni.yml.j2
+++ b/templates/patroni.yml.j2
@@ -180,7 +180,7 @@ postgresql:
     - {{ 'hostssl' if enable_tls else 'host' }} replication replication {{ endpoint }}/32 md5
     {%- endfor %}
     {%- for peer_ip in peers_ips %}
-    - {{ 'hostssl' if enable_tls else 'host' }}     replication    replication    {{ peer_ip }}/0    md5
+    - {{ 'hostssl' if enable_tls else 'host' }}     replication    replication    {{ peer_ip }}/32    md5
     {% endfor %}
   pg_ident:
   - operator snap_daemon backup


hloeung:
Thanks, much appreciated. This isn't so urgent and more of an improvement - from IS's side, we have multiple layers of security already - password authentication which the charm by default uses, OpenStack security group firewalling as well as NGFW / dedicated firewalling in front of the databases we manage.

taurus-forever (commented):
The changes itself looks OK and logical, but it looks they brake TLS test at least ([here](https://github.com/canonical/postgresql-operator/actions/runs/19717189903/job/56626163550#step:6:1919)):
```
Exception: Expected command 'grep 'connection authorized: user=rewind database=postgres SSL enabled' /var/snap/charmed-postgresql/common/var/log/postgresql/postgresql-*.log' to succeed instead it failed: 1
```

So, it looks like PostgreSQL was not able to connect and rewind SQL transactions from peer. It requires careful re-checking before merging.


The self-healing test is also [unhappy](https://github.com/canonical/postgresql-operator/actions/runs/19717189903/job/56626163551#step:6:2048) after the full cluster restart (rewind in place?):
```
AssertionError: secondary not up to date with the cluster after restarting.
```

---

CC: @hloeung 

I believe the main trouble-maker here is a [cluster<->cluster async replication](https://canonical-charmed-postgresql.readthedocs-hosted.com/14/how-to/cross-regional-async-replication/) functionality, where IPs can be anything for different cluster. Sure `/0` security can be improved, but it looks like it is not as simple as `/32` nor `/24`.

taurus-forever:
@marceloneppel should this be abandoned due to the "proper" fix in https://github.com/canonical/postgresql-operator/pull/1506 ?

marceloneppel:
> @marceloneppel should this be abandoned due to the "proper" fix in #1506 ?

Yes. I'm closing this now.

PR #1318, opened 2025-11-27
DPE-8980 Support Juju 4: use 'ip' instead of 'private-address' (if available)

taurus-forever:
## Issue

No Juju 4 support by the charm.

## Solution

The Juju 4 has removed support databag fiesl `private-address`, `ingress-address` and more. The field we should use is `ip` now. The PG16 charm still have to support Juju 3.6 LTS, so adding support of the ip field with backward compatibility.

We are not going to allow Juju 4 deployments (assume in metadata.yaml), 
however users can use the following command to test Juju 4:
> juju deploy postgresql --channel 16/edge --force

### DIFF
--- a/src/relations/tls.py
+++ b/src/relations/tls.py
@@ -67,7 +67,9 @@ class TLS(Object):
             peer_addrs.add(addr)
         if addr := self.charm.unit_peer_data.get("replication-offer-address"):
             peer_addrs.add(addr)
-        if addr := self.charm.unit_peer_data.get("private-address"):
+        if addr := self.charm.unit_peer_data.get("ip") or self.charm.unit_peer_data.get(
+            "private-address"
+        ):
             peer_addrs.add(addr)
         return peer_addrs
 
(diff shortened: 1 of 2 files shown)


taurus-forever:
The issue https://github.com/juju/juju/issues/21344 has been fixed, and now PostgreSQL cluster can be deployed.
However no success with the relation to postgresql-test-app nor data-integrator :cry: 
Juju belives the relation has been established succesfully, but no juju secret created with app credentials => app stays pending for data... the further investigation required.

src/relations/tls.py
```diff
@@ -67,7 +67,9 @@ def _get_peer_addrs(self) -> set[str]:
             peer_addrs.add(addr)
         if addr := self.charm.unit_peer_data.get("replication-offer-address"):
             peer_addrs.add(addr)
-        if addr := self.charm.unit_peer_data.get("private-address"):
+        if addr := self.charm.unit_peer_data.get("ip") or self.charm.unit_peer_data.get(
+            "private-address"
+        ):
```
carlcsaposs-canonical:
```suggestion
        if addr := self.charm.unit_peer_data.get("ip", self.charm.unit_peer_data.get("private-address")):
```
nit

carlcsaposs-canonical (dismissed):

carlcsaposs-canonical (approved):

marceloneppel (approved):

dragomirp (approved):

Issue #794, opened 2025-03-17
Failures with `create-backup` doesn't surface the error


hloeung:
## Steps to reproduce

1. `juju run postgresql/5 create-backup --wait 60m`
1. `juju show-operation 109`

## Expected behavior

On errors, the full error should be shown, or at least a reason as to why it failed rather than an incomplete error message.

## Actual behavior

![Image](https://github.com/user-attachments/assets/ae4ff73b-b621-4d12-b31f-77e5290e5dc0)

## Versions

Operating system: ubuntu@22.04
Juju CLI: 3.6.3-ubuntu-amd64
Juju agent: 3.6.1
Charm revision: 14/stable      553

## Log output

Juju debug log:

hloeung:
In this case, it looks to be `ERROR: [082]: WAL segment 0000000A0000000000000060 was not archived before the 60000ms timeout`

![Image](https://github.com/user-attachments/assets/f622283e-7bb4-4764-aa42-03c08167e63b)

marceloneppel:
Hi, @hloeung! Revision 980 and 981 from `14/edge` contains the fix for this issue. Thanks for the report!

Issue #784, opened 2025-03-03
Units stuck `blocked` with `failed to initialize stanza, check your S3 settings`


hloeung:
## Steps to reproduce

Use Terraform plans to provision new databases - postgresql along with s3-integrator

## Expected behavior

All units should come up fine and showing as `active` and `idle` in Juju status.

## Actual behavior

In some places, it's likely a race condition where units show up as `blocked` with `failed to initialize stanza, check your S3 settings`

## Versions

Operating system:  Ubuntu 22.04

Juju CLI: 3.6.3-ubuntu-amd64

Juju agent: 3.6.1

Charm revision: r553 from latest/stable

## Additional context

```
is-managed-database-stg-data-mesh-el-salesforce@is-bastion-ps6:~$ juju status postgresql | grep '^postgresql/'
postgresql/0*         blocked   idle       3        10.146.64.69    5432/tcp  failed to initialize stanza, check your S3 settings
postgresql/1          blocked   idle       4        10.146.64.77    5432/tcp  failed to initialize stanza, check your S3 settings
postgresql/2          waiting   executing  5        10.146.64.68    5432/tcp  (leader-settings-changed) awaiting for member to start
is-managed-database-stg-data-mesh-el-salesforce@is-bastion-ps6:~$
```

```
dbaas-prod-ranger-db@is-bastion-ps6:~$ juju status postgresql | grep '^postgresql/'
postgresql/0*         active    idle   4        10.146.65.185   5432/tcp
postgresql/1          blocked   idle   5        10.146.65.179   5432/tcp  failed to initialize stanza, check your S3 settings
postgresql/2          active    idle   6        10.146.65.184   5432/tcp
dbaas-prod-ranger-db@is-bastion-ps6:~$
```

```
dbaas-stg-neppel@is-bastion-ps6:~$ juju status postgresql | grep '^postgresql/'
postgresql/0*         blocked   idle   3        10.146.64.106   5432/tcp  failed to initialize stanza, check your S3 settings
postgresql/1          active    idle   4        10.146.64.105   5432/tcp  Primary
postgresql/2          active    idle   5        10.146.64.104   5432/tcp
```

hloeung:
FWIW, the cluster looks healthy on these.

```
ubuntu@juju-58c37e-dbaas-stg-neppel-3:~$ sudo -H -u snap_daemon charmed-postgresql.patronictl -c /var/snap/charmed-postgresql/current/etc/patroni/patroni.yaml list
[...]

Issue #1280, opened 2025-11-07
Improve error-handling when failing to create a S3 backup


ggouzi:
When a backup fails, error message is not appended in debug-log

## Steps to reproduce
```
juju run postgresql/leader -- create-backup
```

## Expected behavior
- When running the create-backup action
```
Failed to backup PostgreSQL with error: <CONTENT OF THE ERROR MESSAGE FROM S3>
```

- When checking debug-log:
```
**unit-postgresql-0: 07:11:59 ERROR unit.postgresql/0.juju-log backups:Backup failed: Failed to backup PostgreSQL with error:** : <CONTENT OF THE ERROR MESSAGE FROM S3>
```

## Actual behavior
No error message appended. The fact the line ends with a `:` makes me think we are supposed to append the S3 error message.
```
juju run postgresql/0 -- create-backup
Action id 14 failed: Failed to backup PostgreSQL with error: 
```

debug-log: Same, no error message appended `backups:Backup failed: Failed to backup PostgreSQL with error:`
```
unit-postgresql-0: 07:11:58 ERROR unit.postgresql/0.juju-log backups:
unit-postgresql-0: 07:11:58 INFO unit.postgresql/0.juju-log backups:Uploading content to bucket=ddev-postgresql-backup, path=backup/ddev-postgresql.postgresql/20251107-071158F/backup.log
**unit-postgresql-0: 07:11:59 ERROR unit.postgresql/0.juju-log backups:Backup failed: Failed to backup PostgreSQL with error:** 
unit-postgresql-0: 07:12:01 WARNING unit.postgresql/0.juju-log tracing:charm tracing buffer exceeds max history length (100 events)
unit-postgresql-0: 07:12:04 WARNING unit.postgresql/0.juju-log tracing:charm tracing buffer exceeds max history length (100 events)
```

## Versions

OS 24.04
Juju 3.6.11
Postgresql 16/stable (rev 952)

## Additional information
Could this issue come from the fact the error message is too verbose and juju cannot append it ?
```
unit.postgresql/0.juju-log tracing:charm tracing buffer exceeds max history length (100 events)
```
backup.log file retrieved using command-line: https://pastebin.ubuntu.com/p/PQN3DD5CTp/

PR #1328, opened 2025-12-02
[DPE-6762] Catch archive command timeout

marceloneppel:
## Issue
If pgBackRest cannot archive a WAL file in less than 60 seconds, it will fail it's `check` command and the charm will be stuck in a blocked status.

## Solution
In the stanza check step, handle the error code `82` (Archive operation timeout) in the same way we already do for error code `49` (Cannot connect to host) in the stanza initialisation step: https://github.com/canonical/postgresql-operator/blob/6ffe0b4afded1eec7b73078108984f025f5673cb/src/backups.py#L643. This way, the user can fix the problem and run `juju resolve`.

Fixes https://github.com/canonical/postgresql-operator/issues/784
Fixes https://github.com/canonical/postgresql-operator/issues/794

Tested manually by adding `sleep 90 && ` before the actual archive command in `templates/patroni.yml.j2` and deploying the charm + connecting it to Microceph's RadosGW.

pgBackRest error codes extracted from source code: [error-codes.csv](https://github.com/user-attachments/files/23888713/error-codes.csv). I think we shouldn't turn the charm into an error state for every error code, so I'm limiting it to error code `82` for now. We may expand the list of error codes in which we do that in the charm in the future.

Also, sometimes the error message from pgBackRest is sent to stdout instead of stderr (and we don't log them, as seen in https://github.com/canonical/postgresql-operator/issues/1280), so it's important to check both. I ported the changes from https://github.com/canonical/postgresql-operator/pull/1320 to this PR to handle that situation.

## Checklist

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -750,9 +750,9 @@ class PostgreSQLBackups(Object):
                     if return_code != 0:
                         raise Exception(stderr)
             self.charm._set_primary_status_message()
-        except TimeoutError as e:
+        except TimeoutError:
             # Re-raise to put charm in error state (not blocked), allowing juju resolve
-            raise e
+            raise
         except Exception as e:
             # If the check command doesn't succeed, remove the stanza name
             # and rollback the configuration.


taurus-forever (dismissed):
Thank you!

src/backups.py
```diff
@@ -369,6 +370,41 @@ def result():
...
+
+        # If we found error/warning lines, return them joined
+        if error_lines:
+            return " ".join(error_lines)
+
+        # Otherwise return the last non-empty line from stderr or stdout
+        if stderr.strip():
+            return stderr.strip().splitlines()[-1]
+        if stdout.strip():
+            return stdout.strip().splitlines()[-1]
```
onkolahmet: Is there a scenario where we might want to see both `stdout` and `stderr` instead of just one?

tests/unit/test_backups.py
```diff
@@ -38,6 +38,82 @@ def harness():
...
+
+
+def test_extract_error_message_with_multiple_errors():
+    """Test extracting multiple ERROR/WARN lines."""
+    stdout = """P00 ERROR: first error
+P00 WARN: warning message
+P00 ERROR: second error"""
+    stderr = ""
+    result = PostgreSQLBackups._extract_error_message(stdout, stderr)
+    assert result == "ERROR: first error WARN: warning message ERROR: second error"
```
onkolahmet: Following up on my point about using a different separator than white space: For example, if a backup fails with 3 different errors, the current implementation returns:

`"ERROR: first error WARN: warning message ERROR: second error"`

Whereas it could be: (Semicolon-separated)
`"ERROR: first error; WARN: warning message; ERROR: second error"
`

Just from a readability point of view.

src/backups.py
```diff
@@ -369,6 +370,41 @@ def result():
...
+        error_lines = []
+        for line in combined_output.splitlines():
+            if "ERROR:" in line or "WARN:" in line:
+                # Clean up the line by removing debug prefixes like "P00  ERROR:"
+                cleaned = re.sub(r"^.*?(ERROR:|WARN:)", r"\1", line).strip()
+                error_lines.append(cleaned)
+
+        # If we found error/warning lines, return them joined
+        if error_lines:
+            return " ".join(error_lines)
```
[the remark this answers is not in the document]
marceloneppel: Great idea! I updated to use "; " as the separator on https://github.com/canonical/postgresql-operator/pull/1328/commits/1f1b6010db1f37952f4fc48be3e6c8216d6aa0a4.

tests/unit/test_backups.py
marceloneppel: Updated on https://github.com/canonical/postgresql-operator/pull/1328/commits/1f1b6010db1f37952f4fc48be3e6c8216d6aa0a4.

src/backups.py
```diff
@@ -692,15 +731,26 @@ def check_stanza(self) -> bool:
...
+                        extracted_error = self._extract_error_message(stdout, stderr)
+                        logger.error(
+                            f"error: {extracted_error} - please fix the error and call juju resolve on this unit"
+                        )
+                        raise TimeoutError
                     if return_code != 0:
                         raise Exception(stderr)
             self.charm._set_primary_status_message()
+        except TimeoutError as e:
+            raise e
```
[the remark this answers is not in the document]
marceloneppel: It is. If we don't catch it here, the next except clause will catch it (because that one checks for a broader Exception), and it will block the charm instead of putting in error state. In the blocked state the user cannot run something like `juju resolve` in the unit to make it retry the operation (in this case, the `pgbackrest check` call inside the charm code) after they fix the issue, like a network firewal not working properly.

src/backups.py
marceloneppel: I don't have any documented situation where we might need that, so just kept the logic for getting the error from both `stdout` and `stderr` when we see either the `WARN` or `ERROR` prefixes.

onkolahmet (approved):
LGTM! Thanks Marcelo!

src/backups.py
```diff
@@ -692,15 +732,27 @@ def check_stanza(self) -> bool:
...
+                        logger.error(
+                            f"error: {extracted_error} - please fix the error and call juju resolve on this unit"
+                        )
+                        raise TimeoutError
                     if return_code != 0:
                         raise Exception(stderr)
             self.charm._set_primary_status_message()
+        except TimeoutError as e:
+            # Re-raise to put charm in error state (not blocked), allowing juju resolve
+            raise e
```
carlcsaposs-canonical:
```suggestion
        except TimeoutError:
            # Re-raise to put charm in error state (not blocked), allowing juju resolve
            raise
```
nit

carlcsaposs-canonical (dismissed):

carlcsaposs-canonical (approved):

PR #1339, opened 2025-12-07
[MISC] Parallel cluster call

dragomirp:
* Update llibs and switch to snap lib package
* Split off TLS transfer from the internal TLS lib
* Switch to parallel calls to the cluster Patroni API with httpx

## Checklist


src/relations/tls_transfer.py
dragomirp: Splitting off from the Postgresql tls lib, so that we can also update the lib for PGB, which doesn't implement TLS transfer ATM.

carlcsaposs-canonical (commented):
looks good except disabling TLS verification

(will defer to @taurus-forever if it's safe to merge without TLS verification)

taurus-forever (dismissed):
LGTM! Great!

carlcsaposs-canonical (dismissed):

carlcsaposs-canonical (approved):

taurus-forever (approved):

PR #1378, opened 2026-01-15
[DPE-9205] Update endpoint in pre-refresh check

dragomirp:
Update the relation endpoints when switching over, so that clients don't have to wait for topology observer to update the relation.

## Checklist


taurus-forever:
That actually fixes pre-refresh-check (as always executed on /leader).
The action promote-to-primary is also works faster (due to topology observer events). Still it takes ~20 seconds.
@dragomirp merging? 

BTW, does all CI tests passed with the first try?

taurus-forever (dismissed):

marceloneppel (dismissed):

carlcsaposs-canonical (approved):

marceloneppel (approved):

PR #1392, opened 2026-01-21
[MISC] Wait for TLS enablement before trying to restart

dragomirp:
Should fix the leader unit getting stuck in `failed to restart PostgreSQL` e.g. https://github.com/canonical/pgbouncer-operator/actions/runs/21205354687/job/61001472414?pr=609#step:11:14

Reducing the amount of cluster call retries can lead to the leader unit trying to restart before TLS is enabled/disabled on the Patroni REST API. Adding call to `/health` endpoint to make sure that the TLS changes have been applied before starting the TLS rolling restart.

## Checklist


carlcsaposs-canonical (approved):

taurus-forever (approved):
This is a nice catch! Thank you!

PR #1420, opened 2026-02-04
[DPE-8932] Expose PG stats configuration (16/edge)

dragomirp:
Expose config parameters for pg_stat_statements.

Specification is DA249.

## Checklist


src/charm.py
```diff
@@ -2594,6 +2594,18 @@ def _build_postgresql_parameters(self) -> dict[str, str] | None:
             logger.debug(f"pg_parameters set to worker_configs = {pg_parameters}")
         pg_parameters.pop("maximum_lag_on_failover", None)
 
+        # TODO shouldn't be needed with a prefix
```
taurus-forever: Please pass this into spec to Mohamed to choose, 
there are no good match so far, the best we have already are: durability- or optimizer-
but maybe new performance- will be his choise.

marceloneppel (approved):

carlcsaposs-canonical (approved):

PR #1507, opened 2026-03-05
refactor: use jubilant for main integration tests

imanenami:
## Issue

Migration of tests to `jubilant`. All test files in the `tests/integration` root folder are migrated. HA & new_relations tests are skipped for now, to see how comfortable we are with this approach...

## Solution

Experimented with an Adapter-based approach. Quick and easy. Each test file took 5-10 min. to migrate once the adapters module was mature. 

### Findings:

- The `model.block_until` method of `python-libjuju` is hard to adapt, the current implementation tries to leverage `jubilant.Juju.wait`, but comes with a performance penalty.

## Checklist


taurus-forever (commented):
Wow, this is massive, but promising.
LGTM from my side, but some questions first.
The decision here is on @marceloneppel

tests/integration/jubilant_helpers.py
```diff
@@ -428,3 +451,1265 @@ def check_for_fix_log_message(juju: jubilant.Juju, unit_name: str) -> bool:
         "This may indicate the code path was not triggered or permissions were already correct."
     )
     return False
+
+
+#################################################
+#                                               #
+#              Adapted Helpers                  #
+#                                               #
+#################################################
```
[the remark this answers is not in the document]
imanenami: These are helpers which had not been ported to `jubilant_helpers`. Ideally, I think we should not extend the old helpers module, and get rid of that when all tests are migrated...

marceloneppel (approved):
LGTM! Thanks for the awesome work on this migration — it will help us a lot. The adapter approach works well for the tests migrated here, and the `block_until` fix was the right call.

One thing to be aware of: `ModelAdapter.relations` (adapters.py:505) is missing a `return` statement — it calls `self.get_relations(self.units.values())` but discards the result, so the property always returns `None`. No migrated test hits it yet, but `has_relation_exited` / `wait_for_relation_removed_between` will crash with `TypeError` when they're called.

marceloneppel (dismissed):
LGTM! Thanks for the awesome work on this migration — it will help us a lot. The adapter approach works well for the tests migrated here, and the `block_until` fix was the right call.

One thing to be aware of: `ModelAdapter.relations` (adapters.py:505) is missing a `return` statement — it calls `self.get_relations(self.units.values())` but discards the result, so the property always returns `None`. No migrated test hits it yet, but `has_relation_exited` / `wait_for_relation_removed_between` will crash with `TypeError` when they're called.

marceloneppel (dismissed):

dragomirp (dismissed):
LGTM. Great work!

tests/integration/jubilant_helpers.py
```diff
@@ -428,3 +451,1265 @@ def check_for_fix_log_message(juju: jubilant.Juju, unit_name: str) -> bool:
...
+
+def get_cluster_roles(
+    juju: JujuFixture, unit_name: str, use_ip_from_inside: bool = False
+) -> dict[str, str | list[str] | None]:
+    """Returns whether the unit a replica in the cluster."""
+    unit_ip = (
+        get_ip_from_inside_the_unit(juju, unit_name)
+        if use_ip_from_inside
+        else get_unit_ip(juju, unit_name)
+    )
```
dimaqq: `use_ip_from_inside` doesn't appear to be used, or at least I can't see any match in this PR.

imanenami:
@marceloneppel @dragomirp 

Sorry for making the reviews stale, I just had to rebase and sign the commits, and did some final, small tidy ups in this commit: https://github.com/canonical/postgresql-operator/pull/1507/commits/4193ae9cd4a92a070805842e195886f45a08faf8

I'd also like to thank all the reviewers who provided valuable input. 

No need to mention that the decision to merge this or not is completely on the repo's maintainers, specially @marceloneppel. I have plans to make the migration full and remove `pytest-operator` dependency altogether in a follow-up PR (The same way I did in this [Kafka PR](https://github.com/canonical/kafka-operator/pull/449)) so I'd highly appreciate if we keep the branch and not delete it :heart:

marceloneppel (dismissed):
Thanks for the work on this, @imanenami! I've re-reviewed after the rebase, and the changes look good. Approving. Looking forward to the follow-up PR to go fully Jubilant-native, like the Kafka one! If the branch gets auto-deleted after the merge, we can restore it.

dragomirp (dismissed):

taurus-forever (dismissed):
Let's give it a try as 1) it works 2) build more experience and confidence in Jubilant 3) can help us merge PG14+PG16 tests.

We are open for AI tests rewrite to native Jubilant. @tonyandrewmeyer can you help us make it success story?

marceloneppel (approved):

tests/integration/jubilant_helpers.py
```diff
@@ -430,3 +453,1255 @@ def check_for_fix_log_message(juju: jubilant.Juju, unit_name: str) -> bool:
...
+
+
+#################################################
+#                                               #
+#              Adapted Helpers                  #
+#                                               #
+#################################################
+
+
+CHARM_BASE = "ubuntu@22.04"
```
taurus-forever, line 465: should we use Nobel by default for 16/edge?

taurus-forever (approved):
LGTM. Let's give it a try! Thank you Imam!

tests/integration/jubilant_helpers.py
dragomirp: We can most probably ignore the base in cases the charm isn't multibase.

dragomirp (approved):

PR #1579, opened 2026-04-03
Remove on push check_docs.yaml

carlcsaposs-canonical:
16/edge was missing, but doesn't make sense to run on push in addition to PRs—it will only impact things on PRs. Still available as workflow_dispatch if needed


a-velasco (dismissed):

.github/workflows/check_docs.yaml
taurus-forever: @a-velasco should this workflow be completely removed as no longer necessary?

taurus-forever (dismissed):

dragomirp (dismissed):

marceloneppel (dismissed):

.github/workflows/automatic-doc-checks.yml
a-velasco: This workflow is for RTD-hosted docs, we need it! (unrelated to Discourse sync)

dragomirp (approved):

marceloneppel (approved):

PR #1587, opened 2026-04-06
[DPE-10370] fix(backups): reject standby backup actions and harden async replication

marceloneppel:
## Issue

In async replication setups, backup actions on standby clusters should fail with clear guidance to run on the primary cluster. Today those actions can still run and fail later with less actionable errors.

Also, `create-replication` could fail when the relation exists but remote units have not published their addresses yet, and relation-changed re-emission could run before any related unit exists.

## Solution

- Add standby-cluster detection in `PostgreSQLBackups` and fail early for `create-backup`, `list-backups`, and `restore` with explicit action-specific messages.
- Harden async replication relation handling by failing early when the relation has no remote units yet, and by guarding relation-changed re-emission when there are no related units.
- Reword the async-replication action failures from "pod addresses" to "unit addresses" to match VM terminology (user-facing message change).
- Add a jubilant-based Ceph-backed async-replication backup integration test under `tests/integration/high_availability/`, mirroring the jubilant implementation: add the `jubilant` dependency, a `juju` conftest fixture, and a `high_availability_helpers_new` helper module.
- The test verifies a backup succeeds on the primary cluster and is rejected on the standby with the expected message. It configures the s3-integrator at deploy time and waits for it to reach blocked status before syncing credentials (avoids a provisioning race), runs `create-backup` on the primary unit (the charm requires TLS for replica backups), and gives the second model its own `Juju` instance so teardown doesn't leak the temp model.
- Update the microceph fixture to use a non-loopback host IP for TLS SAN/certificate usage (this allows local testing in more environments).

## Checklist

Port of https://github.com/canonical/postgresql-operator/pull/1602 and https://github.com/canonical/postgresql-operator/pull/1635.


taurus-forever:
@marceloneppel it looks like the only failing test here is lack of Juju 2.9 support in testcase:
```
jubilant._juju.CLIError: Command '['juju', 'deploy', '--model', 'testing', '/root/snap/juju/common/tmp26du3zd6/_temp.charm', 'db1', '--base', 'ubuntu@22.04', '--config', 'profile=testing', '--constraints', 'arch=amd64', '--num-units', '3']' returned non-zero exit status 2.
Stderr:
ERROR option provided but not defined: --base
```

--base is missing for Juju 2.9 --series should be used there.... otherwise we should run this test for Juju 3.6+ only.
WDYT?

marceloneppel:
> @marceloneppel it looks like the only failing test here is lack of Juju 2.9 support in testcase:
> 
> ```
> jubilant._juju.CLIError: Command '['juju', 'deploy', '--model', 'testing', '/root/snap/juju/common/tmp26du3zd6/_temp.charm', 'db1', '--base', 'ubuntu@22.04', '--config', 'profile=testing', '--constraints', 'arch=amd64', '--num-units', '3']' returned non-zero exit status 2.
> Stderr:
> ERROR option provided but not defined: --base
> ```
> 
> --base is missing for Juju 2.9 --series should be used there.... otherwise we should run this test for Juju 3.6+ only. WDYT?

You're right, @taurus-forever. I fixed that on https://github.com/canonical/postgresql-operator/pull/1587/changes/7bf37f43bbb701360d68805963c07ca9767869b6, as the async replication feature is 3.4.2+ only.

taurus-forever (dismissed):
Nice, thank you!

taurus-forever (approved):
Thank you!

dragomirp (approved):

Issue #1329, opened 2025-12-03
S3 backups fail on the non-primary site if cross-site replication is enabled with the error message `Stanza was not initialised`


przemeklal:
## Steps to reproduce

1. Set up cross-site replication in postgresql.
2. Set up s3-integrator charm.
3. `juju run -m pgsddev-swarm-a postgresql/0 create-backup` fails on all units in the non-primary postgresql model with a confusing message
```
$ juju run postgresql/0 create-backup
Running operation 71 with 1 task
  - task 72 on unit-postgresql-0

Waiting for task 72...
Action id 72 failed: Stanza was not initialised

$ juju run postgresql/1 create-backup
Running operation 73 with 1 task
  - task 74 on unit-postgresql-1

Waiting for task 74...
Action id 74 failed: Stanza was not initialised

```

Backups work as expected on the primary site.

## Expected behavior

Either the backups should work or the error message should be clearer if it's not supported.
Also, this results in a number of COS alerts about `PgBackRestBackupTooOld` since the backups don't work.

## Actual behavior
Explained above `Action id xx failed: Stanza was not initialised` on all non-primary-site units.

## Versions

Operating system: Ubuntu 24.04.3 LTS

Juju CLI: 3.5.7-genericlinux-amd64

Juju agent: 3.6.7

Charm revision: 16/stable      952

LXD: 5.21.4 LTS

## Log output

Juju debug log: 
```
unit-postgresql-0: 15:25:25 INFO unit.postgresql/0.juju-log backups:A full backup has been requested on unit
unit-postgresql-0: 15:25:25 ERROR unit.postgresql/0.juju-log backups:Backup failed: Stanza was not initialised
```

PR #1602, opened 2026-04-08
[DPE-9010] fix(backups): reject standby backup actions and harden async replication

marceloneppel:
## Issue

In async replication setups, backup actions on standby clusters should fail with clear guidance to run on the primary cluster. Today those actions can still run and fail later with less actionable errors.

Also, `create-replication` could hit an uncaught `StopIteration` when the relation exists but no remote units have published addresses yet.

## Solution

- Add standby-cluster detection in `PostgreSQLBackups` and fail early for `create-backup`, `list-backups`, and `restore` with explicit action-specific messages.
- Harden async replication relation handling by failing early when the relation is missing or has no remote units yet, and by guarding relation-changed re-emission when there are no related units.
- Add a Juju3/Ceph-backed integration test that verifies backup works on the primary cluster and is rejected on the standby cluster with the expected message.
- Update the microceph fixture to use a non-loopback host IP for TLS SAN/certificate usage (this allows local testing in more environments).

## Checklist

Fixes https://github.com/canonical/postgresql-operator/issues/1329.


taurus-forever (approved):
THANK YOU! ❤️

dragomirp (approved):

Issue #1336, opened 2025-12-05
initdb: error: directory "/var/snap/charmed-postgresql/common/var/lib/postgresql" exists but is not empty


alanbach:
## Steps to reproduce

1. Deploy a Charmed PostgreSQL 16 cluster that uses a second disk as the data storage. A sample bundle has been provided below, please customize this as per your environment configuration:
```
    default-base: ubuntu@24.04/stable
    applications:
      postgresql:
        charm: postgresql
        channel: 16/stable
        revision: 952
        num_units: 3
        to:
        - "0"
        - "1"
        - "2"
        options:
          profile: production
        storage:
          data: nvme,1,150G
    machines:
      "0":
        constraints: arch=amd64 instance-type=c1.large zones=az1 root-disk-source=volume root-disk=50000
      "1":
        constraints: arch=amd64 instance-type=c1.large zones=az1 root-disk-source=volume root-disk=50000
      "2":
        constraints: arch=amd64 instance-type=c1.large zones=az1 root-disk-source=volume root-disk=50000
```

## Expected behavior
A three node cluster is expected to form and come online.

## Actual behavior
Cluster never properly forms because nodes fail to initialize PostgreSQL due to a `lost+found` directory being present in the data storage disk (default `ext4` behavior):
```
initdb: error: directory "/var/snap/charmed-postgresql/common/var/lib/postgresql" exists but is not empty
initdb: detail: It contains a lost+found directory, perhaps due to it being a mount point.
initdb: hint: Using a mount point directly as the data directory is not recommended.
Create a subdirectory under the mount point.

The files belonging to this database system will be owned by user "_daemon_".
This user must also own the server process.
The database cluster will be initialized with locale "C.UTF-8".
The default text search configuration will be set to "english".
Data page checksums are enabled.
initdb: error: directory "/var/snap/charmed-postgresql/common/var/lib/postgresql" exists but is not empty
initdb: detail: It contains a lost+found directory, perhaps due to it being a mount point.
initdb: hint: Using a mount point directly as the data directory is not recommended.
Create a subdirectory under the mount point.
```

## Versions

Operating system: `Ubuntu 24.04 LTS (Noble)`

Juju CLI: `3.6.12`

Juju agent: `3.6.12`

Charm revision: `952` on `16/stable` channel

LXD: `Not Applicable`

## Workaround

Use a bash script tied to a `systemd` timer to monitor the data disk and delete `lost+found` to allow the cluster to form properly:

```
[...]

PR #1612, opened 2026-04-13
[DPE-9049] feat: implement automatic cleanup of lost+found storage directories

marceloneppel:
## Issue

When deploying Charmed PostgreSQL with a dedicated storage volume (e.g., NVMe drives formatted with ext4), the filesystem automatically creates a `lost+found` directory at the root of the mount point. `initdb` treats the storage directory as non-empty and refuses to initialise the database cluster, resulting in an error such as:

```
initdb: error: directory "/var/snap/charmed-postgresql/common/var/lib/postgresql" exists but is not empty
initdb: detail: It contains a lost+found directory, perhaps due to it being a mount point.
initdb: hint: Using a mount point directly as the data directory is not recommended.
```

This prevents the PostgreSQL cluster from ever forming without manual intervention.

## Solution

The charm now automatically removes any `lost+found` directory found at the root of each storage path (data, archive, logs, and temp) during the `install` and `start` events, eliminating the need for external workarounds.

Tested with ext4 volumes in a Testflinger machine (it's needed to use LXD VMs instead of containers for it to work).

## Checklist

Fixes https://github.com/canonical/postgresql-operator/issues/1336.


dragomirp (approved):

src/charm.py
```diff
@@ -3044,6 +3048,17 @@ def is_restart_pending(self) -> bool:
...
+        """Remove the lost+found directory from the root of each storage if it exists."""
+        for storage_path in STORAGE_PATHS:
+            lost_and_found_path = Path(storage_path) / "lost+found"
+            if lost_and_found_path.is_dir():
+                logger.info(f"Removing {lost_and_found_path}")
+                try:
+                    shutil.rmtree(lost_and_found_path)
+                except OSError:
+                    logger.exception(f"Failed to remove {lost_and_found_path}")
+
```
taurus-forever, line 3061: For the history, we cannot rename/move even the PostgreSQL does ccare about dot_files and lost+found here: https://github.com/postgres/postgres/blob/66ad764c8d517f59577d41ac3dad786729c9e10e/src/port/pgcheckdir.c#L60

it still aborts basebackup execution in https://github.com/postgres/postgres/blob/66ad764c8d517f59577d41ac3dad786729c9e10e/src/bin/pg_basebackup/pg_basebackup.c#L771-L772

This is a new Juju volume which obviously has empty lost+found folder, so we are cleaning it to have deployment fixed. The folder will be re-created automatically.

taurus-forever (approved):

delgod (changes requested):
I think we can be smarter here, clean up the directory only if initdb is run nessesary.
The problem right now is that if we have a healthy PostgreSQL and the kernel crashes, and the machine is restarted, it can get a valid `lost+found` folder needed for recovery.

marceloneppel:
Closing this PR as we might not have this approach signed.

Issue #922, opened 2025-05-29
Postgresql Cluster failed to start on Dual stack network if the IPv4 subnet is from a private address space (IPv6 is then prefered)


mathmarchand:
## Steps to reproduce

1. Deploy postgresql on 14/stable on 3 machines with dual stack (IPv4 and IPv6)
2. Cluster failt to start.

## Expected behavior
Cluster start properly.

## Actual behavior
The cluster fail to start because the pg_hba.conf created by partroni has only 0.0.0.0/0 in the entries and nothing for IPv6. The error in the patroni.log:

```
2025-05-29 20:59:35 UTC [91991]: ERROR: post_bootstrap 
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/patroni/postgresql/bootstrap.py", line 384, in post_bootstrap
    self.create_or_update_role(superuser['username'], superuser['password'], ['SUPERUSER'])
  File "/usr/lib/python3/dist-packages/patroni/postgresql/bootstrap.py", line 366, in create_or_update_role
    END;$$""".format(quote_literal(name), quote_ident(name, self._postgresql.connection()), ' '.join(options))
  File "/usr/lib/python3/dist-packages/patroni/postgresql/__init__.py", line 335, in connection
    return self._connection.get()
  File "/usr/lib/python3/dist-packages/patroni/postgresql/connection.py", line 29, in get
    self._connection = psycopg.connect(**self._conn_kwargs)
  File "/usr/lib/python3/dist-packages/patroni/psycopg.py", line 103, in connect
    ret = _connect(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 122, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: connection to server at "XXXX:120:1::2f", port 5432 failed: FATAL:  no pg_hba.conf entry for host "XXXX:120:1::2f", user "operator", database "postgres", no encryption
```

## Versions

Operating system: 
Ubuntu 22.04.5 LTS

Juju CLI: 
3.6.5

Juju agent: 

Charm revision: 553

LXD: 
Not installed

## Log output

Juju debug log:

mathmarchand:
I believe that if the pg_hba.conf would be created with this additional rules:
`host all all ::/0 md5  `
It would start correctly.

taurus-forever:
Dear @mathmarchand 

[...]

PR #1648, opened 2026-04-24
[DPE-7316] Additive changes for stereo mode

dragomirp:
Factor out some of the changes for stereo mode unified charm to reduce PR size:

* Update snaps
* Factor out async calls for reuse
* Add RaftController class
* Add templates for controller service and config
* Pull systemd charmlib dep


src/relations/tls.py
```diff
@@ -217,7 +217,7 @@ def get_peer_ca_bundle(self) -> str:
         operator_ca = str(certs[0].ca) if certs else ""
         old_operator_ca = self.charm.get_secret(UNIT_SCOPE, "old-ca") or ""
         internal_ca = self.charm.get_secret(APP_SCOPE, "internal-ca") or ""
-        return "\n".join((operator_ca, old_operator_ca, internal_ca))
+        return "\n".join((operator_ca, old_operator_ca, internal_ca)).strip()
```
dragomirp, line 220: Will be passed to the watcher, so it's able to call the Patroni REST API with verify flag.

src/cluster.py
```diff
@@ -249,9 +246,28 @@ def cached_cluster_status(self):
 
     def cluster_status(self, alternative_endpoints: list | None = None) -> list[ClusterMember]:
         """Query the cluster status."""
+        if not self._patroni_async_auth:
+            raise RetryError(
+                last_attempt=Future.construct(1, Exception("Unable to reach any units"), True)
+            )
```
dragomirp, line 252: To maintain the current behaviour.

src/utils.py
```diff
@@ -78,3 +86,46 @@ def _change_owner(path: str) -> None:
...
+            await wait(tasks)
+            return result
+
+
+def parallel_patroni_get_request(
+    uri: str,
+    endpoints: list[str],
+    cafile: str,
+    auth: BasicAuth | None = None,
+    verify: bool = True,
```
dragomirp, line 128: Decoupled from the attributes of the Patroni helper class.

src/utils.py
dragomirp, line 1: TODO can also be reused by the observer script.

templates/watcher.service.j2
dragomirp, line 1: Templated unit for the RAFT observer.

templates/watcher.yml.j2
dragomirp, line 1: RAFT observer config.

templates/watcher.yml.j2
```diff
@@ -0,0 +1,18 @@
...
+# please refer to https://patroni.readthedocs.io/en/latest/SETTINGS.html.
+
+raft:
+  {% if partner_addrs -%}
+  partner_addrs:
+  {% endif -%}
+  {% for partner_addr in partner_addrs -%}
+  - {{ partner_addr }}:2222
+  {% endfor %}
+  self_addr: '{{ self_addr }}:{{ self_port }}'
```
dragomirp, line 16: Hard dependency for the observer. We will need to start/stop based on the number of partners.

src/raft_controller.py
```diff
@@ -0,0 +1,414 @@
...
+    """Type definition for the cluster status mapping."""
+
+    running: bool
+    connected: bool
+    has_quorum: bool
+    leader: str | None
+    members: list[str]
+
+
+def install_service() -> bool:
```
dragomirp, line 77: Out of the class since it's not dependent on existing relations.

src/raft_controller.py
```diff
@@ -0,0 +1,414 @@
...
+        create_directory(self.data_dir, 0o700)
+        create_directory(f"{self.data_dir}/raft", 0o700)
+
+        if not self_addr or not password:
+            logger.warning("Cannot install service: not configured")
+            return False
+
+        # Validate addresses to prevent injection into the systemd unit file
+        try:
+            IPv4Address(self_addr)
```
dragomirp, line 162: Should we only check for IPv4 addresses here?

src/raft_controller.py
dragomirp, line 1: Not really called by anything at the moment so it should be safe to merge.

src/raft_controller.py
```diff
@@ -0,0 +1,414 @@
...
+    rendered = template.render(config_file=RAFT_BASE_DIR)
+    render_file(SERVICE_FILE, rendered, 0o644, change_owner=False)
+
+    # Reload systemd to pick up the new service
+    try:
+        daemon_reload()
+        logger.info(f"Installed systemd service {SERVICE_FILE}")
+    except SystemdError as e:
+        logger.error(f"Failed to reload systemd: {e}")
+        return False
```
dragomirp, line 95: Maybe we should just raise the exception here?

src/raft_controller.py
```diff
@@ -0,0 +1,414 @@
...
+
+        Returns:
+            True if configuration changed, False if unchanged.
+        """
+        if not partner_addrs:
+            partner_addrs = []
+
+        # Ensure data directory exists
+        create_directory(self.data_dir, 0o700)
+        create_directory(f"{self.data_dir}/raft", 0o700)
```
dragomirp, line 154: Make sure that necessary directories exist. Had no luck with `600` so far, `_daemon_` user didn't get access.

templates/watcher.service.j2
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=PostgreSQL Watcher Raft Service (%i)
+After=network.target
+Wants=network.target
+
+[Service]
+Type=simple
+# charmed-postgresql.patroni-raft-controller app lacks network interfaces
+# in the snap profile, so run the controller under the patroni app profile.
+ExecStart=/snap/bin/charmed-postgresql.patroni-raft-controller {{ config_file }}/%i/patroni-raft.yaml
```
dragomirp, line 10: Ran as root, snap's setpriv shim will deescalate to `_daemon_`.

src/raft_controller.py
```diff
@@ -0,0 +1,414 @@
...
+        """
+        try:
+            service_restart(self.service_name)
+            logger.info(f"Restarted Raft controller service {self.service_name}")
+            return True
+        except SystemdError as e:
+            logger.error(f"Failed to restart Raft controller: {e}")
+            return False
+
+    def get_status(self, self_port: int, password: str | None) -> ClusterStatus:
```
dragomirp, line 262: Will be used for health check part of the spec.

src/raft_controller.py
marceloneppel: I think that's a good approach, so the user can call `juju resolve` to retry it.

src/cluster.py
```diff
@@ -249,9 +246,28 @@ def cached_cluster_status(self):
...
+
+        # TODO we don't know the other cluster's ca
+        verify = not bool(alternative_endpoints)
```
dragomirp, line 255: Async rel doesn't share CAs. Existing behaviour.

src/raft_controller.py
marceloneppel: For now, yes, as we still need to properly implement IPV6 (https://github.com/canonical/postgresql-operator/issues/922#issuecomment-2931255500).

marceloneppel (approved):
LGTM!

src/cluster.py
```diff
@@ -249,9 +246,28 @@ def cached_cluster_status(self):
 
     def cluster_status(self, alternative_endpoints: list | None = None) -> list[ClusterMember]:
         """Query the cluster status."""
+        if not self._patroni_async_auth:
+            raise RetryError(
+                last_attempt=Future.construct(1, Exception("Unable to reach any units"), True)
+            )
+
+        # TODO we don't know the other cluster's ca
```
taurus-forever, line 254: Time to introduce receive-ca-cert for cluster<>cluster (if they use self-signed CAs)... otherwise we have to mandate the same/trusted CAs on the both sides.

src/raft_controller.py
taurus-forever: LGTM, the common problem from my past: this folder will be created on root drive (before the pgdata mount point finished)... would be nice to ensure we are creting it on Juju data mount point... somehow...

src/raft_controller.py
taurus-forever: Well... for clarity... before this commit, PostgreSQL works well on IPv6 AFAIK.
It doesn't work on mixed IPv4+IPv6 due to Juju.

Can we avoid locking us to IPv4 intentionally?

templates/watcher.service.j2
taurus-forever: Q: will it be systemd disabled (on boot)?

taurus-forever (approved):
TL;DR: LGTM.
Q: about IPv4 only... can it be improved?

src/raft_controller.py
dragomirp: I will switch to [ip_address](https://docs.python.org/3.12/library/ipaddress.html#ipaddress.ip_address) in the follow up PRs, it should be able to parse both IPv4 and IPv6.

src/cluster.py
dragomirp: We'll need to also add CA sharing for self signed peer cert. Out of scope for stereo mode.

templates/watcher.service.j2
dragomirp: We will need to enable individual @ services when we setup the relations.

PR #1649, opened 2026-04-24
[DPE-9049] feat(storage): migrate all data mounts to versioned 16/main subdirectories

marceloneppel:
## Issue

PostgreSQL warns against using a mount point directly as a data directory — for example, a `lost+found` directory at the mount root causes `initdb` to fail:

```
initdb: error: directory "…/var/lib/postgresql" exists but is not empty
initdb: detail: It contains a lost+found directory, perhaps due to it being a mount point.
initdb: hint: Using a mount point directly as the data directory is not recommended.
         Create a subdirectory under the mount point.
```

All four Juju-managed storage mounts need to move from the mount root to a versioned `16/main` subdirectory:

| Storage | Mount point | Old directory | New directory |
|---------|-------------|---------------|---------------|
| data | `{SNAP_COMMON}/var/lib/postgresql` | `${MOUNT}` | `${MOUNT}/16/main` |
| archive | `{SNAP_COMMON}/data/archive` | `${MOUNT}` | `${MOUNT}/16/main` |
| logs | `{SNAP_COMMON}/data/logs` | `${MOUNT}` | `${MOUNT}/16/main` |
| temp | `{SNAP_COMMON}/data/temp` | `${MOUNT}` | `${MOUNT}/16/main` |

The bidirectional data migration (file moves, `pg_wal` symlink repair, permissions) is handled by the snap's pre-refresh and post-refresh hooks, introduced in [canonical/charmed-postgresql-snap#277](https://github.com/canonical/charmed-postgresql-snap/pull/277). This PR provides the charm-side counterpart: path constants, Patroni configuration, temp tablespace catalog migration, and snap revision pinning.

### Versioned path constants (`src/constants.py`)

Add `DATA_DIR_SUBFOLDER = "16/main"` and derive four new constants — `POSTGRESQL_DATA_DIR`, `ARCHIVE_DATA_DIR`, `LOGS_DATA_DIR`, `TEMP_DATA_DIR` — each appending `16/main` to its respective storage root. All internal path references across the charm (`src/charm.py`, `src/backups.py`, `src/cluster.py`, `src/relations/async_replication.py`) and tests are updated to use the versioned paths.

### Patroni configuration (`templates/patroni.yml.j2`, `src/cluster.py`)

Replace the hardcoded WAL directory path in `patroni.yml.j2` with a `{{ wal_dir }}` template variable, set to `LOGS_DATA_DIR` during rendering.

### Temp tablespace directory (`src/charm.py` — `_ensure_storage_layout`)

The temp storage may live on a tmpfs mount that is wiped on reboot. `_ensure_storage_layout()` recreates `TEMP_DATA_DIR` unconditionally and chowns it (plus the `16/` parent) to `_daemon_`, which is needed for the snap's pre-refresh hook to run `DROP/CREATE TABLESPACE` during rollback.

Called from `_on_start`, `_post_snap_refresh`, and `refresh_snap` (before the snap refresh begins).

### Temp tablespace catalog migration (`src/charm.py` — `_migrate_temp_tablespace_location`)

On the primary, after a successful snap refresh, `_migrate_temp_tablespace_location()` performs a one-shot `DROP TABLESPACE temp` / `CREATE TABLESPACE temp LOCATION '<TEMP_DATA_DIR>'` to move the PostgreSQL catalog entry from the old storage root to the versioned path. Uses `autocommit=True` since DDL tablespace statements cannot run inside a transaction block.

Safety guards:
- Skips on replicas (WAL replay handles it).
- Skips when `primary_endpoint` is not yet available.
- Skips when cross-cluster async replication is active (the standby may not have the versioned directory yet, so WAL replay of the `DROP/CREATE` would crash it).
- Skips when the tablespace is already at the versioned path or at an unexpected location.
- Clears stale `PG_*` version directories from the target path before `CREATE TABLESPACE` to prevent `ObjectInUse` errors.

The reverse catalog migration (rollback) is handled by the snap's pre-refresh hook ([canonical/charmed-postgresql-snap#277](https://github.com/canonical/charmed-postgresql-snap/pull/277)).

### Pre-refresh temp tablespace object check (`src/charm.py` — `_check_temp_tablespace_objects`)

`_check_temp_tablespace_objects()` is called during `run_pre_refresh_checks_after_1_unit_refreshed` (i.e. from both `pre-refresh-check` and `force-refresh-start` actions). It queries `pg_class` for any objects residing in the `temp` tablespace and raises `PrecheckFailed` if any are found — persistent objects in the temp tablespace would block the snap's `migrate-data.sh` from moving the temp directory.

### Snap revision and workload version pinning (`refresh_versions.toml`)

Pin snap revisions to 346 (amd64) and 347 (arm64), which include the bidirectional `migrate-data` daemon and the `SNAP_CURRENT` → `SNAP_DATA` variable fix. Bump the workload version from 16.13 to 16.14.

### Test updates

- **Unit tests** (`tests/unit/test_charm.py`): Add dedicated tests for `_ensure_storage_layout`, all branches of `_migrate_temp_tablespace_location` (not primary, no endpoint, already migrated, tablespace missing, unexpected location, DB error, successful migration), and `_reconfigure_cluster` gating in `_on_update_status`. Split the monolithic `test_on_start` into focused tests (`test_on_start_no_password`, `test_on_start_bootstrap_failure`, `test_on_start_create_user_error`, `test_on_start_success`).
- **Integration tests**: Update hardcoded paths to versioned layout, add `arm64` constraint support, increase refresh/upgrade timeouts (5 min → 10–40 min), and soften the persistent temp storage permission-fix assertion to a warning (permissions are already correct for units initialized with the new layout).

> **Note:** The timeout increases need investigation to identify which specific change required them. A follow-up PR should restore the original timeouts once the root cause is understood.

## Testing

To test and validate the PR changes, use the `16/edge/neppel` channel:

```bash
# Deploy from stable
juju deploy postgresql --channel 16/stable --config profile=testing

# Upgrade to the versioned storage layout
juju refresh postgresql --channel 16/edge/neppel
```

Charm revisions on `16/edge/neppel`:
- **amd64**: revision 1135
- **arm64**: revision 1136

## Follow-up PRs regarding automated testing:

* https://github.com/canonical/postgresql-operator/pull/1699
* https://github.com/canonical/postgresql-operator/pull/1703
* https://github.com/canonical/postgresql-operator/pull/1704

## Checklist


dragomirp (dismissed):

taurus-forever (dismissed):

taurus-forever (dismissed):

tests/integration/high_availability/test_upgrade_skip_pre_upgrade_check.py
```diff
@@ -30,13 +31,15 @@ def test_deploy_stable(juju: Juju) -> None:
         base="ubuntu@24.04",
         channel="16/stable",
         config={"profile": "testing"},
+        constraints={"arch": "arm64"} if platform.machine() == "aarch64" else None,
```
taurus-forever, line 34: @marceloneppel BTW, why do we need this here?

marceloneppel:
https://github.com/canonical/charmed-postgresql-snap/pull/281 will fix the failing CI tests on this PR.

tests/integration/high_availability/test_upgrade_skip_pre_upgrade_check.py
marceloneppel: No. It's a leftover from my local tests. I'll create a follow-up PR to avoid needing to retrigger all the integration tests.

dragomirp (approved):

tests/integration/high_availability/test_upgrade_skip_pre_upgrade_check.py
marceloneppel: Next PR in queue: https://github.com/canonical/postgresql-operator/pull/1727

taurus-forever (approved):
Thank you, @marceloneppel !

PR #1681, opened 2026-05-10
[MISC] Faster standby cluster host change

dragomirp:
Try to speed up the standby cluster host configuration.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1028,6 +1028,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         # This ensures pg-endpoints stay current when unit IPs change
         if self.unit.is_leader():
             self.watcher_offer.update_endpoints()
+            self.async_replication.update_async_replication_data()
 
         self._update_new_unit_status()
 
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -1028,6 +1028,7 @@ def _on_peer_relation_changed(self, event: HookEvent):
         # This ensures pg-endpoints stay current when unit IPs change
         if self.unit.is_leader():
             self.watcher_offer.update_endpoints()
+            self.async_replication.update_async_replication_data()
```
dragomirp, line 1031: Promote action will trigger peer change on the Juju leader unit.

src/charm.py
```diff
@@ -1530,6 +1532,7 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:  # noqa: C901
         # a failed switchover, so wait until the primary is elected.
         if self.primary_endpoint:
             self._update_relation_endpoints()
+            self.async_replication.update_async_replication_data()
```
dragomirp, line 1535: If cluster topology change event happens while the leader is down, it might not be processed, so updating the async rel when the leader is elected.

taurus-forever (approved):
let me try this PPA, it looks promising! Tnx!

taurus-forever:
> let me try this PPA, it looks promising! Tnx!

It is about 20 seconds on tests on my laptop.

marceloneppel (approved):

PR #1690, opened 2026-05-12
[MISC] Remove dead functions

dragomirp:
Remove dead functions not expected to be moved to the single kernel (for k8s/watcher compatibility).


taurus-forever (approved):
🥳 

Step 2: backport to main?

marceloneppel (approved):

PR #1700, opened 2026-05-22
[DPE-10017] Leader switchover update and faster topology cycle

dragomirp:
* Update endpoints when switching over to the Juju leader
* Make the topology observer run every 10s
* Set snap and Postgresql version in the per unit data for the observer relation

### DIFF
--- a/refresh_versions.toml
+++ b/refresh_versions.toml
@@ -6,6 +6,6 @@ name = "charmed-postgresql"
 
 [snap.revisions]
 # amd64
-x86_64 = "289"
+x86_64 = "332"
 # arm64
-aarch64 = "288"
+aarch64 = "331"
--- a/scripts/cluster_topology_observer.py
+++ b/scripts/cluster_topology_observer.py
@@ -170,7 +170,7 @@ async def main():
             )
 
         # Wait some time before checking again for a cluster topology change.
-        sleep(30)
+        sleep(10)
 
 
 if __name__ == "__main__":
--- a/src/charm.py
+++ b/src/charm.py
@@ -436,6 +436,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         self._setup_exporter()
         self.backup.start_stop_pgbackrest_service()
         self._setup_pgbackrest_exporter()
+        self.watcher_offer.update_unit_address()
 
         # Wait until the database initialise.
         self.set_unit_status(WaitingStatus("waiting for database initialisation"), refresh=refresh)
--- a/src/charm.py
+++ b/src/charm.py
@@ -1996,6 +1996,9 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             try:
                 self._patroni.switchover(self._member_name)
                 self.unit_peer_data.update({"timestamp": str(datetime.now())})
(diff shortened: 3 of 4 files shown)


scripts/cluster_topology_observer.py
```diff
@@ -170,7 +170,7 @@ async def main():
             )
 
         # Wait some time before checking again for a cluster topology change.
-        sleep(30)
+        sleep(10)
```
dragomirp, line 173: Run topology observer every 10s.

src/relations/watcher.py
```diff
@@ -414,7 +414,6 @@ def _update_relation_data(self, relation: Relation) -> None:
         relation.data[self.charm.app].update({
             "cluster-name": self.charm.cluster_name,
             "raft-secret-id": secret_id,
-            "version": self.charm._patroni.get_postgresql_version(),
```
dragomirp, line 417: Should not be used currently. Set per unit.

src/relations/watcher.py
```diff
@@ -439,27 +438,23 @@ def update_unit_address(self, relation: Relation | None = None) -> None:
...
-            changed = True
 
         unit_az = os.environ.get("JUJU_AVAILABILITY_ZONE")
         current_az = relation.data[self.charm.unit].get("unit-az")
         if unit_az and current_az != unit_az:
             relation.data[self.charm.unit]["unit-az"] = unit_az
-            changed = True
-
-        if changed:
-            logger.debug("Updated watcher relation unit data")
```
dragomirp, line 462: Log spam IMHO.

src/charm.py
```diff
@@ -436,6 +436,7 @@ def _post_snap_refresh(self, refresh: charm_refresh.Machines):
         self._setup_exporter()
         self.backup.start_stop_pgbackrest_service()
         self._setup_pgbackrest_exporter()
+        self.watcher_offer.update_unit_address()
```
dragomirp, line 439: Reset snap rev and PG version after snap refresh.

src/charm.py
```diff
@@ -1996,6 +1997,9 @@ def promote_primary_unit(self, event: ActionEvent) -> None:
             try:
                 self._patroni.switchover(self._member_name)
                 self.unit_peer_data.update({"timestamp": str(datetime.now())})
+                if self.unit.is_leader():
+                    self.postgresql_client_relation.update_endpoints()
+                    self.async_replication.update_async_replication_data()
```
dragomirp, line 2002: If the unit is leader, we can just reset the IPs and topology.

refresh_versions.toml
dragomirp, line 1: Snap bump to 16.14.

src/relations/watcher.py
taurus-forever: Risky to mix here... OK.

src/charm.py
taurus-forever: Worth to explain for us in the future: there are no peer-relation-chnged event on Juju leader when user switch from from non-leader to leader => slow promotion till the next topology observer call. We speedup promotion here when we promote to leader => we can write databag => do it instantly.

taurus-forever (approved):

marceloneppel (approved):

PR #1743, opened 2026-06-03
[DPE-9737] fix(storage): chown versioned data parent so patronictl reinit can rebuild a replica

marceloneppel:
## Issue

On the versioned-storage layout the data-dir parent (`.../postgresql/16`) can be owned by `root:root` while Patroni runs as the unprivileged `_daemon_` user. During `patronictl reinit`, Patroni can empty `16/main` (it owns the contents) but cannot remove or rename the directory itself, which needs write permission on the parent. Reinit still succeeds via basebackup, but logs "Could not remove/rename data directory" errors and `PermissionError` tracebacks, which CI flags.

The parent ends up root-owned on two paths:

- **Fresh deploy:** the charm creates it via `os.makedirs(POSTGRESQL_DATA_DIR)` in `configure_patroni_on_unit` (which runs as root); only the `16/main` leaf was chowned.
- **Upgrade/rollback:** the snap's `migrate-data.sh` creates the parent as root.

## Solution

- Chown the `16/` data parent to `_daemon_` in `configure_patroni_on_unit`, right after `os.makedirs`, so it is correctly owned on a fresh deploy.
- Also chown it (existence-guarded) in `_ensure_storage_layout`, covering the upgrade/rollback path where `migrate-data.sh` already created the parent, mirroring the temp-tablespace parent handling.
- Unit tests assert both the parent and leaf chowns in `configure_patroni_on_unit` and cover the `_ensure_storage_layout` data-parent chown.
- Add a reinit integration test that catches this: wait for Patroni to stop before deleting pg_control; confirm the replica is genuinely broken before reinit — re-registered in the cluster, not streaming, and Patroni has logged a pg_control startup failure (so it can't false-pass); require it to return to "streaming" via the Patroni REST API; and grep the `patroni.log*` files (not journalctl, which never receives these errors) for the data-directory permission-error signatures.

As a follow-up, the ownership migration could be moved to the snap's script.

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -698,7 +698,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
         return "stopped" in self.unit_peer_data
 
     def _ensure_storage_layout(self) -> None:
-        """Ensure the temp tablespace directory exists.
+        """Ensure the temp tablespace dir exists and versioned parents are _daemon_-owned.
 
         Data migration between storage roots and versioned 16/main
         subdirectories is handled by the snap hooks (pre-refresh for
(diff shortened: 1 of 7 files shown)


src/charm.py
```diff
@@ -708,17 +708,24 @@ def _ensure_storage_layout(self) -> None:
...
         temp_dir = Path(TEMP_DATA_DIR)
         temp_dir.mkdir(parents=True, exist_ok=True)
         shutil.chown(temp_dir, user=SNAP_DAEMON_USER, group=SNAP_DAEMON_USER)
         if temp_dir.parent.exists():
             shutil.chown(temp_dir.parent, user=SNAP_DAEMON_USER, group=SNAP_DAEMON_USER)
 
+        data_parent = Path(POSTGRESQL_DATA_DIR).parent
+        if data_parent.exists():
+            shutil.chown(data_parent, user=SNAP_DAEMON_USER, group=SNAP_DAEMON_USER)
+
```
taurus-forever, line 728: Are we changing here to `main` part of `16/main` only? 
Will it bite us later? Should we change it for both `16` and `main`?

taurus-forever (approved):
Approved as it looks better, but one question to be sure.

dragomirp (approved):

src/charm.py
marceloneppel: We're changing `16` in this case, which is the parent directory from `main` (the data directory). `main` is already owned by the snap daemon user.

PR #1744, opened 2026-06-04
[MISC] Add ty typechecking

dragomirp:
Add ty typechecking to 14/edge and fix violations.


tests/unit/test_async_replication.py
```diff
@@ -121,14 +120,14 @@ def test_handle_database_start():
...
-    mock_peers_data = {
+    mock_charm.all_peer_data = {
         mock_charm.unit: MagicMock(),
         mock_unit1: MagicMock(),
         mock_unit2: MagicMock(),
         mock_charm.app: MagicMock(),
     }
     mock_charm._peers = MagicMock()
-    mock_charm._peers.data = mock_peers_data
+    # mock_charm._peers.data = mock_peers_data
```
marceloneppel:
```suggestion
```

src/relations/postgresql_provider.py
```diff
@@ -135,7 +144,7 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
             )
 
             # Set the database name
-            self.database_provides.set_database(event.relation.id, database)
+            self.database_provides.set_database(event.relation.id, database or "")
```
marceloneppel: Is ` or ""` really needed with the above guard?

```python
        if database is None:
            logger.warning("Database name is not set in the relation data, skipping.")
            return
```

src/relations/db.py
```diff
@@ -422,7 +428,7 @@ def update_endpoints(self, relation: Relation = None) -> None:
                 data["version"] = postgresql_version
 
             # Set the data only in the unit databag.
-            unit_relation_databag.update(data)
+            unit_relation_databag.update({k: v for k, v in data.items() if v is not None})
```
marceloneppel: Out of curiosity: was this change needed because the primary endpoint may be `None` or something else?

src/charm.py
```diff
@@ -1766,7 +1792,7 @@ def _was_restore_successful(self) -> bool:
                 "stanza": "",
                 "s3-initialization-start": "",
                 "s3-initialization-done": "",
-                "s3-initialization-block-message": validation_message,
+                "s3-initialization-block-message": validation_message or "",
```
marceloneppel: Is this change still needed with the change in the signature from `can_use_s3_repository` to return an empty string in the validation message instead of `None`?

src/relations/async_replication.py
```diff
@@ -223,7 +227,7 @@ def _get_highest_promoted_cluster_counter_value(self) -> str:
                 continue
             for databag in [
                 async_relation.data[async_relation.app],
-                self.charm._peers.data[self.charm.app],
+                self.charm.unit_peer_data,
```
marceloneppel:
```suggestion
                self.charm.app_peer_data,
```

marceloneppel (changes requested):
Thanks for this great work, @dragomirp!

src/relations/postgresql_provider.py
dragomirp: No, cleaned up.

src/charm.py
dragomirp: Cleaned up.

src/relations/db.py
dragomirp: Yes, the type issue was primary being `None`. Cast to str, since that's what we are already doing implicitly.

taurus-forever (approved):
Nice!

marceloneppel (approved):

PR #1745, opened 2026-06-04
[DPE-10017] Wait for the primary cluster to stop on async switchover

dragomirp:
Wait for old primary cluster to stop during async replication switchover to reduce the possibility of drift between clusters.

### DIFF
--- a/src/relations/async_replication.py
+++ b/src/relations/async_replication.py
@@ -174,6 +174,15 @@ class PostgreSQLAsyncReplication(Object):
     ) -> bool:
         """Configure the primary cluster."""
         if self.charm.app == primary_cluster:
+            counter = self._get_highest_promoted_cluster_counter_value()
+            if not all(
+                event.relation.data[unit].get("stopped") == counter
+                for unit in event.relation.units
+                if unit.app != event.relation.app
+            ):
+                logger.info("Other cluster not yet down.")
+                event.defer()
+                return False
             self.charm.update_config()
             if self.is_primary_cluster() and self.charm.unit.is_leader():
                 self._update_primary_cluster_data()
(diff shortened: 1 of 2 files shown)


taurus-forever (dismissed):

src/relations/async_replication.py
```diff
@@ -505,11 +514,6 @@ def _handle_replication_change(self, event: ActionEvent) -> bool:
 
         self._update_primary_cluster_data(promoted_cluster_counter, system_identifier)
 
-        # Emit an async replication changed event for this unit (to promote this cluster before demoting the
-        # other if this one is a standby cluster, which is needed to correctly set up the async replication
-        # when performing a switchover).
-        self._re_emit_async_relation_changed_event()
```
dragomirp, line 511: We want to wait for the other cluster to stop before promoting.

src/relations/async_replication.py
```diff
@@ -572,6 +576,9 @@ def _on_async_relation_changed(self, event: RelationChangedEvent) -> None:
 
         if not self._stop_database(event):
             return
+        event.relation.data[self.charm.unit]["stopped"] = (
+            self._get_highest_promoted_cluster_counter_value()
```
dragomirp, line 580: Use the counter value to make sure rel data is updated over multiple switchovers.

carlcsaposs-canonical (dismissed):

src/relations/async_replication.py
```diff
@@ -174,6 +174,15 @@ def _configure_primary_cluster(
...
         if self.charm.app == primary_cluster:
+            counter = self._get_highest_promoted_cluster_counter_value()
+            if not all(
+                event.relation.data[unit].get("stopped") == counter
+                for unit in event.relation.units
+                if unit.app != event.relation.app
+            ):
+                logger.info("Other cluster not yet down.")
+                event.defer()
+                return False
```
marceloneppel: Shouldn't it be the following? I think `event.relation.app` is the remote app:

```suggestion
                if unit.app == event.relation.app
            ):
                logger.info("Other cluster not yet down.")
                event.defer()
                return True
```

I suspect the above return change may be needed to avoid cleaning up the primary's data folder.

marceloneppel (approved):

taurus-forever (approved):

PR #1750, opened 2026-06-06
fix(backups): configure pgBackRest stderr logging

### LINKED ISSUE #1353: Move `--log-level-stderr=warn` to pgBackRest's configuration file
Move `--log-level-stderr=warn` to pgBackRest's configuration file, as described at https://github.com/canonical/postgresql-operator/pull/1320#discussion_r2622432819.

nanookclaw:
## Summary

Adds `log-level-stderr=warn` to the rendered pgBackRest global configuration so the stderr logging level is configured once for every pgBackRest invocation that uses `pgbackrest.conf`, as requested in #1353.

The existing `--log-level-console=debug` backup argument is left unchanged because it controls console output for the backup action and is distinct from the stderr logging level being moved into the config file.

## Verification

- `python3 -m compileall -q tests/unit/test_backups.py`
- `ruff check tests/unit/test_backups.py`
- `ruff format --check tests/unit/test_backups.py`
- `git diff --check`
- Jinja render smoke check confirmed `log-level-stderr=warn` appears in the rendered config

Could not run the focused pytest locally because this container lacks the project test environment: host pytest autoloads a missing `evalcraft` plugin, plugin-autoload-disabled pytest then lacks `botocore`, and `poetry`/`tox` are not installed here.

Closes #1353

### DIFF
--- a/templates/pgbackrest.conf.j2
+++ b/templates/pgbackrest.conf.j2
@@ -2,6 +2,7 @@
 backup-standby=y
 compress-type=zst
 lock-path=/tmp
+log-level-stderr=warn
 log-path={{ log_path }}
 repo1-retention-full-type=time
 repo1-retention-full={{ retention_full }}
--- a/tests/unit/test_backups.py
+++ b/tests/unit/test_backups.py
@@ -1870,6 +1870,13 @@ def test_render_pgbackrest_conf_file(harness, tls_ca_chain_filename):
 
         # Check the template is opened read-only in the call to open.
         assert mock.call_args_list[0][0] == ("templates/pgbackrest.conf.j2",)
+        rendered_pgbackrest_conf = next(
+            call_args[0][1]
+            for call_args in _render_file.call_args_list
+            if call_args[0][0]
+            == "/var/snap/charmed-postgresql/current/etc/pgbackrest/pgbackrest.conf"
+        )
+        assert "log-level-stderr=warn" in rendered_pgbackrest_conf
 
         # Get the expected content from a file.
         with open("templates/pgbackrest.conf.j2") as file:


taurus-forever (commented):
Should this https://github.com/canonical/postgresql-operator/blob/82b1194bf8f927238a1ff5865d3d5763cdd630f9/src/constants.py#L34 be reverted (as discussed [here](https://github.com/canonical/postgresql-operator/pull/1320/changes#r2622432819)) ?

nanookclaw:
I don’t think a separate constant needs to be reverted in this PR. The command-level `PGBACKREST_LOG_LEVEL_STDERR` approach from #1320 is not present on current `main`; `src/constants.py#L34` is just `PGBACKREST_EXECUTABLE` now.

This PR applies the earlier suggestion in the config layer instead: `templates/pgbackrest.conf.j2` now sets `log-level-stderr=warn`, so pgBackRest commands using `--config=.../pgbackrest.conf` pick it up without adding a per-command flag.

The only remaining command-level logging flag I see on `main` is `_run_backup()`’s `--log-level-console=debug`, which is backup-specific and looks intentionally preserved for backup log capture. If you want this PR to also remove or change that backup-specific flag, I can adjust it, but I kept the diff scoped to the shared pgBackRest config.

carlcsaposs-canonical:
@taurus-forever perhaps we should set the default branch to `16/edge`?

taurus-forever:
> @taurus-forever perhaps we should set the default branch to `16/edge`?

It was pending stereo mode PRs release to perform massive default branch upgrade as discussed in Madrid.
It is coming ...

marceloneppel (approved):
LGTM!

I created https://github.com/canonical/postgresql-operator/issues/1767 to also work on `16/edge` and remove the constant @taurus-forever mentioned.

PR #1761, opened 2026-06-10
[DPE-10401] Raft observer

dragomirp:
Add a Raft connection observer to detect stuck raft connections and try to stabilise stereo mode tests.


templates/raft-observer.timer.j2
dragomirp, line 1: Systemd timer.

src/relations/watcher.py
```diff
@@ -110,7 +110,8 @@ def disable_watcher(self) -> None:
 
         self._relation.data[self.charm.app].update({"disable-watcher": "True"})
         try:
-            self.charm._patroni.remove_raft_member(self.watcher_raft_address)
+            if self.watcher_raft_address:
+                self.charm._patroni.remove_raft_member(self.watcher_raft_address)
```
dragomirp, line 114: For simpler typing.

src/charm.py
```diff
@@ -1616,6 +1618,38 @@ def _on_cluster_topology_change(self, _):
...
+    def _on_raft_reconnect(self, _) -> None:
+        logger.debug(f"Local raft status: {self._patroni.get_raft_status()}")
+        if (
+            not self._unit_ip
+            or not self.is_cluster_initialised
+            or self._unit_ip not in self.members_ips
+            or self.has_raft_keys()
+            or (not self.members_ips and not self.watcher_offer.watcher_raft_address)
+        ):
+            return
```
dragomirp, line 1633: Are the other conditions where we shouldn't try?

taurus-forever (approved):
WOW! This is amazing!!!!

src/charm.py
```diff
@@ -1616,6 +1618,38 @@ def _on_cluster_topology_change(self, _):
             self._set_primary_status_message()
             self.async_replication.update_async_replication_data()
 
+    def _on_raft_reconnect(self, _) -> None:
+        logger.debug(f"Local raft status: {self._patroni.get_raft_status()}")
```
taurus-forever: Maybe we should show Patroni cluster status nearby?
E.g. I see [here](https://github.com/canonical/postgresql-operator/actions/runs/27472001422/job/81210849106#step:13:8707):
```
unit-postgresql-2: 16:48:33 DEBUG unit.postgresql/2.juju-log __main__:Local raft status: {'version': '0.3.12', 'revision': 'deprecated', 'self': TCPNode('10.153.30.97:2222'), 'state': 0, 'leader': TCPNode('10.153.30.55:2222'), 'has_quorum': True, 'partner_nodes_count': 2, 'partner_node_status_server_10.153.30.55:2222': 2, 'partner_node_status_server_10.153.30.198:2222': 2, 'readonly_nodes_count': 0, 'log_len': 123, 'last_applied': 701, 'commit_idx': 701, 'raft_term': 7, 'next_node_idx_count': 0, 'match_idx_count': 0, 'leader_commit_idx': 701, 'uptime': 18, 'self_code_version': 0, 'enabled_code_version': 0}
unit-postgresql-2: 16:48:33 INFO unit.postgresql/2.juju-log __main__:Potentially stuck Raft connection detected. Re-adding Raft member.
```

Which is for me the healthy Raft cluster status (has_quorum, 3 notes total, all partner_node status `2`,... why do we re-adding a member here? Should we show a Raft leader raft cluster status here?

The last place I see Patroni cluster status is [here](https://github.com/canonical/postgresql-operator/actions/runs/27472001422/job/81210849106#step:13:8694) and it is also healthy.

taurus-forever (dismissed):

src/charm.py
taurus-forever: Traced with @dragomirp :
1) Update observer systemd to add Requires/Wants as After is not enough (raft-observers starts in parallel to Patroni and detects Raft issues -> triggering this event and remove/add raft member for no real need).
2) consider to re-check in reconnect_event handler the current Raft status and execute remove+add when necessary only, as we might have outdated alarms from raft-observer (due to async call thorough events nature).

src/charm.py
```diff
@@ -1616,6 +1619,48 @@ def _on_cluster_topology_change(self, _):
...
+            logger.debug("All raft members are active.")
+            return
+
+        logger.info("Potentially stuck Raft connection detected. Re-adding Raft member.")
+        local_addr = f"{self._unit_ip}:{RAFT_PORT}"
+        remote_addr = (
+            watcher_addr
+            if (watcher_addr := self.watcher_offer.watcher_raft_address)
+            and self.watcher_offer.is_active
+            else f"{next(member for member in self.members_ips)}:{RAFT_PORT}"
```
marceloneppel: Do we need to exclude `local_addr` from the list retrieved from `self.members_ips`?

marceloneppel (dismissed):
Amazing work!

marceloneppel (dismissed):

templates/raft-observer.service.j2
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Patroni Raft observer service
+After=snap.charmed-postgresql.patroni.service
```
dragomirp, line 3: Tried `Wants` instead of `After` but it starts the service immediately and gets the charm stuck. We should revisit in a follow up.

taurus-forever: I believe you need both Requires and After

dragomirp: Both [Wants](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Wants=):
```
Units listed in this option will be started if the configuring unit is.
```
and [Requires](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Requires=):
```
If this unit gets activated, the units listed will be activated as well.
```
state they will start the dependency unit. Checking for running snap services from the observer script should be simpler and safer than trying to do it via systemd dependency resolution.

taurus-forever: We either should disable systemd raft-observer service or configure systemd startup dependencies/order.
AFAIK, 
* raft-observer should always start on server boot => systemd enabled.
* raft-observer should always start After patroni 
* raft-observer should NOT start if patroni failed to start (to avoid unnecessary events from observer) => Requires

The message `If this unit gets activated, the units listed will be activated as well.` only means, calling `systemctl raft-observer start` will also start `patroni` as a dependency for the service => OK.

Do I miss something here?

templates/raft-observer.timer.j2
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Patroni Raft observer timer
+
+[Timer]
+OnActiveSec=10s
+OnUnitInactiveSec=10s
```
dragomirp, line 6: Wait for after deactivation, instead of after activation.

scripts/raft_observer.py
```diff
@@ -0,0 +1,81 @@
...
+            print(f"Potentially stuck connection with {addr}")
+            dispatch("raft_reconnect")
+            return
+
+
+def main() -> None:
+    """Main watch and dispatch."""
+    if not SnapCache()["charmed-postgresql"].services["patroni"]["active"]:
+        print("Patroni is not running")
+        return
```
dragomirp, line 74: Exit if the snap service is running.

marceloneppel (approved):

templates/raft-observer.service.j2
taurus-forever: Discussed in private, decided to isolate systemd to minimal level: ensure process is always running as a single instance, all other checks will be inside the process (to isolate business logic).

taurus-forever (approved):
Thank you!

PR #1765, opened 2026-06-11
[DPE-10284] fix(charm): skip user/role setup on read-only standby cluster

marceloneppel:
## Issue
On a cluster<>cluster async-replication deployment, after a full outage (all units on both clusters stopped, then restarted), the standby cluster's leader gets stuck in error. Its re-emitted `start` hook runs user/role setup (`CREATE EXTENSION ...`) against the read-only standby and fails with `psycopg2.errors.ReadOnlySqlTransaction: cannot execute CREATE EXTENSION in a read-only transaction`.

## Solution
- Add `is_standby_cluster` (relation-based, independent of the Patroni API) and skip `_setup_users` on a standby cluster — those roles/users/extensions are created on the primary and arrive via replication, so the write DDL must not run on the read-only standby.
- Refactor `backups._is_standby_cluster` to delegate to the new charm property.
- Add unit tests and an integration test (`test_async_replication_outage.py`) that does a full outage + restart and asserts the standby leader recovers (validated on a live LXD deployment, before/after).

## Checklist

### DIFF
--- a/src/backups.py
+++ b/src/backups.py
@@ -164,7 +164,7 @@ class PostgreSQLBackups(Object):
 
     def _can_unit_perform_backup(self) -> tuple[bool, str | None]:
         """Validates whether this unit can perform a backup."""
-        if self._is_standby_cluster():
+        if self.charm.is_standby_cluster:
             return False, STANDBY_CLUSTER_CREATE_BACKUP_ERROR_MESSAGE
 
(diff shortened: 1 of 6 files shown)


taurus-forever (approved):
Nice! Thank you!

dragomirp (approved):

PR #1770, opened 2026-06-15
[DPE-10012] Add PostgreSQL 16.14 (r1157/1158) to releases.md

taurus-forever:
## Issue

PostgreSQL release notes have lack of 16/stable revisions 1157/1158.

## Solution

Add a missing item to release.md

## Checklist


a-velasco (dismissed):

carlcsaposs-canonical:
updated refresh docs in da01c216 (looks like we missed the last stable release as well)

carlcsaposs-canonical (approved):

taurus-forever (commented):
Thank you @carlcsaposs-canonical , LGTM. 
Let's discuss it on 1:1, in theory we support from/to=all withing the same track with growing revisions. I am not sure about the real benefits of growing table long-term.

We might/should document known issues we can't address... (aka inverted table). Let's discuss it in private first. Tnx!

a-velasco (approved):

carlcsaposs-canonical:
> Thank you @carlcsaposs-canonical , LGTM. Let's discuss it on 1:1, in theory we support from/to=all withing the same track with growing revisions. I am not sure about the real benefits of growing table long-term.
> 
> We might/should document known issues we can't address... (aka inverted table). Let's discuss it in private first. Tnx!

long-term:
- would like to surface this information in a better format than table—e.g. you enter what your current version is, get a suggested version to refresh to (with option to override), and then see the release notes for all in-between versions & specific refresh instructions for that refresh (automatically composed)
    - IMO release notes and refresh docs should be combined/the same thing
- for each version, I think we should have a suggested version to refresh to (and that we test)—otherwise combinations will be too many to test each one well. (also if our users go down the same path, that path will get better tested/bug fixed)
- if we have semantic-like version that is exposed to users instead of revision, we can simplify greatly with `>=` instead of listing revisions. can't do that with revisions since revision indicates build time instead of semantic version

PR #1792, opened 2026-06-18
[DPE-10450] Fix crash restoring non-settable unit status in PG16

taurus-forever:
## Issue

The unit.status getter can return statuses the setter rejects (e.g. an "error" status left over from a previously failed hook). When storage is already attached, _check_detached_storage restored the cached status verbatim, raising InvalidStatusError ("status must be in active, blocked, maintenance, waiting, not 'error'") and deadlocking the unit in error.

## Solution

Only restore the cached status when it is a settable status type.
(cherry picked from commit 0aaa1e28e33909b6d24881b13f089a86ab6e7494)

Assisted-by: Claude:claude-4.8-opus

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -533,6 +533,17 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             return
         self.unit.status = status
 
+    def _restore_unit_status(self, status: ops.StatusBase) -> None:
+        """Restore a previously cached unit status.
+
+        The status getter can return statuses that the setter rejects (e.g. an
+        "error" status left over from a previously failed hook). Skip restoring
+        any status juju does not allow us to set, to avoid an InvalidStatusError
+        that would deadlock the unit.
+        """
+        if isinstance(status, ActiveStatus | BlockedStatus | MaintenanceStatus | WaitingStatus):
+            self.set_unit_status(status)
+
     def _reconcile_refresh_status(self, _=None):
         if self.unit.is_leader():
             self.async_replication.set_app_status()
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -2613,7 +2613,13 @@ def _check_detached_storage(self) -> None:
...
                     self.unit.status = WaitingStatus("Data directory not attached")
                     raise StorageUnavailableError()
-        self.unit.status = cached_status
+        # The status getter can return statuses that the setter rejects (e.g. an
+        # "error" status left over from a previously failed hook). Only restore
+        # the cached status if it is one juju allows us to set back.
+        if isinstance(
+            cached_status, ActiveStatus | BlockedStatus | MaintenanceStatus | WaitingStatus
+        ):
+            self.unit.status = cached_status
```
marceloneppel: One gap: `enable_disable_extensions` does the same cache-then-restore (original_status = self.unit.status → self.set_unit_status(original_status)), so an `ErrorStatus` cached there hits the identical crash. Worth a follow-up (or a shared guard) so this doesn't resurface from the extensions path.

The test only covers ErrorStatus, though the comment names `unknown` too.

Also, does the error recover in the case you're making the fix for? I ask because we need to set the proper state after the error is fixed.

taurus-forever: Tnx, I saw enable_disable_extensions, but decided to stay focused here, will expand it.

Tested using update-config=1s, the charm has recovered from the error with this fix... while it was permanent without it.

taurus-forever: Improved in 6795bc0

marceloneppel (approved):

dragomirp (approved):

PR #1795, opened 2026-06-22
[DPE-10454] Use better status instead of blocked for transient update failures

taurus-forever:
Use maintenance/waiting status instead of blocked one.

The 'failed to update cluster members on member' BlockedStatus was
misleading: these are transient, self-recovering conditions (a RetryError
from update_config(), or a peer IP not yet published) rather than states
needing operator intervention. Replace them with MaintenanceStatus /
WaitingStatus and a 'updating cluster members' message.

Update test_on_peer_relation_changed accordingly, and set an explicit
BlockedStatus before exercising the early-exit guard so that path stays
covered.

Assisted-by: Claude:claude-5-fable

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1254,7 +1254,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             # Update the members of the cluster in the Patroni configuration on this unit.
             self.update_config()
         except RetryError:
-            self.set_unit_status(MaintenanceStatus("updating cluster members"))
+            self.set_unit_status(MaintenanceStatus("cluster member update failed, retrying"))
             return
         except ValueError as e:
             self.set_unit_status(BlockedStatus("Configuration Error. Please check the logs"))
@@ -1473,7 +1473,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             try:
                 self.update_config()
             except RetryError:
-                self.set_unit_status(MaintenanceStatus("updating cluster members"))
+                self.set_unit_status(MaintenanceStatus("cluster member update failed, retrying"))
         else:
             self.set_unit_status(WaitingStatus("waiting for peer IP"))
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -1254,7 +1254,7 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             # Update the members of the cluster in the Patroni configuration on this unit.
             self.update_config()
         except RetryError:
-            self.set_unit_status(BlockedStatus("failed to update cluster members on member"))
+            self.set_unit_status(MaintenanceStatus("updating cluster members"))
             return
         except ValueError as e:
             self.set_unit_status(BlockedStatus("Configuration Error. Please check the logs"))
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -1412,9 +1412,9 @@ def add_cluster_member(self, member: str) -> None:
             try:
                 self.update_config()
             except RetryError:
-                self.set_unit_status(BlockedStatus("failed to update cluster members on member"))
+                self.set_unit_status(MaintenanceStatus("pending update cluster members on member"))
         else:
-            self.set_unit_status(BlockedStatus("failed to update cluster members on member"))
+            self.set_unit_status(WaitingStatus("pending update cluster members on member"))
```
taurus-forever: Using WaitingStatus here as a flag to recognize RetryError==MaintenanceStatus.
Should I sync all to the same status?

dragomirp (dismissed):

tests/unit/test_charm.py
```diff
@@ -2526,8 +2527,8 @@ def test_add_cluster_member(harness):
         _update_config.side_effect = RetryError(last_attempt=None)
         harness.charm.add_cluster_member("postgresql-0")
         _update_config.assert_called_once_with()
-        assert isinstance(harness.charm.unit.status, BlockedStatus)
-        assert harness.charm.unit.status.message == "failed to update cluster members on member"
+        assert isinstance(harness.charm.unit.status, MaintenanceStatus | WaitingStatus)
```
dragomirp, line 2532: Better to  assert the exact status we are expecting here.

src/charm.py
marceloneppel: What if we have something like `WaitingStatus("waiting for peer IP")` here and `MaintenanceStatus("updating cluster members")` in L1415 and L1196 to make the status messages correspond to each state?

taurus-forever: Fixed in 55e5495ff4a45fe1a1175c28fc9eed9e6a242458

tests/unit/test_charm.py
taurus-forever: Fixed in 55e5495ff4a45fe1a1175c28fc9eed9e6a242458

src/charm.py
```diff
@@ -1254,7 +1254,7 @@ def _on_peer_relation_changed(self, event: HookEvent):
             # Update the members of the cluster in the Patroni configuration on this unit.
             self.update_config()
         except RetryError:
-            self.set_unit_status(BlockedStatus("failed to update cluster members on member"))
+            self.set_unit_status(MaintenanceStatus("updating cluster members"))
```
dragomirp: I'm not sure we should remove the failed part of the status message here, since technically we got a retry error even if the charm will heal.

dragomirp (dismissed):

src/charm.py
taurus-forever: Requested by @marceloneppel  in https://github.com/canonical/postgresql-operator/pull/1795#discussion_r3462173056 

Should it be `MaintenanceStatus("updating cluster members, pending...")` ?

marceloneppel: I think Dragomir is right about making the failure clear. Maybe use something like `cluster member update failed, retrying`.

taurus-forever: Updated in 33c1ae49cd1bb1d08ad91f708b12748b08f955f6

marceloneppel (approved):

dragomirp (approved):

Issue #1809, opened 2026-06-24
Transient alert PostgresqlHighRollbackRate


ggouzi:
## Steps to reproduce
On a machine cloud:
relate postgresql charm to otelcolector (2/stable rev 256)
relate postgresql to maas-region:db (3.7/candidate 462)

## Expected behavior
`PostgresqlHighRollbackRate` does not fire.
Or `PostgresqlHighRollbackRate` fires but fires continuously if the rollack rate is indeed high (>2%)

## Actual behavior
`PostgresqlHighRollbackRate` alert keeps firing and closing regularly

## Versions

Operating system: 24.04
Juju CLI: 3.6.23
Juju agent: 3.6.23
Charm revision: 16/stable rev 1089

## Additional context

The alert is defined as:
```
    - alert: PostgresqlHighRollbackRate
      expr: 'sum by (namespace,datname,instance,datid) ((rate(pg_stat_database_xact_rollback{datname!~"template.*|postgres",datid!="0"}[3m])) / ((rate(pg_stat_database_xact_rollback{datname!~"template.*|postgres",datid!="0"}[3m])) + (rate(pg_stat_database_xact_commit{datname!~"template.*|postgres",datid!="0"}[3m])))) > 0.02'
      for: 0m
      labels:
        severity: warning
      annotations:
        summary: PostgreSQL instance {{ $labels.instance }} has a high rollback rate instance.
        description: |
          The ratio of transactions being aborted compared to committed is > 2 %.
          This is probably happening due to unoptimized configurations related to commit delay, connections, memory, and WAL files.
          LABELS = {{ $labels }}
```

Please note the  `for: 0m` which is really weird, given that the metrics are aggregate with a rate function over a 3m interval. I suspect this is why alertmanager reports so many times the flapping alert.

taurus-forever:
Merged and backported for all PG VM+K8s 14+16.
Resolved, please test it on your side and reopen if necessary.
Tnank you for reporting and PR!

PR #1810, opened 2026-06-24
Change alert evaluation time of PostgresqlHighRollbackRate from 0m to 5m

ggouzi:
Attempt to solve https://github.com/canonical/postgresql-operator/issues/1809


taurus-forever (approved):
LGTM in general. Thank you for the bugreport and PR!
PostgresqlHighRollbackRate 0m=5m... smells like miss in testing.
BTW, it should be also backported to PostgreSQL VM 16/edge and PostgreSQL K8s 14+16.

7annaba3l (approved):
LGTM
Note on the side: An improvement that might be added later is to ensure a minimum amount of activity so that a transaction being rolled-back when you have 2 tps is not flagged as problematic.

marceloneppel (approved):

PR #1816, opened 2026-06-26
[DPE-10615] Migrate TLS to the single-kernel library (VM)

marceloneppel:
## Issue

The VM charm owns its TLS subsystem (`src/relations/tls.py`, `push_tls_files_to_workload`) duplicated from the K8s charm and the single-kernel library's now-equivalent `events.tls.TLS` + `TLSManager`.

## Solution

Migrates the VM charm's TLS onto the library's operator-cert handler (`events.tls.TLS`) + `TLSManager`. The events handler owns the two requirers and is constructor-injected into `TLSManager`; live-fetch getters (no operator-cert peer secrets, rollback-clean).

- Bump `postgresql-charms-single-kernel` from 16.3.2 to 16.3.4 to consume the lib TLS handler + manager; redirect the operator-cert getters to `tls_manager` (incl. `watcher.py` `patroni-cas` and `postgresql_provider.py`).
- Reload bridge `_reload_tls_after_push`: reloads after the lib pushes certs; fires on `relation_broken` so detaching the TLS operator re-renders Patroni with TLS off.
- Internal-cert regeneration routed through `_regenerate_internal_cert` (generate → push → reload; `reload=False` at bootstrap, where the leader/replica already renders `patroni.yml` and the internal cert does not toggle `ssl`).
- Remove `src/relations/tls.py` and `push_tls_files_to_workload` (now lib-owned).
- Rewrote `tests/unit/test_tls.py` to cover the lib-backed wiring (reload bridge guards, `is_tls_enabled` redirect, internal-cert CN-mismatch path) instead of the removed `relations.tls.TLS` module; renamed `TLS`→`TLSManager` patch targets in `tests/unit/test_charm.py` and dropped the now-deleted `test_push_tls_files_to_workload`.

The bridge and its four observers are transitional — they stay charm-side only until the `update_config`/Patroni-config subsystem ports to the library, at which point push+reload become a single lib-owned atomic step and these charm methods are removed.

## Checklist


taurus-forever (approved):
Nice!

dragomirp (approved):

Issue #1550, opened 2026-03-26
Removal/cleanup hooks not implemented


nvinuesa:
While investigating the Juju issue https://github.com/juju/juju/issues/21860 I found out that what was actually happening is that the postgresql (16/edge) charm was never releasing the storage (looks like one of the snaps is still running). This prevents the correct storage removal which itself prevents the machine removal which prevents the graceful model removal. This works of course if we run `destroy-model --force ...`.
The storage of the model after running model destroy is the following:
```
  Storage Unit  Storage ID  Type        Status     Message                                                
                data/10     filesystem  detaching  umount failed: umount:                                 
  /var/snap/charmed-postgresql/common/var/lib/postgresql: target is busy.: exit status 32                 
...
                logs/3      filesystem  detaching  umount failed: umount:                               
  /var/snap/charmed-postgresql/common/data/logs: target is busy.: exit status 32                          
```
After discussing with @marceloneppel we came to the conclusion that the removal-related hooks handlers are not implemented yet in the postgresql charm (namely `stop` or `storage-detaching`).

__Note: One important implementation detail is that the storage detaching and removal in Juju happens in a concurrent worker with retries. This means that even if the snaps are definitely stopped during the `stop` hook (which will run after `storage-detaching`) it will not block the eventual storage removal.__

taurus-forever:
IMHO, there are two orthogonal topics touched:
 1) attach/detach storage event inside charm.py (initial bugreport)
 2) keep the storage after the unit removal ([#1550 comment](https://github.com/canonical/postgresql-operator/issues/1550#issuecomment-4132354585))

---

Let's start with case 2): 
 * all below is Juju storage provider specific!
 * normally storage stays after unit removal (`juju remove-storage` is expected).
 * always pay attention to Juju message on `juju remove-application postgresql`:
   *  `- will remove storage pgdata/0` OR `- will detach storage pgdata/0`
 * the default lxd storage pool (dir) is a bit special: it always remove storage (even no `--destroy-storage` requested), so you have to use lvm storage to play this well:
```
> lxc storage create lvm lvm
Storage pool lvm created
> lxc storage volume create lvm foo
Storage volume foo created
> juju create-storage-pool juju-lvm lxd driver=lvm
> juju deploy postgresql --channel 14/stable --storage pgdata=5G,juju-lvm

> juju remove-application postgresql 
WARNING This command will perform the following actions:
will remove application postgresql
- will remove unit postgresql/0
[...]

PR #1827, opened 2026-07-02
[DPE-9685] Release storage on teardown

marceloneppel:
## Issue

`storage-detaching` runs before the relation-departed and stop hooks, and nothing else stops the `charmed-postgresql` snap, so on teardown the snap keeps the storage mounts busy and Juju's unmount fails (`target is busy`). Juju 3.6 masks this by force-removing still-Dying storage; 4.0 leaves it to unmount, so the failure surfaces there. Fixes https://github.com/canonical/postgresql-operator/issues/1550 on the teardown path.

## Solution

Stop the workload in the `storage-detaching` hook when the whole application is going away (`planned_units == 0`): stop the cluster topology observer, stop the log-rotation background process (new `RotateLogs.stop_log_rotation`), and stop the snap services, so Juju can unmount the storage. Scale-down (`remove-unit`) is intentionally not handled here — the surviving cluster still needs this unit's Patroni reachable to remove it from raft — and is covered by the stacked #1848. The regression test is pinned to the `juju40` spread variant, since 3.6 masks the bug. Force re-attaching/detaching of storage is out of scope.

## Checklist


src/charm.py
```diff
@@ -2563,6 +2567,19 @@ def _install_snap_package(
...
             raise
 
+    def _on_storage_detaching(self, _) -> None:
+        """Stop the workload so Juju can unmount the storage on app teardown."""
+        # On scale-down the surviving cluster still needs this unit's Patroni to
+        # remove it from raft; only stop when the whole app is going away.
+        if self.app.planned_units() > 0:
+            return
+        self._observer.stop_observer()
+        self._rotate_logs.stop_log_rotation()
```
taurus-forever, line 2577: 1) stop exporters IF COS is related?
2) (force) stop pgbackrest IF removal has been called during backup in progress on some unit

src/charm.py
```diff
@@ -2563,6 +2567,19 @@ def _install_snap_package(
             )
             raise
 
+    def _on_storage_detaching(self, _) -> None:
+        """Stop the workload so Juju can unmount the storage on app teardown."""
+        # On scale-down the surviving cluster still needs this unit's Patroni to
+        # remove it from raft; only stop when the whole app is going away.
+        if self.app.planned_units() > 0:
+            return
```
taurus-forever, line 2575: Will the scale-down to 0 be OK? The further scale-up would be able to re-use the storage?

tests/integration/adapters.py
```diff
@@ -379,7 +379,8 @@ def destroy_unit(
     def list_storage(self, filesystem: bool = False, volume: bool = False) -> list[TStorageInfo]:
         """Lists storage details."""
         raw = self._juju.cli("list-storage", "--format", "json")
-        json_ = json.loads(raw)
+        # Juju 4.0 prints nothing (not `{}`) for a model with no storage.
+        json_ = json.loads(raw) if raw.strip() else {}
```
taurus-forever, line 383: Ask Juju team to be consistent?
It might be an invisible for them regression.

tests/spread/test_storage_detaching.py/task.yaml
```diff
@@ -0,0 +1,12 @@
+summary: test_storage_detaching.py
+environment:
+  TEST_MODULE: test_storage_detaching.py
+  # #1550 only reproduces on Juju 4.0 (3.6 masks it by removing still-Dying storage),
+  # so pin this task to a juju40 variant and drop the default juju36.
+  CONCIERGE_JUJU_CHANNEL/juju40: 4.0/stable
```
taurus-forever, line 7: I see the logical need to run tests on Juju 3.6 as well. e.g. to make sure it doesn't produce regression here. WDYT?

tests/spread/test_storage_detaching.py/task.yaml
```diff
@@ -0,0 +1,12 @@
...
+variants:
+  - -juju36
```
taurus-forever: juju36 vs juju40 obove?

taurus-forever (commented):
Nice! Some questions.

src/charm.py
```diff
@@ -2563,6 +2567,19 @@ def _install_snap_package(
...
+        self._observer.stop_observer()
+        self._rotate_logs.stop_log_rotation()
+        try:
+            snap.SnapCache()[charm_refresh.snap_name()].stop()
```
carlcsaposs-canonical: would it be worth disabling (in addition to stopping) in case unit restarts after storage_detaching and before finishing teardown?

carlcsaposs-canonical (dismissed):

src/charm.py
marceloneppel: Makes sense. Done on https://github.com/canonical/postgresql-operator/pull/1827/changes/b78b4562bdd8b0a3a2198f0f020a8f27f51c254d.

tests/spread/test_storage_detaching.py/task.yaml
marceloneppel: I think we should do that as well. Done on https://github.com/canonical/postgresql-operator/pull/1827/changes/47fa98ca07dca543219e25dc4350cd5ecf048419.

tests/spread/test_storage_detaching.py/task.yaml
marceloneppel: Removed on https://github.com/canonical/postgresql-operator/pull/1827/changes/47fa98ca07dca543219e25dc4350cd5ecf048419 to allow the test to run on Juju 3.6.

carlcsaposs-canonical (approved):

tests/integration/adapters.py
marceloneppel: I'll get more details to share with them.

dragomirp (approved):

src/charm.py
marceloneppel: It will. I tested it locally and it worked fine (reused the storage).

marceloneppel: It will. I tested it locally and it worked fine (reused the storage).

src/charm.py
marceloneppel: The exporters are snap services so they will be stopped.

pgBackRest TLS server will be stopped two. Either way, even the pgBackRest executable process (used in a backup) would be stopped when the unit is turned down.

PR #1836, opened 2026-07-07
[MISC] Limit solver to 1d old dependencies (14/edge)

dragomirp:
Follow up to https://github.com/canonical/data-platform/pull/42.

Add minimum age for Poetry lock


poetry.toml
```diff
@@ -0,0 +1,2 @@
+[solver]
+min-release-age = 1
```
dragomirp, line 2: `14/edge` doesn't use the single kernel, so it doesn't need the exception.

marceloneppel (approved):

carlcsaposs-canonical (approved):

PR #1837, opened 2026-07-08
[DPE-10203] fix(async-replication): recover async replication after a dead datacenter

marceloneppel:
## Issue

DPE-10203: when the primary datacenter of an async-replication pair dies, the surviving cluster cannot re-establish replication to a fresh cluster. After force-promoting the standby and force-removing the dead consumed offer (`remove-saas --force`, which never delivers `relation-broken`), `create-replication` fails: the force-promotion leaves an orphaned `promoted-cluster-counter` that makes the cluster report itself as an existing primary ("There is already a replication set up."), the Patroni DCS keeps a stale `standby_cluster` that keeps the promoted cluster read-only, and reads of the dying relation's databags crash every hook.

## Solution

Owner-side and teardown fixes, each surviving the dead-DC sequence:

- own the shared cluster-credentials secret under a distinct offer-side label (`async-replication-secret-offer`), so an owner-create can never collide with a stale consumer alias of the legacy shared label;
- reconcile the orphaned `promoted-cluster-counter` from update-status — including after a new offer relation has already formed (the recovery sequence offers before running `create-replication`), when no live relation mirrors the counter;
- clear the stale DCS `standby_cluster` when this cluster is the primary, so the promoted cluster converges out of read-only standby-leader state;
- treat an unreadable dying async relation as absent in relation reads, so hooks reconcile instead of crashing;
- regression test driving the full dead-DC ordering (3 models, 2xPG + watcher each).

Also synced with current `16/edge` (single-kernel lib 16.3.5 `update_config` adoption among 28 commits); the stale-standby fix moved to the new `update_config` seam that refactor created.

## Checklist


marceloneppel:
Superseded by #1913 (same fix set restructured: labelless secret ownership replaces the distinct-offer-label workaround, per review). CI was green modulo documented base flakes.

PR #1838, opened 2026-07-08
[DPE-10203] fix(async-replication): reference the offer secret by id on the consumer

marceloneppel:
## Issue

DPE-10203 (stacked on #1837): the consumer/standby side of async replication fetched the offer's shared secret with `get_secret(id=..., label=SECRET_LABEL)`, registering a consumer-side alias under the legacy shared label. Juju keeps that alias reserved even after the remote secret is gone, which is the other half of the post-dead-DC deadlock: a former standby that is later promoted can neither read nor create a secret under that label.

## Solution

Consumer-side and robustness fixes:

- reference the offer secret purely by the id published in relation data — no label, so no consumer-side alias is ever registered and nothing can go stale;
- drop the legacy `SECRET_LABEL` constant from production code: nothing attaches it anymore, and the tests that guard against reintroducing it use the literal string (the integration test already did, to stay black-box);
- tolerate an unreadable dead peer in all relation-databag reads, so the teardown path reconciles instead of crashing hooks;
- fall back to the current unit count when `planned_units()` fails on the goal-state read (which errors while a force-removed SAAS churns), keeping the `_patroni` property usable in every hook.

The owner-side label (`OFFER_SECRET_LABEL`) stays by design: the owner needs a stable handle to its own secret across hooks, one that survives the loss of the relation data carrying the published id — exactly what a dead-DC teardown destroys.

## Checklist


marceloneppel:
Superseded by #1913 — the consumer-by-id change is included there; the label constants are gone entirely (labelless design).

PR #1864, opened 2026-07-20
refactor(tls): reload on tls_files_pushed instead of certificate_available

marceloneppel:
## Issue

The lib TLS handler pushes cert files to the workload on `certificate_available`, and this charm reloads PostgreSQL in a *separate* observer of the same event. `defer` is per-observer, so a push that defers (e.g. a Pebble write fails during a cert renewal) while the reload observer completes leaves the retried push writing the new cert with nothing reloading — the reload ran against the stale on-disk files and won't run again. The bridge also duplicated the handler's two readiness guards and relied on ops observer ordering. Raised in review on the TLS migration.

## Solution

The lib now emits a `tls_files_pushed` event after a successful push (canonical/postgresql-single-kernel-library#189). Observe that single event instead of the four `certificate_available` / `relation_broken` observers, and drop the two guards (`internal-ca` present, `client_tls_files_on_disk()`): the event fires only once the files are on disk, so a deferred push no longer triggers a reload against files that were never written, and the observer-order dependency is gone. Re-pin the lib to the commit that adds the event.

Stacked on the update_config adoption.

## Checklist


marceloneppel:
Folded into #1849: the TLS `tls_files_pushed` rewire now rides with the update_config adoption in a single per-charm PR (the two only existed separately because development was incremental). Superseded.

Issue #1859, opened 2026-07-14
Unexpected exception in Patroni: BlockingIOError: [Errno 11] Resource temporarily unavailable


alnvdl-work:
## Steps to reproduce
Not really sure to be honest... This seems to have happened out of the blue in a deployment in PS7. Reporting it here because the charm logged `Unexpected exception raised, please report it as a BUG`.

I reported this separately from https://github.com/canonical/postgresql-operator/issues/1358 because no units were removed, AFAIK.

## Expected behavior
Units auto-recover from whatever caused them to break.

## Actual behavior
Cluster is unstable. It is impossible to establish CMRs with it.

## Versions
Sorry, we would need IS for this. This DBaaS cluster was provisioned in late June, if it helps.

## Log output
Full logs with output from units 0 and 2 in the cluster (unit 1 seems to be missing for some reason): https://pastebin.canonical.com/p/XwPpqnrZNb/

Relevant segment:
```
[postgresql postgresql/2] 2026-07-14 08:34:29 UTC [3749374]: user=[unknown],db=[unknown],app=[unknown],client=10.156.72.47,line=1 LOG:  00000: connection received: host=10.156.72.47 port=55322
[postgresql postgresql/2] 2026-07-14 08:34:29 UTC [3749374]: user=[unknown],db=[unknown],app=[unknown],client=10.156.72.47,line=2 LOCATION:  BackendInitialize, postmaster.c:4354
[postgresql postgresql/2] 2026-07-14 08:34:29 UTC [3749374]: user=operator,db=postgres,app=[unknown],client=10.156.72.47,line=3 FATAL:  57P03: the database system is starting up
[postgresql postgresql/2] 2026-07-14 08:34:29 UTC [3749374]: user=operator,db=postgres,app=[unknown],client=10.156.72.47,line=4 LOCATION:  ProcessStartupPacket, postmaster.c:2336
[postgresql postgresql/2] 2026-07-14 08:34:31 UTC [3749384]: user=,db=,app=,client=,line=1 FATAL:  08006: could not connect to the primary server: connection to server at "10.156.72.53", port 5432 failed: Connection refused
[postgresql postgresql/2] Is the server running on that host and accepting TCP/IP connections?
[postgresql postgresql/2] 2026-07-14 08:34:31 UTC [3749384]: user=,db=,app=,client=,line=2 LOCATION:  WalReceiverMain, walreceiver.c:303
[...]

PR #1897, opened 2026-08-11
fix(relations): skip database request without a database name

marceloneppel:
## Issue

On a leader unit, a `database_requested` event that was deferred while Patroni was unavailable gets replayed by ops on every subsequent dispatch. If the relation has gone away in the meantime, ops supplies a dead relation whose remote application cannot be resolved, so the requested database name reads back as absent. `_collect_databases` coerced that missing value to an empty string, and `_on_database_requested` then indexed it, raising an uncaught `IndexError: string index out of range`.

ops replays deferred notices before the hook's own event, and only drops a notice once its handler returns. The exception therefore aborts every later hook on that unit permanently, actions included. The leader can no longer provision client relations or refresh endpoints, so any client relation that had not yet been served stays unserved. The unit itself sits in `error` on a failed hook.

Reported downstream as ISREQ-3825.

## Solution

Return early from `_collect_databases` when the relation no longer carries a database name, instead of coercing the missing value to an empty string. The handler already treats a `None` return as "nothing to do", so the stale notice drains on its next replay. Note this only applies once the fixed revision is running: a unit already wedged on an older revision needs to be refreshed and have its failed hook cleared before the guard can execute.

An empty name is not reachable from a live requirer: Juju drops a databag key whose value is an empty string, and the library only emits `database_requested` when that key is added. The guard therefore only affects the replayed-stale-notice path. The remaining `database[-1]` sites in this module are already protected by truthiness checks and are left unchanged.

Adds a regression test that drives the handler with an absent database name, which reproduces the `IndexError` without the guard.

## Checklist

### DIFF
--- a/src/relations/postgresql_provider.py
+++ b/src/relations/postgresql_provider.py
@@ -216,7 +216,10 @@ class PostgreSQLProvider(Object):
         self, user: str, event: DatabaseRequestedEvent
     ) -> tuple[str, list[str]] | None:
         # Retrieve the database name and extra user roles using the charm library.
-        database = event.database or ""
+        database = event.database
+        if not database:
+            logger.warning("Database name is not set in the relation data, skipping.")
+            return
         if database and database[-1] == "*":
             if len(database) < 4:
                 self.charm.unit.status = BlockedStatus(PREFIX_TOO_SHORT_MSG)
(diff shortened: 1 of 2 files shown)


alnvdl-work:
@marceloneppel

We reported the same issue in two places:
- https://github.com/canonical/postgresql-operator/issues/1859
- https://warthogs.atlassian.net/browse/ISREQ-3825

Should we update #1859 with the learnings from ISREQ-3825  and link to #1859 in this PR?

marceloneppel:
> We reported the same issue in two places:
> 
> * [Unexpected exception in Patroni: BlockingIOError: [Errno 11] Resource temporarily unavailable #1859](https://github.com/canonical/postgresql-operator/issues/1859)
> * https://warthogs.atlassian.net/browse/ISREQ-3825
> 
> Should we update #1859 with the learnings from ISREQ-3825 and link to #1859 in this PR?

Hi, @alnvdl-work! Yes, we should.

PR #1908, opened 2026-08-18
[DPE-10851] Secure peer IP to the address itself

taurus-forever:
Test...


taurus-forever:
TL;DR: the IP can be changed (e.g. on VM reboot => new IP => can't connect). All self-heal tests are red.
Aborting here. Smarter approach is necessary.

PR #1912, opened 2026-08-21
[DPE-10203] test(async-replication): cover the ticket's full teardown sequence

marceloneppel:
## Issue

DPE-10203 (stacked on #1837, like #1838): the dead-DC recovery regression test covered the core ordering but skipped two steps of the reported sequence: the watcher's machine goes down with the dead datacenter ("all Rome units"), and the dead relation is first attacked with `remove-relation --force` — the ticket's Issue 1, for which Juju delivers no events — before the `remove-saas --force` workaround clears the offer.

## Solution

Extend `test_async_replication_dead_dc_recovery.py` to replicate the ticket's full teardown sequence: the whole datacenter dies (cluster units and Raft-witness watcher alike), and the dead relation is attacked with `remove-relation --force` before the `remove-saas --force` workaround. The workaround tolerates "not found" (the `_consumer_alias_exists` idiom), so the test stays valid if Juju ever honors `remove-relation --force` — the current behavior keeps exercising the exact ordering that orphaned the consumer label and promotion counter.

The test's consumer-side assertions need #1838's changes, so the full sequence passes at the stack tip.

## Checklist

### DIFF
--- a/tests/integration/high_availability/test_async_replication_dead_dc_recovery.py
+++ b/tests/integration/high_availability/test_async_replication_dead_dc_recovery.py
@@ -7,10 +7,13 @@
 After a dead-datacenter failover, re-establishing async replication to a fresh
 cluster used to deadlock. The teardown ordering that triggers it:
 
-  1. the primary cluster's units all go away (a dead DC),
+  1. the primary datacenter dies — every machine in it is force-stopped, the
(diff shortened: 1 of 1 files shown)


marceloneppel:
Superseded by #1914 (same extended test, rebased on the restructured fix set).

PR #1913, opened 2026-08-21
[DPE-10203] fix(async-replication): recover async replication after a dead datacenter

marceloneppel:
## Issue

DPE-10203: when the primary datacenter of an async-replication pair dies, the surviving cluster cannot re-establish replication to a fresh cluster. After force-promoting the standby and force-removing the dead consumed offer (`remove-saas --force`, which never delivers `relation-broken`), `create-replication` fails: the promotion leaves an orphaned `promoted-cluster-counter` that makes the cluster report itself as an existing primary ("There is already a replication set up."), the Patroni DCS keeps a stale `standby_cluster` that keeps the promoted cluster read-only, and reads of the dying relation's databags crash every hook.

## Solution

Full fix set, landed as one revision because the pieces interlock:

- the shared cluster-credentials secret is owned **labelless** — the owner creates it without a label and re-finds it by the id persisted in app peer data; the consumer references it purely by the id published in relation data. Labels were the collision surface in both deadlock directions: a stale consumer alias blocks a later owner-create, and an owner label cannot survive the secret's id churn (a refreshed owner that misses its label mints a second secret, and Juju refuses to rebind a consumer label to the new id). A cluster refreshed from the legacy charm adopts its existing secret via the id in its own relation data, so the id never changes across a refresh;
- reconcile the orphaned `promoted-cluster-counter` from update-status, including after a new offer relation has already formed (the recovery sequence offers before running `create-replication`), when no live relation mirrors the counter — and `create-replication` / `promote-to-primary` also clear it inline before their guards read `_get_primary_cluster`, so recovery does not wait for an update-status cycle to happen to run first (a counter mirrored on a live relation survives, both in the reconciler and the actions);
- clear the stale DCS `standby_cluster` when this cluster is the primary, so a force-promoted cluster converges out of read-only standby-leader state;
- treat an unreadable dying async relation as absent in relation reads, and fall back to the current unit count when `planned_units()` fails on the goal-state read — so hooks reconcile instead of crashing during teardown.

The goal-state failure mode guarded was reproduced live against the current `16/edge` base charm (no guard): two clusters deployed with replication established, the primary's machines force-stopped, then `remove-saas --force` on the survivor — after which both units of the surviving cluster stayed in `error` status for the entire 12-minute observation window, hook-looping uncaught exceptions of the form:

```
ops.model.ModelError: ERROR saas application "db1" not found
  _on_async_relation_broken (async_replication.py:559)
    -> charm.update_config (charm.py:2774)
      -> lib config.py:509 update_config
        -> config.py:687 render_patroni_yml_file
          -> state.py:505 synchronous_node_count   [cached_property]
            -> peer_relation.py:414 planned_units
              -> ops planned_units -> goal_state()  [fails]
```

Regression test added through https://github.com/canonical/postgresql-operator/pull/1914/.

## Checklist


src/relations/async_replication.py
```diff
@@ -67,8 +70,36 @@
...
+
+def _same_secret_id(a: str | None, b: str | None) -> bool:
+    """Whether two Juju secret ids refer to the same secret.
+
+    Juju/ops may render an id as ``secret:<key>`` or ``secret://<uuid>/<key>``; compare on the
+    trailing key so a format difference doesn't mask a real match.
+    """
+    if not a or not b:
+        return False
+    return a.rsplit("/", 1)[-1].split(":")[-1] == b.rsplit("/", 1)[-1].split(":")[-1]
```
taurus-forever, line 90: Does it worth to branch it as a practical usability issue to Ops team?
Can they always return the full format `secret://<uuid>/<key>` or alike?

taurus-forever (dismissed):
Thank you!

src/relations/async_replication.py
marceloneppel: There is a still-open [bug](https://bugs.launchpad.net/juju/+bug/2075153) for Juju/ops.

taurus-forever: IMHO, it should be moved to ops: `FYI, this will be fixed in the next minor version of ops, most likely at the end of September 2024.` (the last message there). WDYT?

taurus-forever: But here https://github.com/canonical/operator/issues/1312 it is marked as completed.

src/charm.py
```diff
@@ -1566,6 +1566,21 @@ def _hosts(self) -> set[str]:
                 hosts.append(unit.name.replace("/", "-"))
         return set(hosts)
 
+    @property
+    def _planned_units(self) -> int:
```
dragomirp, line 1570: Heavy on the controller. We should cache the result and reuse across the charm.

marceloneppel: The charm reads `_planned_units` only once per hook invocation (each hook is a fresh dispatch process, so a cache here dedupes nothing - and this cost predates this PR, `16/edge` calls the same uncached read twice). The second `goal-state` round-trip per hook comes from the lib's render (`synchronous_node_count` ->
 `planned_units`), which a charm-side cache can't reach. Filed where a fix can serve both reads: https://github.com/canonical/postgresql-single-kernel-library/issues/294

src/relations/async_replication.py
marceloneppel: I'm checking whether that function was needed for my previous attempt to fix this async replication issue in https://github.com/canonical/postgresql-operator/pull/1838 and can be removed in this PR.

marceloneppel: I fixed the docstring (https://github.com/canonical/postgresql-operator/pull/1913/changes/9dd7924df15fa2a13b0eb81801ac492a0b91beac). It was an issue with the secret URI generated by previous revisions of the charm (with a duplicated model UUID):
```sh
~ juju show-unit -m lxd:lisbon db2/0 --format json \
  | jq -r '.[] | .["relation-info"][] | select(.["endpoint"] == "replication")
           | .["application-data"]["primary-cluster-data"]' | jq .
{
  "endpoint": "10.197.227.56",
  "secret-id": "secret://4962e04e-ae3c-4334-8e3b-f3e7974d1452///4962e04e-ae3c-4334-8e3b-f3e7974d1452/aflheo3ddfvkerflhdd0"
}
```
This `_same_secret_id ` function is needed if we upgrade an async replication setup.

taurus-forever: Oh. `secret://<model-uuid>//<model-uuid>/<key>` is crazy.
Thank you for tracing it down! LGTM

taurus-forever (approved):

src/relations/async_replication.py
marceloneppel: More details: we did that in https://github.com/canonical/postgresql-operator/blob/ffc4814a69dbb8aa2cf26c0199b0fe7275194b92/src/relations/async_replication.py#L346. After some time, ops started to output the model UUID in the secret ID, so our workaround created that weird secret ID.

src/charm.py
marceloneppel: Fixing on https://github.com/canonical/postgresql-single-kernel-library/pull/298 and https://github.com/canonical/postgresql-operator/pull/1958/changes.

dragomirp (approved):

PR #1961, opened 2026-09-11
[DPE-11134] fix(async-replication): sync system-user password rotations to the standby cluster

marceloneppel:
## Issue

DPE-11134: rotating a system-user password (e.g. `operator`) through a Juju user secret on the primary cluster of a cluster<>cluster (async replication) deployment updated the database on both sides, the primary cluster's Juju app secret and the shared async-replication secret, but never the standby cluster's own Juju app secret.

Two defects, both confirmed on a real two-model deployment (juju 3.6.28, LXD):

1. The standby cluster only pulled the shared secret during initial setup and on `secret_changed`. Juju does not deliver `secret-changed` events to the consumer units when the owner adds a revision across the cross-model relation (observed: owner revision 2 added at 17:50:04; zero `secret-changed` hooks on any consumer unit in the following ~90 minutes while the same units received one at consumption time). The standby's app secret therefore kept the pre-setup password indefinitely — the recovery credential silently went stale.
2. Every unit receives `secret_changed` for the shared secrets, but both branches of `PostgreSQLAsyncReplication._on_secret_changed` write application-scope data, which only the leader can write: the replicas crashed the hook in a loop (`hook failed: "secret-changed"`, `RelationDataAccessError` on `app_peer_data.update` inside `_get_secret`).

The config-changed rotation path additionally pushed the (pre-rotation) secret content to the shared secret before `_update_admin_password` ran, so that entry point never propagated either.

## Solution

- `_update_admin_password` now refreshes the async-replication data right after rotating, so the primary cluster pushes the new passwords into the shared secret in the same event (covers both the `secret_changed` and config-changed paths).
- `update_async_replication_data` now pulls the shared secret content into the standby cluster's app secret on every reconciliation of the consumer side — self-healing, not dependent on cross-model secret notifications.
- `_update_internal_secret` skips unchanged passwords: every rewrite is a new Juju secret revision, so unchanged values must not churn revisions.
- `_on_secret_changed` now exits early for non-leader units (leader-only app data writes).

Stacked on #1914 (base is its head branch).

## Verification

- TDD: event-driven unit test failed pre-fix on the missing refresh (`update_async_replication_data` never called), green after; 6 new unit tests total. Full unit suite: 217 passed. `tox run -e lint` passes.
- Integration (two-model cluster<>cluster on LXD, matching the ticket STR):
  - Pre-fix: after `juju update-secret`, the primary's app secret and the shared secret held the new password while the standby's app secret kept the original password, and both databases accepted the new password (replication) — the reported divergence.
  - Post-fix (charm refreshed on both apps): fresh rotation converged both clusters' app secrets within a minute with no hook errors on any unit, and the same at the default 5-minute update-status interval (no fast-forward), with both databases accepting the rotated password.
- New regression test module `test_async_replication_password_rotation.py` (+ spread task) mirrors the STR for CI: deploy both clusters, relate, create-replication, configure the user secret, rotate it, and assert both clusters' app secrets and database access converge.

## Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -2330,6 +2330,11 @@ class PostgresqlOperatorCharm(TypedCharmBase[CharmConfig]):
             self.set_unit_status(BlockedStatus("Password update for system users failed."))
             return
 
+        # Refresh the shared async-replication secret content. The standby cluster only
+        # learns about password rotations when the primary cluster updates that secret,
+        # otherwise its own Juju app secret keeps the old password (DPE-11134).
+        self.async_replication.update_async_replication_data()
+
         # Update and reload Patroni configuration in this unit to use the new password.
         # Other units Patroni configuration will be reloaded in the peer relation changed event.
         self.update_config()
--- a/src/relations/async_replication.py
+++ b/src/relations/async_replication.py
@@ -828,6 +828,14 @@ class PostgreSQLAsyncReplication(Object):
 
     def _on_secret_changed(self, event: SecretChangedEvent) -> None:
         """Update the internal secret when the relation secret changes."""
+        if not self.charm.unit.is_leader():
+            # Both branches update application-scope data (the shared secret content on
+            # the offer side, the internal app secret on the consumer side), which only
+            # the leader can write. Every unit receives the same secret-changed event;
+            # the leader reconciles for the whole cluster.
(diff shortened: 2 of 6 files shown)


marceloneppel:
Closed because the right fix is at https://github.com/juju/juju/issues/23288.

PR #1965, opened 2026-09-14
[MISC] Assert refresh version (16/edge)

dragomirp:
CI test to assert that workload version matches the refresh definition version.


marceloneppel (approved):

tests/integration/test_pg_hba.py
```diff
@@ -3,10 +3,12 @@
 # See LICENSE file for licensing details.
 import logging
 import re
+from pathlib import Path
 from time import sleep
 
 import psycopg2
 import pytest
+import tomli
```
carlcsaposs-canonical, line 11: if the test runner is >= ubuntu 24.04 (i.e. python >=3.11), `tomllib` from the standard library can be used instead of `tomli`

carlcsaposs-canonical (approved):

